PR #1, opened 2016-04-27
Supporting composer.json dependencies in profiles.

grasmash:


grasmash:
This relies on https://github.com/acquia/lightning/pull/119

grasmash:
This is obsolete, we no longer support drush make workflows.

PR #5, opened 2016-04-28
Removing 80 col line breaks from docs.

### LINKED ISSUE #4: Switch documentation to not use line breaks at 80 cols
This is a pet peeve of mine, since it makes updating documentation so unbearably hard—current docs are formatted thusly:

```
We highly recommend that you *do not use Windows* directly for development. 
Many development tools (e.g., drush, gulp, etc.) are not built or tested for 
Windows compatibility. Furthermore, most CI solutions (e.g., Travis CI, 
Drupal CI, etc.) do not permit testing on Windows OS.
```

If I were to add a few words to the first line, I would then have to spend the next minute reflowing all the rest of the text in that paragraph so it all breaks at 80 cols.

Most modern editors have line wrapping capabilities, so editing text in markdown files should still be simple/easy to do even if each paragraph is on it's own line and is not broken at 80 characters.

This will also make documentation PRs much better, since only individual text changes need to be tracked, and not line-by-line reflows. For example, see this PR: https://github.com/acquia/bolt/pull/3/files#diff-23d3084f68a11bf46e7e79b8892be209L125 — it looks like that whole paragraph was changed, when in reality only one line was changed, but all the lines had changes because of the reflow.

grasmash:
Resolves #4

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -11,8 +11,7 @@ The following high-level steps will be required to generate a new, fully functio
 
 ## System Requirements
 
-You should be able to use the following tools on the command line of your native
-operating system:
+You should be able to use the following tools on the command line of your native operating system:
 
 * [Git](https://git-scm.com/)
 * [Composer](https://getcomposer.org/download/)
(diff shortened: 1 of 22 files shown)


geerlingguy:
aaaand, this is awesome.

PR #9, opened 2016-04-30
Fixes #8 Removing unneeded extra composer install step

### LINKED ISSUE #8: composer install is run twice on initial setup
Following the readme, steps are `bolt.sh create`, cd into project and then `bolt.sh setup`.

Both `create` and `setup` run `composer install` on the created project directory.

Is there a reason why we need to run composer install twice?

CashWilliams:
from initial project creation


grasmash:
I believe this in necessary if you've freshly cloned bolt and generated your first bolt project. Otherwise, you'll need to execute `composer install` manually from the new site dir.

CashWilliams:
Correct, see my update in https://github.com/acquia/bolt/issues/8, going to close this but leave issue open.

PR #13, opened 2016-05-02
Added frontend:build to local:refresh task.

damontgomery:
`frontend:build` takes all the themes listed in `project.yml` and runs `npm run install-tools` and `npm run build`.

If no themes are listed, these tasks shouldn't run. The `npm run build` can be customized in a theme's `package.json` to run their own commands. For Nimbus, this is `npm install` and `gulp`.

There shouldn't be any reason not to include this.

Including this ensures that local devs always have the most up to date CSS files. Nimbus excludes these from the `develop` branch by default and builds them with Travis.


danepowell:
I'm confused, doesn't `setup:build:all` already run `frontend:build` as a dependency?

damontgomery:
@danepowell, you are correct. I was getting confused by a D7 version. There are things that don't really need to be run every time in `setup:build:all`, but I think it's fine for now.

PR #15, opened 2016-05-03
Make Bolt alias compatible with all Unix environments.

danepowell:
This is mostly just a string-replace of ~/.bash_profile (only exists on Mac, and includes .bashrc automatically) to ~/.bashrc (exists on all *nix environments).

I did add a bit of logic to accommodate anyone who is still using the old-style .bash_profile alias.

### DIFF
--- a/template/build/core/phing/tasks/setup.xml
+++ b/template/build/core/phing/tasks/setup.xml
@@ -125,7 +125,15 @@
   </target>
 
   <target name="setup:bolt:alias" description="Installs the Bolt alias for command line usage.">
-    <exec dir="${repo.root}/scripts/bolt" command="./install-alias.sh" passthru="true" checkreturn="true"/>
+    <echo>Bolt can automatically create a Bash alias to make it easier to run Bolt tasks.</echo>
+    <echo>This alias may be created in .bash_profile or .bashrc depending on your system architecture.</echo>
+    <propertyprompt propertyName="create_alias" defaultValue="y" useExistingValue="true" promptText="Create a bash alias now? (y/n)" />
+    <if>
+      <equals arg1="${create_alias}" arg2="y"/>
+      <then>
+        <exec dir="${repo.root}/scripts/bolt" command="./install-alias.sh" passthru="true" checkreturn="true"/>
+      </then>
+    </if>
   </target>
 
 </project>
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
   - ./bolt.sh configure
   - ./bolt.sh create
   # Call targets in the new 'bolted' project.
-  - ../bolted8/bolt.sh build:validate:test -Dbehat.run-server=true -Dbehat.launch-phantom=true
+  - ../bolted8/bolt.sh build:validate:test -Dcreate_alias=false -Dbehat.run-server=true -Dbehat.launch-phantom=true
   # Deploy build artifact.
   - export DEPLOY_PR=true
   - ../bolted8/scripts/deploy/travis-deploy.sh 8.x 8.x-build
--- a/template/.travis.yml
+++ b/template/.travis.yml
@@ -49,7 +49,7 @@ before_script:
   - git diff --exit-code
 
 script:
-  - ./bolt.sh -Dbehat.run-server=true -Dbehat.launch-phantom=true build:validate:test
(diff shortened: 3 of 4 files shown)


danepowell:
All of the documentation and standards say this should work fine and have no functional change on Macs... but Sarah says it might not work, needs to be tested.

sarahjean:
We should do this, but instruct mac users to add this to ~/.bash_profile:
```
# Source .bashrc
if [ -f ~/.bashrc ]; then
     . ~/.bashrc
fi
```

http://www.joshstaiger.org/archives/2005/07/bash_profile_vs.html

geerlingguy:
Yeah, I only use .bash_profile. Also, I hate when applications try to alter _my_ local environment configuration (I have it source controlled, so every time I redeploy my local environment, bolt's stuff will get deleted). Is this disable-able?

https://github.com/geerlingguy/dotfiles

danepowell:
I don't know exactly how you manage your dotfiles, but is there a way to satisfy both use cases? I really think this setup should happen by default (otherwise I can guarantee that some developers won't go through the extra step of setting it up), but is there a less intrusive way to do it, or a way to disable it?

geerlingguy:
For me, if you stick stuff in .bashrc, that's fine with me as I can include it or not from my source-controlled .bash_alias (so Sarah's suggestion works for me). But if a dev uses .bash_alias and you only modify .bashrc, then that dev will have to do the manual step. (and vice-versa)

geerlingguy:
e.g. http://unix.stackexchange.com/a/84688/16194

oxyc:
I version control my [dotfiles](https://github.com/oxyc/dotfiles/blob/master/.bash/shell.sh#L46:L48) as well and second having it disable-able.

If it's somehow possible to disable it then maybe something like this support the most common scenarios:

```sh
if [ "`basename "/$SHELL"`" = "zsh" ]; then
  DETECTED_PROFILE="$HOME/.zshrc"
elif [ -f "$HOME/.bashrc" ]; then
  DETECTED_PROFILE="$HOME/.bashrc"
elif [ -f "$HOME/.bash_profile" ]; then
  DETECTED_PROFILE="$HOME/.bash_profile"
elif [ -f "$HOME/.profile" ]; then
  DETECTED_PROFILE="$HOME/.profile"
fi

if [ ! -z "$DETECTED_PROFILE" ]; then
  if [ "`grep 'function bolt' $DETECTED_PROFILE`" ]; then
    echo "Alias for bolt already exists in $DETECTED_PROFILE"
    exit
  fi
  DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
  cat $DIR/alias >> $DETECTED_PROFILE

  echo "Added alias for bolt to $DETECTED_PROFILE"
  echo "Restart your terminal session to use the new command."
else
  echo "Could not install bolt alias. No profile found. Tried ~/.zshrc, ~/.bashrc, ~/.bash_profile and ~/.profile."
fi
```

danepowell:
It sounds like a solution that would satisfy everyone would be to (a) prompt for whether to install the alias, and (b) use Oskar's code snippet to more intelligently choose where to put it.

grasmash:
@danepowell That sounds good to me. I'd add that the prompt should only occur if the alias isn't already installed.

danepowell:
This should satisfy everyone. I've tested it locally on Ubuntu and it works great.

template/build/core/phing/tasks/setup.xml
```diff
@@ -125,7 +125,15 @@
   </target>
 
   <target name="setup:bolt:alias" description="Installs the Bolt alias for command line usage.">
-    <exec dir="${repo.root}/scripts/bolt" command="./install-alias.sh" passthru="true" checkreturn="true"/>
+    <echo>Bolt can automatically create a Bash alias to make it easier to run Bolt tasks.</echo>
+    <echo>This alias may be created in .bash_profile or .bashrc depending on your system architecture.</echo>
+    <propertyprompt propertyName="create_alias" defaultValue="y" useExistingValue="true" promptText="Create a bash alias now? (y/n)" />
```
grasmash: This message and prompt should go inside the check for create_alias.

kylebrowning:
Were missing fish, which uses functions.

 :(

kylebrowning:
`alias showmeeverything "ls -alh"

PR #18, opened 2016-05-04
ACSF support for deploy step

damontgomery:
- Add a flag in project.yml
- Add a deploy task that primarily copies settings.php needed for ACSF

We cannot re-use setup:acsf:init because it explicitly excludes the settings.php file for ACSF. It also runs against the local site, which doesn't work for deploy.

This allows us to use the default bolt setup for local dev and build an artifact that works on ACSF.

### DIFF
--- a/template/build/core/phing/tasks/deploy.xml
+++ b/template/build/core/phing/tasks/deploy.xml
@@ -106,11 +106,13 @@
 
     <!-- Copy required files from docroot. -->
     <copy todir="${deploy.dir}/docroot" overwrite="true">
-      <fileset dir="${docroot}">
+      <fileset dir="${docroot}" defaultexcludes="false">
         <!-- This should be similar to .gitigore. -->
-        <include name="**"></include>
+        <!-- We remove default excludes because they include .gitignore files we need, like the ACSF one. -->
+        <include name="**" />
         <exclude name="**/local.*" />
         <exclude name=".gitkeep" />
+        <exclude name=".DS_STORE" />
         <exclude name="core/**" />
         <exclude name="drush/contrib/**" />
         <exclude name="example.gitignore" />
@@ -120,8 +122,8 @@
         <exclude name="themes/contrib/**" />
         <exclude name="profiles/contrib/**" />
         <exclude name="modules/contrib/**" />
-        <exclude name="**/node_modules/**"></exclude>
-        <exclude name="**/bower_components/**"></exclude>
+        <exclude name="**/node_modules/**" />
+        <exclude name="**/bower_components/**" />
       </fileset>
     </copy>
 
--- a/template/build/core/phing/tasks/deploy.xml
+++ b/template/build/core/phing/tasks/deploy.xml
@@ -51,6 +51,13 @@
 
   <target name="deploy:artifact:build" description="Generates a deploy-ready build in deploy.dir."
           depends="deploy:artifact:clean, deploy:artifact:copy, deploy:artifact:composer:install, deploy:artifact:profile:make, frontend:build, deploy:artifact:sanitize">
+    <!-- If we are using ACSF, run the ACSF Deploy command. -->
+    <if>
+      <equals arg1="${hosting}" arg2="acsf"/>
+      <then>
+        <phingcall target="deploy:acsf:init"/>
+      </then>
+    </if>
   </target>
 
   <target name="deploy:artifact:clean" description="Deletes the contents of the deploy dir.">
@@ -169,5 +176,10 @@
         <include name="**/CHANGELOG.txt"/>
       </fileset>
     </delete>
+  </target>
+
+  <target name="deploy:acsf:init" description="Re-initialize ACSF with the settings.php changes required for artifact.">
+    <chmod file="${deploy.dir}/docroot/sites/default/settings.php" mode="0755" />
+    <exec dir="${deploy.dir}/docroot" command="${drush.bin} --include=${deploy.dir}/docroot/modules/contrib/acsf/acsf_init acsf-init -y" logoutput="true" checkreturn="true"/>
   </target>
 </project>
--- a/template/build/core/phing/tasks/deploy.xml
+++ b/template/build/core/phing/tasks/deploy.xml
@@ -182,6 +182,6 @@
 
   <target name="deploy:acsf:init" description="Re-initialize ACSF with the settings.php changes required for artifact.">
     <chmod file="${deploy.dir}/docroot/sites/default/settings.php" mode="0755" />
-    <exec dir="${deploy.dir}/docroot" command="${drush.bin} --include=${deploy.dir}/docroot/modules/contrib/acsf/acsf_init acsf-init -y" logoutput="true" checkreturn="true"/>
+    <exec dir="${deploy.dir}/docroot" command="${drush.bin} --include=${deploy.dir}/docroot/modules/contrib/acsf/acsf_init acsf-init -r ${deploy.dir}/docroot -y" logoutput="true" checkreturn="true"/>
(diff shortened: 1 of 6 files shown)


danepowell:
Why would we configure settings.php on every deployment, instead of just configuring it once and committing the "correct" settings.php to the source repo?

I just worry that this is going to cause confusion, since it's effectively fragmenting the source code and deploy artifact... for instance it will overwrite any custom changes you make to ACSF files in the source repo on deploy (such as .htaccess).

danepowell:
I just spoke with Rok, this really shouldn't be necessary at all.

If you run `bolt acsf:init`, this will prep your repo for ACSF deployment. The only thing it doesn't do is replace `settings.php` with the stock ACSF `settings.php`, but _that's okay_, because the ACSF module is patched (via composer.json) to ignore validation of settings.php:
https://github.com/acquia/bolt/blob/8.x/template/composer.json#L59

When you deploy to ACSF, it runs the version of acsf-init-verify that is in the application codebase, so it will happily ignore the hacked settings.php.

damontgomery:
I don't think we should be modifying ACSF files. The work in this PR supports a local default/settings.php and follows the ACSF guidelines of not touching their code.

The deploy tasks will re-run ACSF init and copy the necessary files over and everyone is happy.

Deploying to multiple hosting environments is an edge case and could be achieved with a separate custom deploy task that does not overwrite settings.php. These are supported by bolt as custom targets.

template/build/core/phing/tasks/deploy.xml
```diff
@@ -99,11 +106,13 @@
 
     <!-- Copy required files from docroot. -->
     <copy todir="${deploy.dir}/docroot" overwrite="true">
-      <fileset dir="${docroot}">
+      <fileset dir="${docroot}" defaultexcludes="false">
```
grasmash: Which necessary files does setting `defaultexcludes="false"` cause to be included?

template/build/core/phing/tasks/deploy.xml
```diff
@@ -170,4 +179,9 @@
       </fileset>
     </delete>
   </target>
+
+  <target name="deploy:acsf:init" description="Re-initialize ACSF with the settings.php changes required for artifact.">
+    <chmod file="${deploy.dir}/docroot/sites/default/settings.php" mode="0755" />
+    <exec dir="${deploy.dir}/docroot" command="${drush.bin} --include=${deploy.dir}/docroot/modules/contrib/acsf/acsf_init acsf-init -y" logoutput="true" checkreturn="true"/>
```
grasmash: Will it cause any problems to run this against both the canonical repo and the deployment artifact? Why is it necessary to do in both places?

danepowell: It's a paradigm shift from how we originally supported ACSF. Instead of Bolt supporting ACSF and ACE simultaneously out of this box, this adds a flag that you set to add ACSF compatibility "on-the-fly" to a build artifact.

If we're taking this approach, we also need to update the ACSF readme and rip out the ACSF-specific logic in the existing settings.php.

template/build/core/phing/tasks/deploy.xml
damontgomery: ACSF .gitignore, see https://www.phing.info/docs/guide/trunk/FileSet.html

Included files are 

```
Default excludes are: *~, #*#, .#*, %*%, CVS, CVS/**, .cvsignore, SCCS, SCCS/**, vssver.scc, .svn, .svn/**, ._*, .DS_Store, .darcs, .darcs/**, .git, .git/**, .gitattributes, .gitignore, .gitmodules
```

The cleanup task already removes .git directories.

template/build/core/phing/tasks/deploy.xml
damontgomery: Composer is updated with this PR. I'll look into the readme and settings.php.

damontgomery: Updated the readme and the default settings.php. Thanks for mentioning those.

kylebrowning:
Also this begs the question.

(And not something that will be normal)

What If I want to deploy to both ACSF and ACE?

damontgomery:
@kylebrowning, if you want to deploy to multiple environments, you can use Bolt as an example and make your custom deployment tasks in Phing.

If it becomes more common, let's add it. If not, I think we try to address the more common situations.

kylebrowning:
This fully works and is required to get deployment artifacts up to ACSF. 

It still allows local development to be unaffected and if settings.php things are needed we can use pre-settings-php and post-settings-php as described here https://docs.acquia.com/site-factory/tiers/paas/workflow/hooks

damontgomery:
I'm going to make a comment about factory hooks in the readme... One second.

template/build/core/phing/tasks/deploy.xml
grasmash: Rather than disabling all of these default excludes, I think we should simply include .gitignore files. E.g., `<include name="**/.gitignore"/>`. This seems like a more precise way to accomplish the end goal.

damontgomery: I tried that and it didn't work for me. The other option is to check for ACSF and explicitly include it's .gitignore.

I didn't see any reason we should be ignoring all those files and most do not apply.

template/readme/acsf-setup.md
```diff
@@ -1,6 +1,15 @@
...
 To configure a project to run on ACSF, perform the following steps after initially setting up Bolt:
- 
+
+1. Add a `hosting` variable set to `acsf` in `project.yml`
 1. Execute `./bolt.sh acsf:init` from the project root.
 1. Add the acsf module as a dependency to your installation profile.
+
+## Factory Hooks and settings.php
+
+Acquia Cloud Site Factory does not allow changes to `settings.php`. In order to make additions normally done with `settings.php`, factory hooks are used.
```
grasmash: It seems like we may want to include example scripts for pre-settings-php and post-settings-php. I'd be interested in how Bolt generated projects on ACSF are using these. I'd imagine at least a script with a require statement for `cache.settings.php`.

damontgomery: Created #21.

I don't think that is necessary for this PR.

grasmash:
As @danepowell said, we shouldn't need to fragment our settings.php configuration in this way. But, it seems that we _do_ need to--it's a constraint imposed by the architecture of ACSF. That may change in the future, but until it does, it seems that we should work within the constraints rather than maintaining a strategy that isn't supported by the ACSF module.

kylebrowning:
Im usually of the opinion that Acquia stuff shouldn't be including in this, but were clearly not stating that. 

The fact that this is all buried under a hosting variable means many will never use it, unless we add more features to the project.yml hosting variable.

template/composer.json
```diff
@@ -28,6 +27,7 @@
     "behat/mink-goutte-driver":     "*",
     "behat/mink-selenium2-driver":  "*",
     "behat/mink-browserkit-driver": "*",
+    "drush/drush":                  "dev-master",
```
danepowell: I think Moshe has said repeatedly that applications should provide their own version of Drush... it seems like we're going back on that here. If so, we should discuss whether that's still best practice.

grasmash: Ah, I hadn't considered that. Will running `drush @alias status` use the drush bin in the remote application's vendor dir?

kylebrowning: I think until ACE is fixed with 1.91 release, we should leave this here so peoples installs for D8 do not break like they did for us.

Once its released, we can move it back.

kylebrowning: #34

PR #30, opened 2016-05-05
Resolves #11: Changing Git unit tests to use --dry-run flag for commit tests.

### LINKED ISSUE #11: PHPUnit Git tests shouldn't create actual commits
Right now the Git tests attempt to create real live Git commits: https://github.com/acquia/bolt/blob/8.x/template/tests/phpunit/GitTest.php

This means that if you run unit tests on a local machine, you're unknowingly left with a contaminated Git history.

I'm not sure I see the value in running these tests at all on a continuous basis in a deployed Bolt installation... they seem more valuable for testing the initial Bolt setup. Assuming we need to keep them, think about using `--dry-run` or some other process to avoid creating permanent commits.

grasmash:


grasmash:
Add this for bad commits too, just in case.

grasmash:
If git hooks have not been installed, this causes undesired commits.

grasmash:
Closing in favor of #168.

PR #34, opened 2016-05-05
Moves drush back [Do not merge until ACE 1.91 release]

kylebrowning:
The issue here is that ACE has a block on using a vendor'ed drush. 

Once ACE 1.91 goes out we can move this back for best practices.

(Unless its not a best practice anymore)

### DIFF
--- a/template/composer.json
+++ b/template/composer.json
@@ -10,6 +10,7 @@
   ],
   "require": {
     "composer/installers": "^1.0.20",
+    "drush/drush":"dev-master",
     "drupal/acquia_connector": "8.1.*",
     "drupal/acsf": "~8.1",
     "drupal/core": "~8",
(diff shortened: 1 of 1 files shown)


grasmash:
@kylebrowning Do you know if this can be merged now? I don't see mention of using vendored drush here: https://docs.acquia.com/node/24746

kylebrowning:
It's not deployed yet AFAICT. I'll check in the morning.

danepowell:
1.91 went to ACP yesterday but there is no ETA for an ACE release.

I don't think providing your own Drush was ever documented best practice, it was just the opinion of a few people (who probably all heard it from Moshe). Given how badly this failed in practice, I would vote to just close this PR and use the bin provided by ACE, unless we find a compelling reason to go back to providing our own.

kylebrowning:
Honestly I'm fine with closing this as well.

PR #37, opened 2016-05-05
Improving dev desktop settings.php compatibility.

grasmash:


template/project.yml
```diff
@@ -1,7 +1,7 @@
 project:
-  # The acquia_subname value is the machine name of your Github repository.
   # This will determine the the directory name of the new repository.
-  acquia_subname: 'bolted8'
+  # Dev Desktop users: this should match your local site name
```
geerlingguy: This docs line makes it look like this configuration directive would be important for Dev Desktop users but might not be as important for others... is there a way this could be worded better to match what it is?

grasmash: Any wording you have in mind? :p

kylebrowning:
I have tested this, thank you so much for looking into, it works great.

PR #42, opened 2016-05-09
Removing branch argument from deploy tasks.

grasmash:


damontgomery:
This looks good to me. Related to #47.

PR #43, opened 2016-05-10
Resolves #40: Disable APCu caching to prevent memory exhaustion

### LINKED ISSUE #40: Provide default services.yml, including APCu config
On one of our ACE sites, we are seeing failing page loads with this error:
"Warning: apcu_store(): Unable to allocate memory for pool. in Drupal\Core\Cache\ApcuBackend->set() (line 177 of core/lib/Drupal/Core/Cache/ApcuBackend.php)."

According to support, this is due to Drupal 8 using APCu for some types of caching, which has a hardcoded memory limit that is too low.

The "strongly recommended" fix is to add the following to services.yml to tell Drupal to use the database instead of APCu for these types of caching:

````
services:
  cache.backend.chainedfast:
    class: Drupal\Core\Cache\ChainedFastBackendFactory
    arguments: ['@settings', 'cache.backend.database', 'cache.backend.database']
    calls:
      - [setContainer, ['@service_container']]
````

The long-term fix is to use Memcache, but memcache is not currently stable or recommended for D8.

Seems like Bolt should provide services.yml with this config by default.

We're testing this fix right now, and if it works will report back and try to open a PR.

danepowell:
Support recommends that D8 sites add the lines here to services.yml to prevent APCu memory exhaustion.

Unfortunately we've found that this can cause other odd regressions, please hold off on merging for now.

### DIFF
--- a/template/docroot/sites/default/services.yml
+++ b/template/docroot/sites/default/services.yml
@@ -153,11 +153,9 @@ parameters:
     - sftp
     - webcal
     - rtsp
+# Added at the advice of Acquia Support to prevent APCu cache exhaustion. Can
+# be removed if memcache is used instead of APCu.
 services:
-  # Added at the advice of Acquia Support to prevent APCu cache exhaustion. Can
-  # be removed if memcache is used instead of APCu.
   cache.backend.chainedfast:
-    class: Drupal\Core\Cache\ChainedFastBackendFactory
-    arguments: ['@settings', 'cache.backend.database', 'cache.backend.database']
-    calls:
-      - [setContainer, ['@service_container']]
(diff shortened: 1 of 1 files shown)


danepowell:
do not merge this yet... we've discovered it may cause regressions

danepowell:
There's a lot of discussion taking place on how to best address the problem of APCu memory exhaustion without upsetting the chained fast cache backend: https://www.drupal.org/node/2662844

I've updated this PR to include the solution we're currently evaluating on Voya. I'll report back here once it checks out and Support confirms that this is best practice.

danepowell:
This fix has worked for us for a week now, and seems to be the currently-accepted best practice.

PR #47, opened 2016-05-12
Feature: Multiple deployment targets/artifacts for different settings

kylebrowning:
Allows for multiple deployment targets if wanted.

### DIFF
--- a/template/build/core/phing/tasks/deploy.xml
+++ b/template/build/core/phing/tasks/deploy.xml
@@ -28,7 +28,7 @@
     <echo message="Running deployment on ${deploy.remote.name}" />
 
     <!-- Delete the existing deploy directory and re-initialize as an empty git repository. -->
-    <!-- <phingcall target="deploy:artifact:prepare-dir"/> -->
+     <phingcall target="deploy:artifact:prepare-dir"/>
 
     <!-- Add remotes and fetch upstream refs. Checkout local branch tracking
      tracking upstream branch, if it exists. Otherwise create new branch
--- a/template/build/core/phing/tasks/deploy.xml
+++ b/template/build/core/phing/tasks/deploy.xml
@@ -14,9 +14,21 @@
         <echo message="./bolt.sh deploy:artifact -Ddeploy.branch=8.x-build -Ddeploy.commitMsg='BLT-123: The commit message.'"/>
       </then>
     </if>
+    <phingcall target="deploy:artifact:all"/>
+  </target>
+
+  <target name="deploy:artifact:all" description="Sets up env variables for deploy">
+    <foreach list="${deployments}" param="deploy.remote" target="deploy:artifact:single" />
+  </target>
+
+  <target name="deploy:artifact:single" description="Full runtime for all deployment remotes">
+    <property name="deploy.remote.name" value="${${deploy.remote}.name}" override="true"/>
+    <property name="deploy.remote.gits" value="${${deploy.remote}.gits}" override="true"/>
+    <property name="deploy.remote.hosting" value="${${deploy.remote}.hosting}" override="true"/>
+    <echo message="Running deployment on ${deploy.remote.name}" />
 
     <!-- Delete the existing deploy directory and re-initialize as an empty git repository. -->
-    <phingcall target="deploy:artifact:prepare-dir"/>
+    <!-- <phingcall target="deploy:artifact:prepare-dir"/> -->
 
     <!-- Add remotes and fetch upstream refs. Checkout local branch tracking
      tracking upstream branch, if it exists. Otherwise create new branch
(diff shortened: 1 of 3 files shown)


kylebrowning:
Note, this is NOT backwards compatible.

template/build/core/phing/tasks/deploy.xml
```diff
@@ -14,9 +14,21 @@
...
+
+  <target name="deploy:artifact:single" description="Full runtime for all deployment remotes">
+    <property name="deploy.remote.name" value="${${deploy.remote}.name}" override="true"/>
+    <property name="deploy.remote.gits" value="${${deploy.remote}.gits}" override="true"/>
+    <property name="deploy.remote.hosting" value="${${deploy.remote}.hosting}" override="true"/>
+    <echo message="Running deployment on ${deploy.remote.name}" />
 
     <!-- Delete the existing deploy directory and re-initialize as an empty git repository. -->
-    <phingcall target="deploy:artifact:prepare-dir"/>
+    <!-- <phingcall target="deploy:artifact:prepare-dir"/> -->
```
damontgomery: You commented out this target which makes /deploy. Since there is no directory, it cannot find it on line 55.

kylebrowning: Thanks, fixed.

damontgomery: This then fails on `/tests/phpunit/DeployTest.php` line 31,

```
$this->assertFileExists($this->deploy_dir);
```

template/project.yml
```diff
@@ -15,21 +15,30 @@ project:
   # These are random strings used to configure the hash salt.
   hash_salt: ${project.hash_salt}
 
-# Configuration settings for new git repository.
-git:
-  default_branch: master
-  remotes:
-    # Defining git remotes allows builds deployed via CI.
-    - bolt8@svn-5223.devcloud.hosting.acquia.com:bolt8.git
-
```
damontgomery: This test, `tests/phpunit/DeployTest.php` line 61 is looking for the variables you are deleting here and fails,

```
foreach ($this->config['git']['remotes'] as $remote) {
```

template/build/core/phing/tasks/deploy.xml
```diff
@@ -14,9 +14,21 @@
...
+
+  <target name="deploy:artifact:single" description="Full runtime for all deployment remotes">
+    <property name="deploy.remote.name" value="${${deploy.remote}.name}" override="true"/>
+    <property name="deploy.remote.gits" value="${${deploy.remote}.gits}" override="true"/>
+    <property name="deploy.remote.hosting" value="${${deploy.remote}.hosting}" override="true"/>
+    <echo message="Running deployment on ${deploy.remote.name}" />
 
     <!-- Delete the existing deploy directory and re-initialize as an empty git repository. -->
-    <phingcall target="deploy:artifact:prepare-dir"/>
+     <phingcall target="deploy:artifact:prepare-dir"/>
```
damontgomery: You have an extra space ;)

template/build/core/phing/tasks/deploy.xml
```diff
@@ -31,29 +43,29 @@
...
 
     <!-- Create the new branch. -->
     <exec command="git checkout -b ${deploy.branch}" dir="${deploy.dir}" logoutput="true" passthru="true"/>
     <!-- Pull the latest updates (if available). -->
-    <exec command="git merge ${remoteName}/${deploy.branch}" dir="${deploy.dir}" logoutput="true" passthru="true"/>
-  </target>
-
-  <target name="deploy:artifact:add-remotes" description="Add remotes and fetch upstream refs.">
-    <foreach list="${git.remotes}" param="deploy.remote" target="deploy:artifact:add-remote"/>
+    <exec command="git merge ${deploy.remote.hash}/${deploy.branch}" dir="${deploy.dir}" logoutput="true" passthru="true"/>
```
damontgomery: I think you still want to only do this once rather than merge from every remote. I guess it doesn't matter too much, but I'd think that there should be one canonical source. Move this back to `add-remotes`?

template/project.yml
```diff
@@ -15,21 +15,30 @@ project:
...
+
+deployments:
+  - acsf
+  - myhosting
+
+acsf:
+  hosting: acsf
+  gits:
+    - bolt8@svn-5223.devcloud.hosting.acquia.com:bolt8.git
+    - bolt@bolt.com:bolt8.git
```
damontgomery: Should this one be removed? I don't see it above.

template/build/core/phing/tasks/deploy.xml
kylebrowning: Correct me if Im wrong, but different repositories could have different tracks.
Also, this was never in add-remotes... unless Im missing something?

template/project.yml
kylebrowning: I added it as an example, Will change the file to be more example friendly.

template/project.yml
```diff
@@ -15,21 +15,30 @@ project:
...
+  name: Profserv ACSF
+  
+myhosting:
+  hosting: ace
```
damontgomery: I would rename the default remote `default` and leave it uncommented. I would then comment out the ACSF part since that won't be very common.

You may also need to include, `default_branch: master` since that was included in the other section.

kylebrowning: `default_bracnh: master` didn't seem to be used at all.

damontgomery: I'm wondering if the default branch needs to be the same for all. It's what will be created when bolt is initialized and it's the branch that will be checked out I believe when the deploy artifacts are being made.

Used in line 130 of `build.xml` in the `create` task.
```
<exec command="git checkout -b ${git.default_branch}" dir="${bolt.new.dir}" logoutput="true" passthru="true" />
```

damontgomery:
@kylebrowning, I think this looks good!

damontgomery:
@kylebrowning, with all the deployment PRs, I believe this is now out of date.

kylebrowning:
Ill re-roll if we still want to do this.

grasmash:
After a lot of messing about, we've decided to put this on the back burner. A few reasons:

1. The YAML formatting is unintuitive. 
2. The terminology of "ace" is a bit misleading since acp and acft would work with this setting as well.
3. It's unclear that this is a feature that many people would use. If it comes up again, we can revisit. 

In future, the follow format would likely be easiest to implement: 

```
deploy.destinations:
  - ascf,bolt8@svn-5223.devcloud.hosting.acquia.com:bolt8.git
  - ace,bolt8@svn-5224.devcloud.hosting.acquia.com:bolt8.git
```

PR #51, opened 2016-05-16
Fix formatting for multi-line preformatted text.

geerlingguy:
I noticed in the Markdown, the examples are on multiple lines, but when printed via GitHub's Markdown-to-HTML renderer, all the examples wrap on one line/inline.

Using this spacing makes the examples render correctly on GitHub (and should work with most any Markdown parser that renders ol/ul correctly).


geerlingguy:
See, for example: https://github.com/geerlingguy/bolt/blob/58fb5e686f0e4f0119c0df35740d5ea29b255af6/template/build/README.md#setting-up-travis-ci-for-automated-deployments

PR #54, opened 2016-05-17
Error when no front end theme is selected in project.yml

kylebrowning:
This checks if no front end is present and skips accordingly.


grasmash:
Something crazy is going on with these merge commits. Can you please rebase?

PR #57, opened 2016-05-17
Documentation on Drush aliases.

danepowell:


CashWilliams:
looks good.

template/drush/README.md
```diff
@@ -1,3 +1,13 @@
+# Drush configuration and aliases
+
 This directory is intended to contain drush configuration that is not site or environment specific.
 
 Site specific drush configuration lives in `sites/[site-name]`.
+
+## Site aliases
+
+You should add the Drush aliases for any remote environments (such as Acquia Cloud) to the `site-aliases` directory. This allows developers to access remote environments using simple aliases such as `drush @mysite.dev uli`. Note that if you are using Acquia Cloud, developers can also download these aliases manually from their Insight account, but providing them with the project makes everyone’s life a little easier.
```
geerlingguy: I also find it helpful to define a drush alias for the local environment (e.g. `site.local`), as that allows all commands to always be accessed through a familiar syntax/alias structure; also it allows people to have a local alias with differing paths/config more easily.

danepowell: Would the instructions for that basically consist of "Copy example.local.aliases.drushrc.php to local.aliases.drushrc.php and insert the appropriate local values"?

It seems like that's something Bolt could do as part of the setup task, since it already knows your project name. The 'root' would vary for each developer, but Bolt could probably figure out the correct value for that pretty easily. Probably want to open a separate PR for something that extensive.

geerlingguy: Yeah - but right now there's no example.local.aliases.drushrc.php... Drush aliases are already confusing enough :P

damontgomery: I did add this to bolt,
https://github.com/acquia/bolt/blob/8.x/template/drush/site-aliases/example.local.aliases.drushrc.php

damontgomery: So there is sort of docs in https://github.com/acquia/bolt/blob/8.x/template/readme/project-tasks.md#local-tasks, but I think this is useful as well.

@grasmash, do you have recommendations on the best place for this documentation?

grasmash:
Yeah I'd add at least a reference to this fact in the project-tasks.md file to answer the question "How do I add drush aliases to the project?"

PR #59, opened 2016-05-18
Revert "Removing branch argument from deploy tasks."

damontgomery:
Reverts acquia/bolt#42

Having issues where Travis cannot run commands to find the current branch. It only checks out a hash.


damontgomery:
We may not want a full revert, but rather a middle ground where the Phing target requires the source branch, but the travis argument still passes it a branch (instead of two).

grasmash:
@damontgomery Is this still necessary after merging https://github.com/acquia/bolt/pull/58 ? There is currently a merge conflict.

damontgomery:
This is no longer relevant. It was an alternative to #58

PR #61, opened 2016-05-19
Run automated tests on live dbs.

danepowell:
This provides documentation on how to run functional tests on update hooks and a live environment. It also forces the exec tasks in the `local` Phing tasks to fail the build on errors.


danepowell:
I think there's a problem with automated tests, they shouldn't be failing here.

grasmash:
Rerunning tests.

PR #63, opened 2016-05-20
Using config dir from $config_directories for site install.

grasmash:


danepowell:
This seems like it could break things. I agree that we shouldn't hardcode the `config/default` directory, but there's a clear distinction between the `vcs` and `sync` config directories, and it's not going to help anything by hardcoding this to the `sync` directory.

For most of the config management workflows I've seen, it should use `vcs` if it uses anything at all.

grasmash:
@danepowell Agreed. This didn't fix the intended issue anyway :p

PR #68, opened 2016-05-20
Make vendor name configurable.

greylabel:


greylabel:
@grasmash -- can you kick Travis to test this again?

grasmash:
I'm tempted to just replace the "acquia" in the template/composer.json with "your-organization". It's the only place that the variable would be used, so should we really make it a variable?

grasmash:
Test has been kicked off again.

greylabel:
@grasmash I agree this is a pretty trivial change, on the surface. I think the crux of the issue is whether we leverage a single file to hold project-specific config (`project.yml`), or if we are ok with config/variables/etc being spread throughout the BLT code and potentially undocumented. Personally, I would favor a more robust `project.yml` rather than need to make project-specific changes throughout the codebase.

PR #69, opened 2016-05-23
Fixed config import on site installs.

danepowell:
When I first wrote the site-install targets, we were still experimenting with various forms of CMI, and the site install targets really only supported full config imports or nothing at all.

In practice, I suspect most projects would want to use partial config imports, not full config imports, which are very restrictive. This is certainly the case for us.


danepowell:
Note that if you don't use the config/default directory at all, this change won't affect you... `drush config-import --partial` will just exit harmlessly.

PR #75, opened 2016-05-24
Fixing deployments

grasmash:


grasmash:
For some reason, running `./vendor/bin/phpunit tests/phpunit --exclude-group deploypush` isn't actually excluding tests tagged with deploypush!

Issue #85, opened 2016-05-26
Enable local Twig Debugging


dpagini:
It seems like by default, local twig debugging is not turned on due to an invalid path to _development.services.yml_ file.

https://github.com/acquia/bolt/blob/8.x/template/docroot/sites/default/settings/default.local.settings.php#L33

I think this file should read:
`$settings['container_yamls'][] = $dir . '/docroot/sites/development.services.yml';`
_Note the additional "/docroot" which is not part of the $dir variable._

damontgomery:
Thanks @dpagini! Since the PR was merged, @grasmash, do you want to close this issue?

PR #86, opened 2016-05-26
Enable local Twig debugging

dpagini:
Addresses #85 

The below lines show the value of $dir being the project root, and not the docroot folder. Maybe this is a problem with the DRUPAL_ROOT value?
`$dir = dirname(DRUPAL_ROOT);`
`$config_directories['sync'] = $dir . '/config/default';`

This PR points to the development.services.yml file in the correct location.

### DIFF
--- a/template/docroot/sites/default/settings/default.local.settings.php
+++ b/template/docroot/sites/default/settings/default.local.settings.php
@@ -30,7 +30,7 @@ $dir = dirname(DRUPAL_ROOT);
 $config_directories['sync'] = $dir . '/config/default';
 
 // Use development service parameters.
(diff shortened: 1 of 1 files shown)


dpagini:
Please be gentle, I think this is my first public repo PR. Any pointers are appreciated.

grasmash:
@dpagini Thanks for catching this. The PR looks good.

`$dir` should reference the directory above the docroot, so that's working as expected. Perhaps we should rename `$dir` to `$repo_root`, which is the terminology used elsewhere in the project.

dpagini:
Thank you @grasmash, and thanks to @danepowell for helping me identify the issue on our project which is utilizing Bolt, a very cool tool.

PR #89, opened 2016-05-31
Check return on deploy targets.

danepowell:
Currently Travis may pass even if deploys fail because Phing doesn't check the return value of deploy steps.


grasmash:
Yeah unfortunately deployments are actually failing for this repo, which is being addressed in #75

danepowell:
Ah thanks, I thought there was already a deploy-related PR out there to fix this but I couldn't find it. I guess the title threw me off.

grasmash:
These changes are in #75, so I'm closing this.

PR #92, opened 2016-05-31
Add missing cweagans/composer-patches requirement.

geerlingguy:
...without this, none of the patches defined in `composer.json` are applied.

### DIFF
--- a/template/composer.json
+++ b/template/composer.json
@@ -11,6 +11,7 @@
   ],
   "require": {
     "composer/installers":                        "^1.0.20",
+    "cweagans/composer-patches":                  "~1.0",
     "drupal/acquia_connector":                    "8.1.*",
     "drupal/acsf":                                "~8.1",
     "drupal/core":                                "~8",


grasmash:
Need to update composer.lock as well.

geerlingguy:
Will do.

geerlingguy:
Er... wrong lock file :/ Updating.

grasmash:
Hm, this is why I was hesitant to remove the lock file. Lightning is failing upstream, and our builds are pulling in the latest version. So, this build fails.

grasmash:
Not sure why this keeps failing... but it keeps failing.

geerlingguy:
@grasmash - It seems to always be something in Lightning...?

grasmash:
Yeah, this makes me wonder if somehow it's affecting the patches that lightning is applying.

danepowell:
I don't think there was anything wrong with this PR, Travis tests were just a little flaky a few weeks ago. This should be resolved by #191.

PR #100, opened 2016-06-02
Added support script for Bamboo CI and added documentation.

damontgomery:


grasmash:
Rather than dropping this script in the repo root, I'd prefer to seclude the bamboo scripts in `template/scripts/bamboo` and add a new task to `ci.xml` called `ci:bamboo:init` that sets up the repo for bamboo. E.g., copies files to the correct locations. 

This is the approach that I recently took with tugboat, and I would like to refactor Travis CI support to work in the same way.

damontgomery:
Sounds good, I'll take a look!

PR #105, opened 2016-06-03
Adding Drupal VM 3.1 integration.

grasmash:


grasmash:
Need to wait for Drupal VM 3.1.0. Will update version in composer.json, update docs.

template/scripts/drupal-vm/config.yml
```diff
@@ -0,0 +1,29 @@
...
+# If you build the makefile using BLT's built-in Phing task (recommended),
+# set `build_makefile` to `false`.
+build_makefile: false
+
+drupal_mysql_user: drupal
+drupal_mysql_password: ''
+drupal_mysql_database: drupal
+
+extra_packages:
+  - unzip
```
oxyc: This is no longer required, Drupal VM added it as a hard dependency https://github.com/geerlingguy/drupal-vm/pull/551.

template/scripts/drupal-vm/drupal-vm.drushrc.aliases.php
```diff
@@ -0,0 +1,13 @@
...
+// [vagrant_machine_name].local
+$aliases['${project.machine_name}.local'] = array(
+  // /var/www/[vagrant_machine_name]/docroot
+  'root' => '/var/www/${project.machine_name}/docroot',
+  // vagrant_hostname
+  'uri' => '${project.local_uri}',
+  // vagrant_hostname
+  'remote-host' => '${project.local_uri}',
+  'remote-user' => 'vagrant',
+  'ssh-options' => '-i ~/.vagrant.d/insecure_private_key',
```
oxyc: this doesn't work for `core-rsync` see https://github.com/geerlingguy/drupal-vm/pull/552.

template/.gitignore
```diff
@@ -105,3 +105,5 @@ vendor/.git
 nbproject
 nbproject/*
 
+# DrupalVM
+.vagrant/
```
oxyc: Drupal VM has two user specific configs as well: `local.config.yml` and `Vagrantfile.local`.

grasmash: Thanks. This should be handled by `local.*` and `*.local`

template/scripts/drupal-vm/config.yml
```diff
@@ -0,0 +1,26 @@
...
+# Set drupal_core_path to the `destination` in the synced folder
+# configuration above, plus `/docroot`.
+drupal_core_path: "/var/www/${project.machine_name}/docroot"
+
+# If you build the makefile using BLT's built-in Phing task (recommended),
+# set `build_makefile` to `false`.
+build_makefile: false
+
+drupal_mysql_user: root
+drupal_mysql_password: ''
```
grasmash: I suspect these mysql settings may cause a problem. Perhaps if this user already exists with a password. @geerlingguy would you happen to know?

geerlingguy:
Drupal VM 3.1.0 was tagged and released today: https://github.com/geerlingguy/drupal-vm/releases/tag/3.1.0

The release includes Solr 5.x as a default, and you can now just write variables you want to override to `config.yml` (and then developers can override with a gitignored `local.config.yml` file). No need to copy the entire `default.config.yml` anymore.

Finally, Drupal VM itself now builds Composer projects and/or composer.json files (this should be disabled for BLT projects, since it's handled by BLT itself via Phing, of course).

I'm going to see if some of the Composer performance optimizations help any with BLT-generated projects, too. I hate idling while waiting for `composer update` :P

compiling

template/scripts/drupal-vm/drupal-vm.drushrc.aliases.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+// [vagrant_machine_name].local
+$aliases['${project.machine_name}.local'] = array(
```
CashWilliams: small and picky of me, but why hard code to `.local`?

CashWilliams: make sense to use `${project.local_uri}` instead?

geerlingguy: That's often an annoyingly long string, though... it's not a huge deal to me either way, but I like [machinename].local as a convention, and then developers can technically override their local site URLs and use whatever they'd like.

Some devs like "mysitename.local", others "www.mysitename.dev", others "local.mysitename.com", etc.

Having this particular string separate from the URI gives devs a lot more flexibility, imo.

template/build/core/phing/tasks/vm.xml
```diff
@@ -0,0 +1,21 @@
+<project name="vm" default="vm:init">
+
+  <target name="vm:init" description="Initializes Drupal VM for this project.">
+    <copy file="${repo.root}/scripts/drupal-vm/drupal-vm.drushrc.aliases.php" todir="${repo.root}/drush/site-aliases">
+      <filterchain>
+        <expandproperties />
+      </filterchain>
+    </copy>
+    <mkdir dir="${repo.root}/box" mode="755"/>
+    <exec dir="${repo.root}" command="composer require geerlingguy/drupal-vm:^3.1" logoutput="true" checkreturn="true"/>
```
oxyc: Should probably be a dev dependency.

template/scripts/drupal-vm/drupal-vm.drushrc.aliases.php
geerlingguy: (Plus it follows the convention for the other environments, of [machine].local, .dev, .prod, etc.)

template/scripts/drupal-vm/config.yml
```diff
@@ -0,0 +1,26 @@
...
+# Set drupal_site_name to the project's human-readable name.
+drupal_site_name: "${project.human_name}"
+
+# Provide the path to the project root to Vagrant.
+vagrant_synced_folders:
+  # Set the local_path for the first synced folder to `../`.
+  - local_path: ../
+    # Set the destination to the Acquia Cloud subscription machine name.
+    destination: ${project.machine_name}
+    type: nfs
```
oxyc: Has this been tested? I believe `local_path` should be kept as `.` and destination should be `/var/www/${project.machine_name}`

template/scripts/drupal-vm/drupal-vm.drushrc.aliases.php
grasmash: Aaaand local_uri doesn't actually exist. Gotta add that.

template/scripts/drupal-vm/config.yml
grasmash: Yes this has been tested. I did initially attempt to keep it as `.` and it did not work. I'll double check the value of the destination.

template/scripts/drupal-vm/config.yml
```diff
@@ -0,0 +1,26 @@
...
+
+# Set drupal_core_path to the `destination` in the synced folder
+# configuration above, plus `/docroot`.
+drupal_core_path: "/var/www/${project.machine_name}/docroot"
+
+# If you build the makefile using BLT's built-in Phing task (recommended),
+# set `build_makefile` to `false`.
+build_makefile: false
```
oxyc: As of 3.1.0: I guess Drupal VM doesn't need to run `composer install` in this case so these two should be:

```yaml
drupal_core_path: "/var/www/${project.machine_name}/docroot"
build_composer_project: false
```

If you wanted to be more explicit the full working configuration with `composer install` running during `vagrant up`would be:

```yaml
build_composer_project: false
# toggling this to `true` would invoke `composer install` with the
# projects own `composer.json` successfully.
build_composer: false
drupal_composer_path: false
drupal_composer_install_dir: "/var/www/${project.machine_name}"
drupal_core_path: "{{ drupal_composer_install_dir }}/docroot"
```

Docs: http://docs.drupalvm.com/en/latest/deployment/composer/#using-composer-when-drupal-vm-is-a-composer-dependency-itself

template/readme/local-development.md
```diff
@@ -27,78 +27,8 @@ If you must use Windows, we recommend that:
...
-      'uri' => '[vagrant_hostname]',
-      // vagrant_hostname
-      'remote-host' => '[vagrant_hostname]',
-      'remote-user' => 'vagrant',
-      'ssh-options' => '-i ~/.vagrant.d/insecure_private_key',
-    );
-    ```
-1. Updating `project.yml` so that `drush.default_alias` is set to your Drupal VM alias.
+1. Execute `./blt.sh vm:init` from the project root directory.
 1. Follow the Quick Start Guide in [Drupal VM's README](https://github.com/geerlingguy/drupal-vm#quick-start-guide)
```
oxyc: `2.` :)

geerlingguy: Technically it can always be `1` (or any number) since Markdown parses it as a typical `ol` item in the resulting HTML... I like numbering them in the markdown itself because I usually read Markdown in its native format... but many people like using `1 1 1 1 1 1` since it requires renumbering if you're off (well, technically you could have any number there, but I'm a little OCD, so I always renumber when editing) :D

template/scripts/drupal-vm/config.yml
```diff
@@ -0,0 +1,30 @@
+# Update the hostname to the local development environment hostname
+vagrant_hostname: ${project.machine_name}.localhost
```
CashWilliams: For some reason I feel like this should be ${project.machine_name}.vm rather then .localhost

geerlingguy: It really should be `${project.local_uri}`—see https://github.com/acquia/blt/pull/113/files#diff-1d79176dc01e3cdc73b4066200be0e02R2

grasmash: true.

geerlingguy:
Note that the default Drupal VM config installs Ubuntu 16.04 and PHP 7.0.x, and this seems to make Lightning's `drush site-install` task fail:

```
project > setup:drupal:install:

    [drush] Executing '/Users/jeff.geerling/Sites/acquia/blted8/vendor/bin/drush @self --site-name="BLTed 8" --site-mail="no-reply@acquia.com" --account-name="admin" --account-pass="admin" --account-mail="no-reply@acquia.com" --l="default" --nocolor --yes --verbose site-install "lightning" "install_configure_form.update_status_module='array(FALSE,FALSE)'"'...
Loaded alias @self from file                                                                                    [notice]
/Users/jeff.geerling/Sites/acquia/blted8/drush/site-aliases/drupal-vm.aliases.drushrc.php
Begin redispatch via drush_invoke_process().                                                                    [notice]
Calling proc_open(ssh -o PasswordAuthentication=no -i /Users/jeff.geerling/.vagrant.d/insecure_private_key vagrant@local.testblt.com 'env COLUMNS=120 drush  --local --alias-path=/Users/jeff.geerling/Sites/acquia/blted8/drush/site-aliases --nocolor --yes --verbose --uri=default --root=/var/www/blted8/docroot  site-install lightning '\''install_configure_form.update_status_module='\''\'\'''\''array(FALSE,FALSE)'\''\'\'''\'''\''   --site-name='\''BLTed 8'\'' --site-mail='\''no-reply@acquia.com'\'' --account-name=admin --account-pass=admin --account-mail='\''no-reply@acquia.com'\'' 2>&1' 2>&1);
End redispatch via drush_invoke_process().                                                                      [notice]
    [drush] Warning: Permanently added the ECDSA host key for IP address '192.168.88.88' to the list of known hosts.
    [drush] Loaded alias @self from file /var/www/blted8/docroot/../drush/site-aliases/drupal-vm.aliases.drushrc.php        [notice]
    [drush] Executing: mysql --defaults-extra-file=/tmp/drush_U2aJQU --database=drupal --host=localhost --port=3306 --silent  < /tmp/drush_cem0wJ > /dev/null
    [drush] You are about to create a /settings.php file and create a sites/sites.php file and DROP all tables in your 'drupal' database. Do you want to continue? (y/n): y
    [drush] bash: line 1:  7470 Segmentation fault      env COLUMNS=120 drush --local --alias-path=/Users/jeff.geerling/Sites/acquia/blted8/drush/site-aliases --nocolor --yes --verbose --uri=default --root=/var/www/blted8/docroot site-install lightning 'install_configure_form.update_status_module='\''array(FALSE,FALSE)'\''' --site-name='BLTed 8' --site-mail='no-reply@acquia.com' --account-name=admin --account-pass=admin --account-mail='no-reply@acquia.com' 2>&1
[phingcall] /Users/jeff.geerling/Sites/acquia/blted8/build/core/phing/tasks/setup.xml:73:85: Drush exited with code 139

BUILD FAILED
/Users/jeff.geerling/Sites/acquia/blted8/build/core/phing/tasks/local-sync.xml:12:30: Execution of the target buildfile failed. Aborting.

Total time: 49.0135 seconds
```

Debugging further, it seems the site installation works fine via Drupal's UI installer (though there are a bunch of entity errors on first page load after install in Lightning), and site installation also works fine using `standard` or `minimal` as the install profile.

CashWilliams:
I think the way DrupalVM is pulled in to `vendor` here makes it such that the VagrantFile (inside DrupalVM) isn't finding a VagrantFile.local override. @geerlingguy would need to confirm though.

oxyc:
@CashWilliams it looks for it in `box/Vagrantfile.local`.

CashWilliams:
@oxyc thanks, that kind makes sense, I'll test it out.

CashWilliams:
With the latest commits, this is now building and installing for me. 👍 

However there is still an issue with drush aliases:
  - Multiple versions are created everywhere. DrupalVM creates some, blt creates some, and there seems to be a few more laying around.
  - The best alias I can see to use only works from Host->VM. Logging into the VM, the alias does not work b/c it is trying to reference a remote.

dpagini:
Can I ask a quick question on this PR, @grasmash? Is the intent that the individual setting up the repo would run this `vm:init` command once, and then commit the `Vagrantfile` (and others)? Or would each individual developer that wants to use Drupal VM run the `vm:init` command?
I only ask b/c I'm wondering if we need more in the .gitignore file? Such as a "box" folder? Or would the  intent be that the box folder gets committed to the repo?

grasmash:
The intent is that after generating a project, someone would run `vm:init` once and then commit the generated files. I suppose that should be documented.

PR #112, opened 2016-06-06
Support ACSF vanity domains.

danepowell:
Many projects will have vanity domains for environments, such as `dev.example.com` instead of `example.dev-project.acsitefactory.com`. The current ACSF alias template doesn't support these vanity domains, which causes confusion at best and blockers/bugs at worst (if you are using trusted host patterns or have a temperamental cache).

This supports adding vanity domains, and also provides backwards compatibility when vanity domains don't exist.


danepowell:
@damontgomery and @kylebrowning I think you wrote this alias template originally if you want to comment. You can see it in action on the project you've been helping me on.

PR #116, opened 2016-06-07
Re-architecture to make things faster

CashWilliams:
Moving some things around to try to make things build faster and improve DX.

`blt create` is faster b/c composer install is removed (see below). This way a user can modify composer.json in the target directory before the 1st run of composer install, and allows for faster BLT development itself.

This changes DX to 
- git clone [blt repo]
- ./blt.sh configure && ./blt.sh create
- cd [target dir]
- ./blt.sh [local:]setup

composer install is auto detected and run if needed.


build/custom/phing/build.xml
```diff
@@ -3,7 +3,7 @@
 <project name="blt" default="list">
 
   <!-- Define blt root directory. -->
-  <property name="blt.root.relative" value="${project.basedir}/../.."/>
+  <property name="blt.root.relative" value="${project.basedir}/../../.."/>
```
grasmash: Why has this been moved into a custom dir? I don't anticipate a case where someone would add custom targets here. If we must use the core/custom paradigm here, this file would belong in build/core.

build/custom/phing/build.xml
```diff
@@ -161,6 +160,13 @@
     <phingcall target="tests:phpunit" />
   </target>
 
+  <!-- This target is meant for developing BLT itself. -->
```
grasmash: The "new-project" target does exactly this.

build/custom/phing/build.xml
CashWilliams: I moved it to allow for ./blt.sh in both places to reference the same path

build/custom/phing/build.xml
CashWilliams: new-project does not clean

build/custom/phing/build.xml
CashWilliams: Right now there are 2 instances of blt.sh which are the same minus the path they call. This means duplicated efforts when making changes to the script. This changes them to both rely on the same patch structure, and the initial create copies the existing blt.sh into the target directory.

CashWilliams:
closing this and breaking it up

PR #117, opened 2016-06-07
Catch PHP messages with Behat.

danepowell:
This will cause Behat to fail on PHP errors, warnings, and notices as long as watchdog is enabled.

### DIFF
--- a/template/tests/behat/features/bootstrap/Drupal/FeatureContext.php
+++ b/template/tests/behat/features/bootstrap/Drupal/FeatureContext.php
@@ -21,4 +21,38 @@ class FeatureContext extends RawDrupalContext implements SnippetAcceptingContext
 
   }
 
+  /**
+   * The BeforeSuite hook is run before any feature in the suite runs.
+   *
+   * @BeforeSuite
+   */
+  public static function prepare($event) {
+    if (db_table_exists('watchdog')) {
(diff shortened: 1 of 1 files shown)


grasmash:
Looks like lightning has some warnings that it generates. Perhaps it would be good to commit this to acquia/lightning?

danepowell:
That sucks. Ideally we wouldn't test upstream dependencies at all... as we've seen here and with the other Behat tests, we can't really rely on them to pass. But I guess committing this to Lightning and fixing the notices there is the next best option.

Although I'm not exactly sure how Behat works... if we add this to Lightning's FeatureContext, will it also catch errors in custom modules that projects create outside the profile?

danepowell:
I'm working on fixing this upstream in Lightning first: https://github.com/acquia/lightning/pull/125

... although I just found that @grasmash beat me to the punch and probably forgot about it :) https://github.com/acquia/lightning/pull/103

Anyway, after we get these errors fixed and automated testing in place in Lightning, then we can see if we need to add them to Bolt as well. I'm not sure if contexts get inherited or not.

danepowell:
Happy to announce that this is now a part of Lightning! https://github.com/acquia/lightning/pull/225

If we want to include it in BLT as well, we can follow the same pattern.

PR #123, opened 2016-06-08
Make phpcs test first test

CashWilliams:
Moves phpcs to the first test that is run. If phpcs does not pass, there is no sense in doing a full install.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,8 @@ before_script:
   - git diff --exit-code
 
 script:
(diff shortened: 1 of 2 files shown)


CashWilliams:
This also changes where the coder standard lives, from the target to blt itself.

CashWilliams:
no travis test? 😢

grasmash:
Yeah that's weird. May need to amend and re-push.

grasmash:
Closed in favor of rebase #127

PR #124, opened 2016-06-08
DX improvement by adding composer install to blt.sh

CashWilliams:
This moves composer install to the blt.sh wrapper.

This allows an easier initial setup (don't have to run composer install after clone), and removes it from the inital `create` task. This makes `create` much faster by delaying composer install to the first run of blt.sh within the target directory.


grasmash:
Close in favor of #125, which has been rebased.

Issue #139, opened 2016-06-09
project.acquia_subname should be defined or otherwise replaced in project.yml


greylabel:
Below is a snippet from the `composer.json` file for a BLT generated project using the default `project.yml`. The default `project.yml` should probably include an addition variable for `project.description` that can be used in place of project.acquia_subname, which I don't see used elsewhere in the project.

```
{
  "name":    "acquia/blted8",
  "license": "proprietary",
  "type":    "project",
  "description": "${project.acquia_subname}",
  "repositories": [
    {
      "type": "composer",
      "url":  "https://packagist.drupal-composer.org"
    }
```

greylabel:
This was fixed.

PR #140, opened 2016-06-09
Make project description configurable.

greylabel:
https://github.com/acquia/blt/issues/139

### DIFF
--- a/template/project.yml
+++ b/template/project.yml
@@ -1,6 +1,8 @@
 project:
-  # This will be used as the vendor name in the new project's composer.json file.
+  # The project's composer.json will be updated with the below vendor_name and description
+  # and should be updated directly to change these values after initial creation.
   vendor_name: 'acquia'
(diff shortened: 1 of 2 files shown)


template/project.yml
```diff
@@ -1,6 +1,8 @@
 project:
   # This will be used as the vendor name in the new project's composer.json file.
   vendor_name: 'acquia'
+  # This will be used as the description in the new project's composer.json file.
+  description: 'A short description of the package. Usually this is just one line long.'
```
grasmash: After the new project is created, this value will still be in the project.yml but it won't do anything. If someone changes the composer.json file, it will diverge from the value here. We've got the same issue with vendor_name. 

We could change the comments to be less about project generation and simply allow these values to be here for anyone reading the file. 

Thoughts?

greylabel: Something like this `The project's composer.json will contain this value and should be updated directly to change this value after initial project creation`?

PR #141, opened 2016-06-10
Created common settings includes file.

danepowell:
This moves `settings.php` and `services.yml` a little closer to their stock / default values by moving all of the includes to a common `includes.settings.php`. There's a few advantages to this.

1. Better ACE / ACSF cooperation. Previously, we had to maintain two entirely separate sets of settings files, one for ACSF (`factory-hooks/pre-settings-php/*`) and one for ACE (`docroot/sites/default/settings/*`). This is extremely error-prone if you are developing for both ACE and ACSF simultaneously. This PR harmonizes them, so that both ACE and ACSF only have to include a single common file (`includes.settings.php`) that will pull in all other settings.
1. Better Bolt updates. If people don't have to customize `settings.php` or `services.yml`, it's easier for them to pull in upstream changes to these files.


grasmash:
We should also update BLT's https://github.com/acquia/blt/blob/8.x/template/scripts/blt/update-scaffold#L7 to update these files.

greylabel:
Maybe not in the scope of this PR, but I think it makes sense to move all common settings files to `sites/all/settings` for the purpose of making BTL more multisite compatible. The convention of common settings residing in `sites/all/settings` follows more closely the convention how Drupal handles shared themes, modules, libraries, etc.  Obviously includes from `sites/all/settings` will function correctly, but feel we should treat the `default` site as any other site in a multisite and not have it contain platform level config, but only its config.

grasmash:
@greylabel we did initially have things set up that way. I don't clearly remember why we switch to placing the files in sites/default/settings, but I believe it was primarily that people couldn't find the settings files. Most people expect the settings to be under sites/default.

greylabel:
@grasmash Are you open to switching back to that configuration. I find it to be much more sane when the number of sites scales and it keeps the `default` site isolated -- sharing amongst site folder is more confusing to me personally.

danepowell:
At least within PS, it's rare that we launch new multisite projects on ACE any more. Multisite almost always goes on ACSF, which doesn't use either `all` or `default` anyway. So to me it actually makes sense to keep using `default`, since that aligns with the most common use case.

I think you could make a good argument either way, but I can tell you we've already flipped back and forth between `all` and `default` at least twice. I don't particularly want to do it again (and force existing users of Bolt to re-learn it) unless there's a pretty compelling reason. It just feels like unnecessary churn at this point.

greylabel:
@danepowell @grasmash -- since this PR is closed, I created https://github.com/acquia/blt/issues/147 to continue this discussion. Can you add your above comment there?

PR #158, opened 2016-06-15
Allow project.local.yml for different local environments.

damontgomery:
- Ignore local files, including sites, settings, project.
- Load project.local.yml
- Example project.local.yml
- Update docs references to project.yml.local

### DIFF
--- a/template/.gitignore
+++ b/template/.gitignore
@@ -1,6 +1,6 @@
 # Ignore configuration files that may contain sensitive information.
 local.*
-*.local
+*.local.*
 
 # Ignore drupal core.
 docroot/core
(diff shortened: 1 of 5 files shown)


template/.gitignore
```diff
@@ -1,6 +1,9 @@
 # Ignore configuration files that may contain sensitive information.
 local.*
 *.local
+project.local.yml
```
grasmash: I wonder if we should just replace these three lines with *.local.* ? Also, where do settings.local.yml and sites.local.yml come into play? I don't see them in this PR.

damontgomery: Yeah, my bad. My copy & paste skills are lacking.

template/.gitignore
```diff
@@ -1,6 +1,6 @@
 # Ignore configuration files that may contain sensitive information.
 local.*
-*.local
+*.local.*
```
damontgomery: @grasmash, per your recommendation!

PR #159, opened 2016-06-15
Add more placeholders to the behat example local configuration.

damontgomery:


template/tests/behat/example.local.yml
```diff
@@ -54,15 +54,15 @@ local:
         # This must be an absolute path.
         drupal_root: ${docroot}
       drush:
-        alias: '@self'
+        alias: alias: ${drush.aliases.local}
```
grasmash: Looks like there's an extra "alias:" here.

damontgomery:
My bad... updated.

template/tests/behat/example.local.yml
```diff
@@ -54,15 +54,15 @@ local:
...
+        alias: ${drush.aliases.local}
 
 example-multisite:
   extensions:
     Behat\MinkExtension:
       javascript_session: selenium2
       # Configure the base url for your site. For multisite, ensure that this
       # corresponds to a valid domain entry in sites.php.
-      base_url: http://127.0.0.1:8889
+      base_url: ${project.local.uri}
```
grasmash: This actually needs to be different than the local uri, since it's a multisite. It would needs it's own uri, hence the 8889 port rather than 8888.

damontgomery: So... `uri: ${project.local.protocol}://${project.local.hostname}` sets the uri to something like http://example.dev. I'm unclear why this wouldn't work? I used this sort of pattern for my multisite install with a sites.php line like `$sites['example.dev'] = 'example.dev';`.

Can you clarify?

Along with #158, I use a local file to set the per-local environment I'm using for dev.

grasmash: This example config is for the second site in a multisite configuration, so it cannot have exactly the same URL as the first site in the multisite configuration. This particular value allows us to start up two different sites on different ports when using `drush runserver` on a  CI machine.

PR #162, opened 2016-06-15
Changing deploy to use rsync.

grasmash:


template/build/core/phing/tasks/deploy.xml
```diff
@@ -118,47 +113,7 @@
...
-    <!-- Copy configuration. -->
-    <copy todir="${deploy.dir}/config" overwrite="true">
-      <fileset dir="${repo.root}/config"/>
-    </copy>
-
-    <!-- Copy local patches. -->
-    <copy todir="${deploy.dir}/patches" overwrite="true">
-      <fileset dir="${repo.root}/patches"/>
-    </copy>
+    <exec dir="${repo.root}" command="rsync -a --no-g --no-p --delete ${repo.root} ${deploy.dir}" logoutput="true" checkreturn="true"/>
```
geerlingguy: This looks _slightly_ simpler than what we had before.

grasmash:
It's also about 2x faster.

template/build/core/files/deploy-exclude.txt
```diff
@@ -0,0 +1,25 @@
+# These files and their children will not be copied to the deployment artifact.
+# If these files already exist in the upstream artifact, they will be removed.
+.drush-use
+.git
+.gitignore
+.travis.yml
+.vagrant
+blt.sh
+bower_components
```
grasmash: Sadly I don't think applies to subdirs, and I haven't had much luck using patterns in the exclude file.

danepowell:
It looks like this will deploy drush aliases in `drush/site-aliases`, which is a big win for ACSF.

PR #163, opened 2016-06-15
Allowing flag to be used to delete existing dir.

grasmash:


build/phing/build.xml
```diff
@@ -65,15 +65,17 @@
 
     <if>
       <and>
+        <not><equals arg1="y" arg2="${do.delete}"/></not>
         <available file="${blt.new.dir}" type="dir" />
```
grasmash: This likely needs to be wrapped in <not>

PR #167, opened 2016-06-15
Removing :artifact from deploy targets.

grasmash:


grasmash:
This will require people to modify their .travis.yml files to call the new target name.

PR #148, opened 2016-06-14
Drupal composer source

grasmash:

PR #173, opened 2016-06-16
Updated update scaffold.

danepowell:
I sense that keeping this script updated is going to be a sisyphean task. We're going to need a better solution before too long.


grasmash:
@danepowell  I've replaced this with the drupal-scaffold project here: https://github.com/acquia/blt/pull/148

PR #178, opened 2016-06-17
Minor Typo fix and updating default selenium port

justinlevi:
Setting the default selenium port to 4444. This is the default port that Drupal-VM ships with so it gets the tests running a bit faster


grasmash:
It looks like this PR also removed the `browser: chrome` setting.

justinlevi:
Do we need the browser setting? Does behat choose a default if you leave that blank?

grasmash:
I'm not sure what the browser setting will be by default, but I'd like it to be chrome. :)

grasmash:
@justinlevi If you remove the chrome change, I will merge.

justinlevi:
Done.

PR #179, opened 2016-06-17
Added file path configuration.

danepowell:
ACSF doesn't have any documentation on how to configure private files. This is the solution that Rok endorsed. The public documentation should be updated to reflect this soon.

### DIFF
--- a/template/docroot/sites/default/settings/files.settings.php
+++ b/template/docroot/sites/default/settings/files.settings.php
@@ -1,0 +1,12 @@
+<?php
+
+/**
+ * @file
+ * Contains Files configuration.
+ */
+if ($is_acsf) {
+  $settings['file_private_path'] = "/mnt/gfs/$ah_group.$ah_env/files-private/$acsf_db_name";
+}
+else {
+  $settings['file_private_path'] = "/mnt/files/$ah_group.$ah_env/files-private";
+}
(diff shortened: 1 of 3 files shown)


template/docroot/sites/default/settings/files.settings.php
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * @file
+ * Contains Files configuration.
+ */
+if ($is_acsf) {
+  $settings['file_private_path'] = "/mnt/gfs/$ah_group.$ah_env/files-private/$acsf_db_name";
+}
+else {
+  $settings['file_private_path'] = "/mnt/files/$ah_group.$ah_env/files-private";
```
greylabel: I think this is another place where we should consider multisite. D7 version would look something like `/mnt/files/' . $_ENV['AH_SITE_NAME'] . '/' . conf_path() . '/files-private'`, need to test the exact D8 syntax, but https://www.drupal.org/node/2275139 has some background since conf_path() is deprecated.

danepowell: We can use `$site_path` instead of `$conf_path` (the ACE settings files provide this variable.)

The bigger problem is how does this file know if we area single-site or multisite install? We probably need a parameter in `project.yml`, as well as a way to propagate that variable into settings files.

kylebrowning: I think the latter is a safer assumption, but I do love being explicit.

greylabel: > Or I guess we could just treat every site as multisite. I don't know if there's any harm in that.

I feel this is the right way to go. 

If we can avoid using Acquia specific variables like `$ah_group.$ah_env` and use Drupal API functions like `conf_path()` instead this code will be a lot more portable and work on local, CI, as well as Acquia.

grasmash:
Closing in favor of #359.

PR #190, opened 2016-06-21
Updating composer to use d.o packagist.

grasmash:


grasmash:
Seems like this doesn't download lightning.

geerlingguy:
@grasmash - I know there is an issue with distributions and composer on Drupal.org: [Give installation profiles/distributions GPLv3+ license as option for packaged downloads](https://www.drupal.org/node/1449452).

grasmash:
Lightning is being downloaded, but it does not include the drupal.org make file.

grasmash:
This is failing because distros are not yet supported by d.o packagist: https://www.drupal.org/node/2718229

downloading drupal/lighting actually downloads a full D8 docroot to docroot/profiles/contrib/lightning.

PR #191, opened 2016-06-21
Support composer patches.

danepowell:


grasmash:
Looks a lot like #92, which is failing for some reason.

danepowell:
Yeah, my theory is that #92 was failing for some reason unrelated to the PR itself. I know Travis tests were kind of flaky until pretty recently.

PR #195, opened 2016-06-21
Adding support for custom commands in frontend and setup targets.

grasmash:


template/project.yml
```diff
@@ -40,6 +40,24 @@ drush:
...
 
+# Custom tasks that are triggered at pre-defined times in the build process.
+# Available keys are frontend-build, post-deploy-build, and post-setup-build.
+tasks:
+  # Executed when front end assets should be generated.
+  frontend-build:
+    # E.g., ${docroot}/sites/all/themes/custom/mytheme.
+    dir: ${docroot}
+    # E.g., `npm install` or `bower install`.
+    command: echo 'Hello world!'
```
CashWilliams: I'm not crazy about "Hello world" here.

I think it could be confusing to see it go past during a build, out of context.

I'd rather see something like `echo 'Frontend build not configured.`

grasmash: Good idea.

template/project.yml
```diff
@@ -40,6 +40,24 @@ drush:
...
+  # Executed after deployment artifact is created.
+  post-deploy-build:
+    dir: ${deploy.dir}/docroot/profiles/contrib/lightning
```
CashWilliams: Wouldn't this be a project.yml variable instead of hardcoded to lightning?

template/build/core/phing/tasks/frontend.xml
```diff
@@ -1,53 +1,9 @@
 <project name="frontend" default="frontend:install">
 
-  <target name="frontend:build" depends="frontend:install" description="Uses gulp to build front end dependencies for all themes.">
-    <if>
-      <isset property="project.themes"/>
-      <then>
-        <foreach list="${project.themes}" param="frontend.theme" target="frontend:build:run"/>
-      </then>
-    </if>
+  <target name="frontend:build" description="Uses gulp to build front end dependencies for all themes.">
```
justinlevi: The description here feels a bit misleading to me. As a user I might expect there to already be some magic built in to find and compile my sass files with gulp. It's not clear by the description that I would need to define the command in my project.yml file. Unless I'm missing something which very well could be the case...

PR #203, opened 2016-06-24
Ignore custom theme node_modules folder

justinlevi:


grasmash:
Any reason not to simply ignore `node_modules` altogether?

justinlevi:
@grasmash good point. I'm wondering if it's a bit heavy handed to ignore node_modules and bower_components from the root .gitignore file? Would it be better to let the user enforce that in their own theme or is it beneficial for people who might forget, etc?

grasmash:
Given that "not committing external dependencies" is one of the core architectural principles of BLT, so I think we can be heavy handed in this case. People can always choose to customize the `.gitignore` file at the root.

PR #207, opened 2016-06-30
Fixed dev env detection on ACSF.

danepowell:
On ACSF all of the environments are prefixed with `01`.

### DIFF
--- a/template/docroot/sites/default/settings/base.settings.php
+++ b/template/docroot/sites/default/settings/base.settings.php
@@ -27,7 +27,7 @@ $is_ah_env = (bool) $ah_env;
 $is_ah_prod_env = ($ah_env == 'prod' || $ah_env == '01live');
 $is_ah_stage_env = ($ah_env == 'test' || $ah_env == '01test');
 $is_ah_dev_cloud = (!empty($_SERVER['HTTP_HOST']) && strstr($_SERVER['HTTP_HOST'], 'devcloud'));
-$is_ah_dev_env = (preg_match('/^dev[0-9]*$/', $ah_env) == TRUE);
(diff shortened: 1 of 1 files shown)


template/docroot/sites/default/settings/base.settings.php
```diff
@@ -27,7 +27,7 @@
 $is_ah_prod_env = ($ah_env == 'prod' || $ah_env == '01live');
 $is_ah_stage_env = ($ah_env == 'test' || $ah_env == '01test');
 $is_ah_dev_cloud = (!empty($_SERVER['HTTP_HOST']) && strstr($_SERVER['HTTP_HOST'], 'devcloud'));
-$is_ah_dev_env = (preg_match('/^dev[0-9]*$/', $ah_env) == TRUE);
```
grasmash: What about 02dev ?

danepowell: Heh, I knew you'd ask that... the answer is I have no clue. I've only seen `01`, and I know that ACSF can't have multiple dev environments, so I don't know how it would be anything else.

I'll try to find out.

template/docroot/sites/default/settings/base.settings.php
```diff
@@ -27,7 +27,7 @@
 $is_ah_prod_env = ($ah_env == 'prod' || $ah_env == '01live');
 $is_ah_stage_env = ($ah_env == 'test' || $ah_env == '01test');
```
greylabel: There are some ACE sites the use `stg` for stage instead of `test`. Do we want to account for these. Not sure how many in number overall.

template/docroot/sites/default/settings/base.settings.php
kylebrowning: It is used when Acquia adds capacity/tangles, highest I have ever seen is 02

PR #210, opened 2016-07-06
Don't overwrite local config files

gapple:
The copy task `overwrite` option defaults to `false`, and means that files will be overwritten if the timestamp of the source file is newer (https://www.phing.info/docs/guide/trunk/CopyTask.html).  This results in local changes being destroyed if any changes are made upstream.

This pull request adds a conditional based on the existence of the `project.local.yml` file, so that the files are only copied the first time setup is run.

### DIFF
--- a/template/build/core/phing/tasks/properties.xml
+++ b/template/build/core/phing/tasks/properties.xml
@@ -21,7 +21,14 @@
 
   <!-- Extract PHP variable values from local.settings.php. -->
   <property name="local.settings.file" value="${docroot}/sites/${multisite.name}/settings/local.settings.php" />
-  <copy file="${docroot}/sites/${multisite.name}/settings/default.local.settings.php" tofile="${local.settings.file}" overwrite="false" verbose="true"/>
+  <if>
+    <not>
+        <available file="${local.settings.file}" type="file" />
+    </not>
+    <then>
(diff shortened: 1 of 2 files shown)


grasmash:
Should we also be checking for the existence of local.settings.php and local.drushrc.php before copying those files as well?

gapple:
Each could be split into separate conditions, but I was making the assumption that a user would run `setup:drupal:settings` (or another task that depends on it), rather than individually copying the example files manually.

I think it's okay to assume that if a user does a manual copy of the project.yml file, they know to copy the other example files when needed as well.

gapple:
I found that properties.xml also contains a command to copy the default.local.settings.php file.  I'm not completely sure why, but in case something is expecting that behaviour I wrapped it in the same conditional.

I then tried making the `setup:drupal:settings` task only copy files if all of them were missing:  

    <if>
      <not>
        <or>
         <available />
         <available />
         <available />
        </or>
      </not>
      <then>
    </if>

but that will always fail if the local.settings.php file is created by properties.xml.

Which lead me to break out each file into a separate check + action, which is more verbose, but more understandable:

    <if>
      <not>
        <available />
      </not>
      <then />
    </if>
    <if>
      <not>
        <available />
      </not>
      <then />
    </if>

Issue #227, opened 2016-07-18
Force certain directories when deploying to Acquia Cloud


webkenny:
## Problem
Some developers will have a global Git ignore file that will exclude directories such as `/vendor` or other required directories to be pushed to ACE in `develop-build`. This causes dependency issues.

## Solution
Pull Request incoming but we want to, at minimum, force `/vendor` to be committed.

webkenny:
See https://github.com/acquia/blt/pull/228

CashWilliams:
I ran into this with something other then /vendor as well.

My global git ignore had something relating to icon files in it, and so some images named '*icon*' from http://cgit.drupalcode.org/drupal/tree/core/themes/seven/images?h=8.1.x were not getting added to the deploy commit.

grasmash:
Merged.

PR #228, opened 2016-07-18
#227 Force certain directories when deploying to Acquia Cloud.

webkenny:
Add vendor dir and force directive

### DIFF
--- a/template/build/core/phing/tasks/deploy.xml
+++ b/template/build/core/phing/tasks/deploy.xml
@@ -83,6 +83,8 @@
   <target name="deploy:commit">
     <!-- We make these commands quiet because they can cause the output to be so long that Travis CI stops logging. -->
     <exec command="git add -A" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
+    <!-- We are forcing vendor to add here in case a global git ignore flags it -->
+    <exec command="git add vendor -f" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
     <exec command="git commit -m '${deploy.commitMsg}' --quiet" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
   </target>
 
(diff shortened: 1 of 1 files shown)


template/build/core/phing/tasks/deploy.xml
```diff
@@ -83,6 +83,8 @@
   <target name="deploy:commit">
     <!-- We make these commands quiet because they can cause the output to be so long that Travis CI stops logging. -->
     <exec command="git add -A" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
+    <!-- We are forcing vendor to add here in case a global git ignore flags it -->
+    <exec command="git add vendor -f" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
```
bhasselbeck: might need to add 'vendor/*'

webkenny: Ah good call.

webkenny: Took it one step further and added wildcards before and after so we capture ALL vendor directories everywhere. I say this because in a `/gitignore` file `/vendor` will refer to any vendor directory anywhere.

template/build/core/phing/tasks/deploy.xml
```diff
@@ -83,6 +83,8 @@
   <target name="deploy:commit">
     <!-- We make these commands quiet because they can cause the output to be so long that Travis CI stops logging. -->
     <exec command="git add -A" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
+    <!-- We are forcing vendor everywhere to add here in case a global git ignore flags it -->
+    <exec command="git add *vendor/* -f" dir="${deploy.dir}" logoutput="true" passthru="true" checkreturn="true"/>
```
bhasselbeck: Well... The wildcard before could open some other issues despite what it fixes.  Not realistic issues... I could name a directory maliciousdirectoryvendor/* and it will be added.  Not realistic, but possible.  Doing "vendor/*" ensures it's docroot/vendor/* only.

grasmash: yeah I think adding only vendor should be fine. It will include subdirectories by default without the following asterisks, which could actually introduce the issue of ignoring hidden files.

grasmash: I.e., 'git add vendor -f' should be sufficient.

webkenny: Yes! Smart catch with the security issue. Fixing now.

Issue #237, opened 2016-07-20
Document how to modify deploy excludes


grasmash:

webkenny:
@sarahjean and I were talking about this and we think there's a bit more to this than just editing the `deploy-excludes.txt` file. Here's a proposal:

- Move the excludes file into something other than `core` so people don't feel intimidated when they change it.
- As a template, we might edit `vendor` to `/vendor` so BLT doesn't exclude recursive instances. Vendor is always needed in a BLT deploy.
- Document the changes.

I can assist with this since it bit me very recently.

sarahjean:
So it looks to me like this is the intended method:

- copy build/core/files/deploy-excludes.txt to build/custom/files/deploy-excludes.txt
- make modifications there
- in build/custom/phing/builld.yml, add this:

```
deploy:
  dir: ${repo.root}/deploy
  exclude_file: ${repo.root}/build/custom/files/deploy-exclude.txt
```

Haven't fully tested yet and could be off base with some of my assumptions.

grasmash:
Yes that is intention.

Vendor should not be copied since it contains dev dependencies and should be rebuilt dynamically prior to deployment.

webkenny:
Ok, so it sounds like what Sarah is highlighting above simply needs to be better documented. I spent way too long trying to resolve why BLT wasn't building my whole build.

I'll issue a PR for this ASAP.

sarahjean:
@grasmash For further context, the type of issue that @webkenny and I were both running into (on separate projects) was that in some of the libraries we were pulling in there were directories named vendor that _did_ need pulled in, where the intent of the deploy-exclude.txt vendor is to exclude specifically the composer vendor directory. 

There's also a line for `bower_components` in the BLT deploy-excludes file and since bower is for pulling in front end dependencies I could see quite a few cases where not deploying that could cause unexpected results. (Fonts, CSS files, JS plugins).

grasmash:
In that case, ignore /vendor but not vendor.

webkenny:
Well here's a start to that then... #227 originally tagged "vendor". I've submitted a PR for `/vendor` to at least make that happen. https://github.com/acquia/blt/pull/247

grasmash:
@webkenny There are two commands that run during the creation of the deployment artifact that we must consider. 

* An rsync command that copies files from the repository root into the `deploy` dir. This will not copy files that have been listed in `build/core/files/deploy-excludes.txt`. 
[...]

PR #247, opened 2016-08-01
#237 Change the vendor force to /vendor so it captures all vendor directories; not just the root.

webkenny:


grasmash:
Closing. Please reroll to match implementation noted here https://github.com/acquia/blt/issues/237#issuecomment-236583497

Issue #249, opened 2016-08-01
Get project prefix from project.yml for git-commit hook


msherron:
I added a bash yaml parser to the /scripts/blt dir so the commit-msg hook can use the project.yml prefix instead of having your JIRA prefix hard coded in the commit-hook itself.

msherron:
PR: https://github.com/acquia/blt/pull/250

msherron:
Today, when you run `./blt.sh create`, the project.prefix is pulled from project.yml and hard coded into the commit-msg hook. The issue is that, at least for my project, we didn't have JIRA at the outset, so the prefix was unknown. It would be a nice feature to just make the change in project.yml and then commit-msg just pulls from there.

PR #250, opened 2016-08-01
#249: Adding yaml parser to git-commit hook

msherron:


grasmash:
Looks like the contents of parse-yaml.sh might be wrong?

msherron:
woops, updated :)

grasmash:
This seems like a great idea, but I'm hesitant to use bash string parsing to find this value. Adding a yaml:get command to drupal/console that leverages Symfony's yaml parser seems like a more reliable path forward.

See https://github.com/hechoendrupal/DrupalConsole/tree/master/src/Command/Yaml

grasmash:
Pinging @jmolivas in case he has an opinion on adding something like `yaml:get:value [key] [filename]` to Drupal Console.

grasmash:
See https://github.com/hechoendrupal/DrupalConsole/pull/2553

grasmash:
`yaml:get:value [filename] [key]` has been added to drupal console and released in beta5.

grasmash:
Closing in favor of https://github.com/acquia/blt/pull/256

PR #253, opened 2016-08-02
Updating template composer.json file to update drupal consoles versio…

haynescw:
…n and PhantomJS Version. The version of PhantomJS that you had was notorious for crashing and Drupal Console updated to include more functionality.


grasmash:
Console was updated in a different PR. Please resubmit if you'd like PhantomJS updated.

haynescw:
OK, i just wanted to make sure that it was the console on the template
composer,json file and not BLT's composer composer.json file. I will
resubmit with only the change to PhantonJS.

Issue #221, opened 2016-07-13
Allow separate setup and build frontend tasks


gapple:
More expensive operations like `npm install && bower install` only need to be run on initial setup and when dependencies change.

Separating them into a `frontend:setup` task would allow `frontend:build` to run fewer operations and take less time during development.

sarahjean:
++

sarahjean:
Would also be good to document caching the npm and bower assets in .travis.yml

grasmash:
@gapple @sarahjean If I run `npm install` or `bower install` against a directory that already has built dependencies, won't it just skip installing them?

sarahjean:
@grasmash That'd be nice but it doesn't seem to behave that way. I think composer is a little better behaved in that regard.

PR #259, opened 2016-08-04
issue-221: Adds a frontend:setup task and also passthru for target-hooks

sarahjean:
Relates to #221

### DIFF
--- a/template/build/core/phing/tasks/frontend.xml
+++ b/template/build/core/phing/tasks/frontend.xml
@@ -1,4 +1,10 @@
-<project name="frontend" default="frontend:install">
+<project name="frontend" default="frontend:setup">
+
+  <target name="frontend:setup" description="Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build.">
+    <phingcall target="target-hook:invoke">
+      <property name="hook-name" value="frontend-setup"/>
+    </phingcall>
+  </target>
 
   <target name="frontend:build" description="Uses project.yml hooks to run custom defined commands to build front end dependencies for custom themes.">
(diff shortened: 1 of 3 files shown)


template/build/core/phing/tasks/frontend.xml
```diff
@@ -1,4 +1,10 @@
-<project name="frontend" default="frontend:install">
+<project name="frontend" default="frontend:setup">
+
+  <target name="frontend:setup" description="Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build.">
```
grasmash: Need to update description.

sarahjean: @grasmash It is different, is "Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build." confusing?

grasmash: oh, I guess I see what you did there. The frontend:build reference confused me initially.

grasmash:
Nothing appears to actually trigger frontend:setup. Should it be a dependency of frontend:build?

sarahjean:
@grasmash The idea is that frontend:build can be run without triggering frontend:setup and would therefore run faster since it doesn't have to do `npm install` or whatever. It would make more sense to trigger it for local:setup and/or for a Travis build but not have it run every time you do a frontend:build.

grasmash:
In that case, I'd introduce a `frontend` target that depends on `frontend:install` and `frontend:build`. setup can then depend on `frontend`.

grasmash:
Closing in favor of https://github.com/acquia/blt/pull/291

PR #263, opened 2016-08-05
Replace blt sh mentions with blt alias.

ChuChuNaKu:
I replaced all mentions of ./blt.sh with blt in the composer-package branch.

### DIFF
--- a/phing/README.md
+++ b/phing/README.md
@@ -8,12 +8,12 @@ This directory should not contain any test files. Those exist in the [/tests](/t
 
 A large number of common build tasks are provided via Phing targets. These include tasks for things like code sniffing, executing tests, building dependencies, installing Drupal, etc.
 
-For a full list of available Phing tasks, run `./blt.sh -list` from the project's root directory.
+For a full list of available Phing tasks, run `blt -list` from the project's root directory.
 
 ### Executing Tasks
 
-* For a full list of the available Phing targets, run `./blt.sh -list`
-* To manually test a Phing target, run the following command matching the the following pattern: `./blt.sh <target-name>`. For example `./blt.sh validate:all`
+* For a full list of the available Phing targets, run `blt -list`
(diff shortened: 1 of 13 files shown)


composer.json
```diff
@@ -22,6 +22,6 @@
   "minimum-stability": "beta",
   "prefer-stable": true,
   "scripts": {
-    "install-blt-alias": "./blt.sh install-alias"
+    "install-blt-alias": "blt install-alias"
```
grasmash: This is actually one place we need to refer to the script explicitly, since it would only be run if the alias weren't already installed.

scripts/tugboat/init.sh
```diff
@@ -11,4 +11,4 @@ composer install
 $DIR/install-node.sh 4.4.1
 
 
-$DIR/../../blt.sh ci:setup
+$DIR/../.blt ci:setup
```
grasmash: This must remain a file reference. We cannot use an alias in combination with file paths.

scripts/tugboat/build.sh
```diff
@@ -3,4 +3,4 @@
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 # Update current database to reflect the state of the Drupal file system
-$DIR/../../blt.sh ci:update
+$DIR/../.blt ci:update
```
grasmash: This will break the script.

scripts/tugboat/update.sh
```diff
@@ -5,4 +5,4 @@ cd /var/lib/tugboat/docroot
 chown -R www-data sites/default/files
 
 # Update current database to reflect the state of the Drupal file system
-$DIR/../../blt.sh ci:update
+$DIR/../.blt ci:update
```
grasmash: This will break the script.

Issue #262, opened 2016-08-05
Make -r in drush.wrapper read from drush.root in project.yml


grasmash:

grasmash:
https://github.com/acquia/blt/pull/265

grasmash:
Resolved in #276

PR #265, opened 2016-08-08
Connected #262: Attempt to set drush.root from project.local.yml if it exists.

pixlkat:
@grasmash - I've changed this so it attempts to set the drush root from project.local.yml if the file exists. The check for the return value is the error message i received when the specified key was not found in the YAML file. The default value is just an empty string.

### DIFF
--- a/template/drush.wrapper
+++ b/template/drush.wrapper
@@ -29,6 +29,15 @@
 # composer-managed Drupal sites.
 #
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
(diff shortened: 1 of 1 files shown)


grasmash:
testDrushConfig FAILED

grasmash:
drush.root needs to be set even if project.local.yml does not exist. It has a default value set in build.yml.

We need to either iteratively check each property file, or else using a Phing target to grab this value.

grasmash:
@pixlkat I've implemented this change. Can you update your blt and test drupal vm to confirm it's fixed?

PR #283, opened 2016-08-12
[BUGFIX] Tests may fail due to PhantomJs not ready

mickaelperrin:
- PB: Under some systems, Behat tests fails due to the fact the PhantomJS server is not ready to handle connections.
- FIX: ensure server is running before continuing.
- NOTE: currently use curl to test, if curl is not available, it doesn't wait

### DIFF
--- a/phing/tasks/tests.xml
+++ b/phing/tasks/tests.xml
@@ -63,6 +63,14 @@
               passthru="true"
               spawn="true"
               checkreturn="true" />
+        <exec command="which curl &gt; /dev/null &amp;&amp; echo $?" checkreturn="false" outputProperty="hasCurl" />
+        <if>
+          <equals arg1="${hasCurl}" arg2="0"/>
+          <then>
+            <echo message="Waiting for PhantomJS to start..." />
+            <exec command="while ! curl -v -L http://127.0.0.1:4444/wd/hub/session; do sleep 1; done" checkreturn="false" />
+          </then>
+        </if>
       </then>
     </if>
   </target>
--- a/phing/tasks/tests.xml
+++ b/phing/tasks/tests.xml
@@ -63,14 +63,10 @@
               passthru="true"
               spawn="true"
               checkreturn="true" />
-        <exec command="which curl &gt; /dev/null &amp;&amp; echo $?" checkreturn="false" outputProperty="hasCurl" />
-        <if>
-          <equals arg1="${hasCurl}" arg2="0"/>
-          <then>
-            <echo message="Waiting for PhantomJS to start..." />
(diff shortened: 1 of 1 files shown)


phing/tasks/tests.xml
```diff
@@ -63,6 +63,14 @@
...
               spawn="true"
               checkreturn="true" />
+        <exec command="which curl &gt; /dev/null &amp;&amp; echo $?" checkreturn="false" outputProperty="hasCurl" />
+        <if>
+          <equals arg1="${hasCurl}" arg2="0"/>
+          <then>
+            <echo message="Waiting for PhantomJS to start..." />
+            <exec command="while ! curl -v -L http://127.0.0.1:4444/wd/hub/session; do sleep 1; done" checkreturn="false" />
+          </then>
+        </if>
```
grasmash: We already use a similar bit of logic to wait for PHP's runserver to repond:

```
        <echo message="Waiting 10 seconds for ${behat.server-url} to become available."/>
        <waitfor maxwait="10" maxwaitunit="second" checkevery="1" checkeveryunit="second">
          <http url="${behat.server-url}"/>
        </waitfor>
```

Perhaps we should use a similar approach? This does have the benefit of preventing and infinite loop in the case that PhantomJS never responds.

mickaelperrin: Interesting, you are right, a timeout should be introduced to prevent an infinite loop.

Regarding the way to do the http request, do you prefer to stick with `http` from Phing which relies on the `PEAR` package `HTTP_Request2`, or should we use the `curl` approach ? or both curl as a failover for example ?

grasmash: Since we already are using the `http` task for Phing, I think we should stick with it. If we find that the dependency on the `HTTP_Request2` package is problematic in future, we can switch both instance on `http` to use `curl` in the way that you've shown.

mickaelperrin: OK I will update the PR.

phing/tasks/tests.xml
```diff
@@ -63,6 +63,10 @@
               passthru="true"
               spawn="true"
               checkreturn="true" />
+        <echo message="Waiting 10 seconds for ${behat.server-url} to become available."/>
```
grasmash: I think that the behat.server-url isn't actually what you'd like to test here. The port is wrong.

mickaelperrin:
Mmm, sorry for that... I hardcoded the url as the port is hardcoded also in the launch of PhantomJS, is it ok for you ?

grasmash:
Yeah, good for now, thanks!

PR #286, opened 2016-08-12
Add documentation for running a single behat test with phing.

marksakurada:


readme/testing.md
```diff
@@ -75,13 +75,17 @@ Each testing type can be either executed directly, or via a corresponding Phing
 * `blt tests:behat`
 * `blt tests:phpunit`
 
+To execute a single test (with Phing):
+
+* `blt tests:behat:execute -Dbehat.param=tests/behat/features/Examples.feature`
```
grasmash: Are you sure this works? I don't see how behat.param is passed to the call to Behat in Phing.

marksakurada: It works on my build to run a single test or tests/behat/features

I was looking for a way to run just a single test via blt when writing tests without running everything that a 'blt tests:behat" runs. the -D sets the option for the function and works just like the param.

Issue #295, opened 2016-08-15
Move tests/phpunit/Bolt tests into /tests


grasmash:
Move tests from `template/tests/phpunit/Bolt` and `template/tests/phpunit/src/TestBase.php` to `tests/phpunit` and ensure that they continue to be executed by `template/.travis.yml` and pass.

pixlkat:
@grasmash - I'll pick this up, but I do not appear to have access to assign myself.

pixlkat:
@grasmash - After setting up a blt-project directory next to blt as outlined in CONTRIBUTING.md, I am able to get all the phpunit tests to pass with the exception of the drush test. However, I'm not able to actually install drupal using any of the blt commands.

My setup:
Follow steps in CONTRIBUTING.md:

```
cp -R blt/blt-project .
cd blt-project
git init
composer install
blt init
blt configure
composer update
```

Follow steps in INSTALL.md which follow `blt init/blt configure/composer update`. I configured DrupalVM as I normally would, but I left all the defaults for configuration of project name, etc.

```
blt vm:init
vagrant up
blt local:setup
```

The drush site-install command results in the following error:

```
blt > setup:drupal:install:

     [echo] Installing Drupal.
    [drush] Executing '/Users/steph.bridges/workspace/blt-project/vendor/bin/drush @blted8.local --site-name="BLTed 8" --site-mail="no-reply@acquia.com" --account-name="admin" --account-pass="admin" --account-mail="no-reply@acquia.com" --l="default" --nocolor --root=/Users/steph.bridges/workspace/blt-project/docroot --yes --verbose site-install "lightning" "install_configure_form.update_status_module='array(FALSE,FALSE)'"'...
Loaded alias @blted8.local from file                                                                              [notice]
/Users/steph.bridges/workspace/blt-project/docroot/../drush/site-aliases/drupal-vm.aliases.drushrc.php
Begin redispatch via drush_invoke_process().                                                                      [notice]
[...]

PR #299, opened 2016-08-16
Moved tests in /template/tests/phpunit to /tests/phpunit.

pixlkat:
See #295


grasmash:
You'll need to modify the psr-4 to the autoload array in the top level composer.json.

grasmash:
Move this https://github.com/acquia/blt/blob/8.x/template/composer.json#L45

PR #313, opened 2016-08-17
Move phpunit tests (#295)

pixlkat:


tests/phpunit/src/TestBase.php
```diff
@@ -21,7 +21,8 @@
   public function __construct($name = NULL, array $data = array(), $data_name = '') {
 
     parent::__construct($name, $data, $data_name);
-    $this->projectDirectory = dirname(dirname(dirname(__DIR__)));
+    $directory = realpath(dirname(__FILE__) . '/../../../');
```
grasmash: This is likely failing because of how the tests are called. They are intended to be called from the child `blt-project`, but they are being called from the parent `blt` here:
https://github.com/acquia/blt/blob/8.x/.travis.yml#L90

We'll want to tag these tests with a group like `@blt-project`. Then, we can run test in two sets:

```
- cd ../blt-project
- phpunit tests/phpunit --group blt-project
- cd ../blt 
- phpunit tests/phpunit --exclude-group blt-project
``` 

The `blt-project` tests can then rely on a $this->project directory variable that refers to `../../../../../../../` given that the actual `TestBase` will live in `vendor/acquia/blt/tests/phpunit/src`.

tests/phpunit/src/DrushTest.php
```diff
@@ -11,13 +11,17 @@ class DrushTest extends TestBase {
...
    */
   public function testDrushConfig() {
 
     // We must define the absolute path of the binary because child shell
     // processes in PHP to not inherit $PATH setting from environment.
     $drush_bin = $this->projectDirectory . '/vendor/bin/drush';
-    $command = "$drush_bin status";
+    // Use --format=json output so we don't have to deal with output line
+    // wrapping when running the tests.
+    $command = "$drush_bin status --format=json";
```
grasmash: Great idea to return this as json.

tests/phpunit/src/TestBase.php
pixlkat: I'm confused here - the tests themselves all now live in vendor/acquia/blt/phpunit/src. Running `phpunit tests/phpunit` from the blt-project directory does not find the tests that live in blt.

.travis.yml
```diff
@@ -84,10 +84,12 @@ script:
   - blt deploy:build
   # Add Drupal VM config to repo.
   - blt vm:init
+  # Run the 'blt-project' phpunit tests.
+  - phpunit vendor/acquia/blt/tests/phpunit --group blt-project
```
grasmash: Don't these get executed already by the ci:build:validate:test target run in blt-project?

pixlkat: You're right, they do appear to be run, I missed them when inspecting the test output earlier. I'll take that back out.

PR #317, opened 2016-08-18
Allow overrides of blt phing targets: fixes #316.

### LINKED ISSUE #316: override blt phing target
I am hoping to customize (override) some of the BLT tasks, but I am having an issue.
I am upgrading from the pre-composer acquia/blt, where my project's `/blt.sh` file was edited/customized to load `build/custom/phing/build.xml`, which then in turn loads `build/core/phing/build.xml`.

In my current project, because the custom buildfile is loaded directly to Phing, I can do this:
````
<target name="deploy:build" description="Override blt's deploy:build command."
   depends="frontend:build, deploy:copy, deploy:composer:install, deploy:profile:make, deploy:sanitize, custom:target">
</target>
````
Then, when I call `$ blt deploy:build` (or even if it's called from _another_ core target), my overridden target is executed (notice the custom target included at the end of _deploy:build_ depends list).

According to the [Target Overridding section](https://www.phing.info/docs/guide/trunk/ImportTask.html) in Phing documentation, the main `build.xml` takes precedence. 
> If a target in the main file is also present in at least one of the imported files, the one from the main file takes precedence.

This explains why I am able to accomplish this in "old blt", but not in "composer blt". 

With the new composer release, this is no longer possible to do as `blt.sh` has been removed from the repo. So I began investigating how to do something like this again.

Eventually, I came around to testing the below build.xml file with phing.

**custom.xml**
````
<project name="custom">
    <target name="mytarget"><echo>mytarget first called!</echo></target>
    <target name="mytarget"><echo>mytarget second called!</echo></target>
    <target name="mytarget"><echo>mytarget third called!</echo></target>
</project>
````
Command Line:
````
$ vendor/bin/phing -f custom.xml mytarget
custom > mytarget:
    [echo] mytarget first called!
````
It seems like the only definition that matters is the earliest appearance of a target. Therefore, in order to make **blt** more extend-able, and allow me to still override the **blt** targets, I'm proposing we move up the custom `import` file in the acquia/blt `phing/build.xml` to just after the properties load.

I will include a PR for this approach and tag this issue. Please comment if there is a better way to do this that I am missing.

dpagini:
See acquia/blt#316 for the long-winded explanation. Essentially, the goal of this PR is to make **BLT** more extensible by allowing custom overrides of BLT phing targets.

By moving the import file declaration up in the `build.xml` file, we are allowing a custom definition of each BLT target to be defined before the BLT library's version.

### DIFF
--- a/phing/build.xml
+++ b/phing/build.xml
@@ -19,6 +19,18 @@
   <!-- Defines a list of default properties, available to all targets. -->
   <import file="${phing.dir}/tasks/properties.xml"/>
 
+  <!-- Allow custom file to be imported. -->
+  <if>
+    <isset property="import" />
+    <then>
+      <echo>Importing custom Phing file ${import}.</echo>
+      <import file="${import}"/>
+    </then>
+    <else>
+      <echo>No custom Phing file specified.</echo>
+    </else>
+  </if>
+
   <!-- Contains BLT tasks. -->
   <import file="${phing.dir}/tasks/blt.xml"/>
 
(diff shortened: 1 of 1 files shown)


grasmash:
I had assumed that defining the custom targets _after_ the blt targets allowed them to be overridden. I assume it's not working that way?

dpagini:
@grasmash I agree, that would make more sense to me, and I was surprised to learn it didn't work that way.

Issue #211, opened 2016-07-07
BLT doesn't define file system storage intelligently


bhasselbeck:
Steps to reproduce: 
1. Export a BLT project using this repository
2. Setup the project
3. Deploy code to Acquia Cloud
./blt.sh deploy -Ddeploy.branch=<branch> -Ddeploy.commitMsg='<commit>'
4. SSH into Acquia Cloud or use Aliases to install Lighting, or a standard profile
5. Note that the filesystem seems to not be configured properly

Also, note that settings.php doesn't define file storage paths:

- https://github.com/acquia/blt/blob/8.x/template/docroot/sites/default/settings.php#L517
- https://github.com/acquia/blt/blob/8.x/template/docroot/sites/default/settings.php#L502

We need to set standards for the filesystem based off of
1. Acquia Cloud
2. Local Development
3. ACSF

pixlkat:
Questions:

1. Should we be enabling private file system by default?
2. I am unsure of what the standards for ACSF are for filesystem paths.  

For now, I have created a PR which adds a filesystem-specific settings file with some generic AC settings.

PR #321, opened 2016-08-22
Add configuration for filesystem settings. (#211)

pixlkat:


pixlkat:
This is failing because lightning requires metatag at ^8.1.0 and applies a patch for [this issue](https://www.drupal.org/node/2765137). Metatag released 8.1.0-beta10 today which includes that patch which apparently breaks the build.

PR #323, opened 2016-08-23
Fix drush.wrapper script to use only last line of output of phing ech…

pixlkat:
…o-property task when discovering the drush root. (#315)

@grasmash - I altered the command a bit so the value returned is saved into an array, and assign drush.root to be the last value in the array. I tested this locally with xdebug enabled and disabled, and it appears to work correctly in both cases.


grasmash:
Can't we just pipe the property through `tail`?

grasmash:
https://github.com/acquia/blt/pull/337

Issue #324, opened 2016-08-24
Add nodejs frontend tools for drupalvm integration


bhasselbeck:
When building out on the virtual machine, some necessary plugins are not included for phantomJS or other front-end engine building tools such as node, npm, bower, and gulp.  We should include these by default. 

https://github.com/acquia/blt/blob/8.x/scripts/drupal-vm/config.yml

We could simply add the following configuration: 

```
# This is required for front-end building tools.
nodejs_version: "0.12"
nodejs_npm_global_packages:
  - name: bower
  - name: gulp-cli
nodejs_install_npm_user: "{{ drupalvm_user }}"
npm_config_prefix: "/home/{{ drupalvm_user }}/.npm-global"
installed_extras:
  - nodejs

extra_packages:
  # This is required for PhantomJS Installer.
  - php-bz2
  - php-bcmath
```

grasmash:
Merged in #325

PR #325, opened 2016-08-24
Githiub BLT Issue #324: Includes NodeJS tools as well as php-bcmath…

bhasselbeck:
… ext.


bhasselbeck:
Looks like the build failed for a completely unrelated reason.

Issue #284, opened 2016-08-12
Install BLT alias automatically for developers


danepowell:
When a developer clones a project repo built on the new composer-based BLT, they have to do something like this:

1. `composer install`
2. `./vendor/bin/blt local:setup`

I think we could consolidate this to automatically run the local setup as part of the composer install. We should at least install the BLT alias so they don't have to poke around the vendor directory for the BLT binary.

grasmash:
In order to do this, we'd need to either:
* Have the project's composer file contain a `post-update-cmd` script, or
* Define blt as a composer plugin in the way that cweagens/composer-patches does

pixlkat:
I've been investigating how to accomplish this today. In its current form, the install-alias.sh script requires appropriate input from a prompt before it will actually install the alias. Scripts run as composer scripts do not support user input as a response to a prompt from a shell script. See https://github.com/composer/composer/pull/4036.

It is relatively straightforward to add a script definition to `template/composer.json` so that the script can be run, however if the alias does not already exist it is not created. If we want to go this route, we will need to modify the script so that it will check for a "-y" flag so it doesn't need to ask when running the command from composer.

I think running `blt local:setup` automatically is not really feasible since between setting up the project directories and installing everything, the local dev environment needs to be set up.  I think the best we can do here is to run `blt setup:drupal:settings`.

grasmash:
Installing automatically causes issues in some environments, but I've made it much simpler: `composer blt-alias`.

I don't think we should run `local:setup` because some people may be adding blt to an existing codebase with an existing db, which would be dropped.

kevinquillen:
One slight thing, the same function that works in fish, causes zsh to exit to iTerm/Term or the window itself will close (if zsh is set as default shell) if blt is ran in a directory fails the conditions in the function.

PR #328, opened 2016-08-25
Add script to install BLT alias to template composer.json script, upd…

pixlkat:
…ate setup instructions in onboarding.md. (#284).

The simplest solution seemed to be to add the script to the template composer.json. I modified the script itself to accept a "-y" flag to automatically install the alias since composer does not support interactive scripts. This does not add the script to existing projects, but I think that is not necessary since it is really only a one-time thing.

### DIFF
--- a/scripts/blt/install-alias.sh
+++ b/scripts/blt/install-alias.sh
@@ -23,11 +23,22 @@ if [ ! -z "$DETECTED_PROFILE" ]; then
     exit
   fi
 
+  while getopts ":y" arg; do
+  case $arg in
+    y)
+      REPLY=y
+      ;;
+    esac
+  done
+
   echo "BLT can automatically create a Bash alias to make it easier to run BLT tasks."
   echo "This alias may be created in .bash_profile or .bashrc depending on your system architecture."
   echo ""
-  read -p "Install alias? (y/n)" -n 1 -r
-  echo ""
+
+  if [ -z $REPLY ]; then
+    read -p "Install alias? (y/n)" -n 1 -r
+    echo ""
+  fi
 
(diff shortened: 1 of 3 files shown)


scripts/blt/install-alias.sh
```diff
@@ -23,11 +23,22 @@ if [ ! -z "$DETECTED_PROFILE" ]; then
...
   echo "BLT can automatically create a Bash alias to make it easier to run BLT tasks."
   echo "This alias may be created in .bash_profile or .bashrc depending on your system architecture."
   echo ""
-  read -p "Install alias? (y/n)" -n 1 -r
-  echo ""
+
+  if [ -z $REPLY ]; then
+    read -p "Install alias? (y/n)" -n 1 -r
+    echo ""
+  fi
```
grasmash: This script already accepts `-y` as an argument. Is it not working?

template/composer.json
```diff
@@ -57,11 +57,14 @@
   },
   "scripts": {
     "install-phantomjs": "PhantomInstaller\\Installer::installPhantomJS",
+    "install-blt-alias": "./vendor/acquia/blt/scripts/blt/install-alias.sh -y",
```
grasmash: call this with `yes | ./vendor/acquia/blt/scripts/blt/install-alias.sh` to accept `y` input.

scripts/blt/install-alias.sh
pixlkat: It didn't when I tried it, maybe that's because i didn't know what I was doing.

pixlkat:
@grasmash I'm going to close this PR and open a new one - I think my problem was not knowing how to pass input to the script. I'll adjust and recreate a new PR.

PR #359, opened 2016-09-02
Fixes #211: Store filepaths intelligently.

### LINKED ISSUE #211: BLT doesn't define file system storage intelligently

grasmash:


grasmash:
Does this work with multisite?

PR #363, opened 2016-09-07
Modify deploy phing to enable multiple environment deployments.

marksakurada:
While attempting to deploy to multiple environments with travis I was blocked by:

[exec] Executing command: git checkout -b master-build
fatal: A branch named 'master-build' already exists.
[phingcall] /vendor/acquia/blt/phing/tasks/deploy.xml:62:59: Task exited with code 128

changing this to a -B checkout argument should create or switch the the correct branch.

### DIFF
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -59,7 +59,7 @@
     <exec command="git fetch ${remoteName} ${deploy.branch}" dir="${deploy.dir}" logoutput="true" level="info" passthru="true"/>
 
     <!-- Create the new branch, "[source-branch-name]-build". -->
-    <exec command="git checkout -b ${deploy.branch}" dir="${deploy.dir}" logoutput="true" checkreturn="true" level="info" passthru="true"/>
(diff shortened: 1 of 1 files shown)


grasmash:
I think I encountered this before and found that simply allowing the command to fail silently with `checkreturn="false"` was a good workaround.

marksakurada:
I don't think we should ignore output. We allow for multiple deployment remotes in the project.yml. By changing to the -B we should never get an error and can facilitate multiple deployments.

"If -B is given, <new_branch> is created if it doesn’t exist; otherwise, it is reset. "

grasmash:
Can you please rebase this? For some reason Travis isn't pulling in a fresh 8.x and the upped memory limit isn't affecting your PR as it should.

marksakurada:
Rebased

grasmash:
I'm merging this... but I'm going to express a little concern. I have a hazy memory of trying this at some point in the past and having it go badly. We will need to do some deployment testing to ensure that this doesn't break anything before rolling it into a release.

grasmash:
This pull request has caused deployments to begin failing. Reverting.

https://travis-ci.org/acquia/blt/jobs/158770723

PR #392, opened 2016-09-14
fix template/.gitattributes

dpagini:
I believe this is just a spelling mistake?


dpagini:
Sorry for the 3 commits... they came over from trying to keep my fork up-to-date. I may be doing that wrong.

PR #414, opened 2016-09-16
Added Videos

audiojak:
Added youtube videos to the Readme. I think they are a great resource.


grasmash:
I just need to update those videos!

grasmash:
Sorry, closing this PR until we have more updated videos. I think adding those would be confusing given that they reference deprecated commands and processes.

PR #418, opened 2016-09-19
Bugfix for ACSF db name.

lcatlett:
This fixes a very major ACSF issue where site-specific ACSF settings.inc template provided by the ACSF module files are overridden in BLT - this caused ACSF sites in each environment to share a single, incorrect database and prevented the creation of new sites in some instances.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -83,7 +83,7 @@ $settings['hash_salt'] = file_get_contents(DRUPAL_ROOT . '/../salt.txt');
 /**
  * Acquia Cloud settings.
  */
-if ($is_ah_env && file_exists('/var/www/site-php')) {
+if ($is_ah_env && file_exists('/var/www/site-php' && !$is_acsf)) {
   require "/var/www/site-php/{$_ENV['AH_SITE_GROUP']}/{$_ENV['AH_SITE_GROUP']}-settings.inc";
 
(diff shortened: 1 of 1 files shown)


settings/blt.settings.php
```diff
@@ -83,7 +83,7 @@
 /**
  * Acquia Cloud settings.
  */
-if ($is_ah_env && file_exists('/var/www/site-php')) {
+if ($is_ah_env && file_exists('/var/www/site-php' && !$is_acsf)) {
```
dpagini: @grasmash & @lcatlett 
Should this be:
````
if (($is_ah_env && !$is_acsf) && file_exists('/var/www/site-php')) {
````
This line seems like it wouldn't work `file_exists('/var/www/site-php' && !$is_acsf)`?

cc: @rjgwiz

lcatlett: This does work when deployed to ACSF, but your logic makes more sense

dpagini: I can submit a PR, just wanted to make sure it wasn't intentional. I assume this works on ACSF b/c the `file_exists()` command returns false (maybe?), but this would then not run correctly on ACE, I believe.

grasmash: Uh yeah I missed that. @dpagini I like the look of your suggestion better.

Issue #384, opened 2016-09-14
Is acquia.inc an appropriate file to include in this repo?


typhonius:
Many Acquia customers use [this](https://gist.github.com/typhonius/c93f10469d153f396eb0f16c7181eab9) or the D7 version to protect their site/paths. Is BLT a suitable place to have this also included.

As a sidenote to this, it's been raised to me that this could potentially be a [D8 Middleware service](https://www.drupal.org/developing/api/8/middleware). Thoughts from smarter people are welcome in this question too.

dpagini:
It looks like there's some code in `blt/template/factory-hooks/post-settings-php/protect_env.php.example`which is supposed to be a ported version of the D7 acquia.inc file, but this file references a non-existent file in the project.
I would be interested to see the D8 acquia.inc approach make its way into acquia/blt, and would love to contribute some changes I've made to the script as well.

steveworley:
My thoughts are that it should be a middleware service, separate configuration from functionality. Core has the [Ban](https://api.drupal.org/api/drupal/core!modules!ban!ban.module/8.2.x) module which allows limited IP blacklisting. It stores IPs in a table and performs queries in the middleware, so I would think checking configuration would pose no problems.

If done as a service it should be a module and could be added via composer during install?

dpagini:
I guess I would suggest that if it's an Acquia [recommendation](https://docs.acquia.com/article/website-access-restriction-methods) for protecting your environment, it makes sense to add it into the acquia (sort of?) framework. 

The approach has the file included into your settings.php and configured there, so it _could_ make sense to add it to `settings/acquia.inc` (I would suggest a different name, but that's a moot point at the moment) in BLT, just to make it available to users who want to use that method. If you dont use it, it's just a 200 line text file that sits in the vendor folder un-referenced.

I don't personally mind too much if it doesn't make it in there, as we can keep it in our repo. I think adding it to the repo gives a great way to open source contribute to the file though.

If I have a chance, I could stage a PR for review & critique at some point, if it's even a road to go down.

grasmash:
Yeah I agree that it makes more sense as middleware than as part of BLT. Though, I'd be happy to add it as a default dependency for new BLT projects.

PR #422, opened 2016-09-19
Delete protect_env.php.example

dpagini:
Does this make sense here? It's an example file that loads `/sites/default/settings/envprotect.settings.php`, which is a file that does not exist in BLT. Per discussion in #384, the functionality that this file is referencing was decided to make more sense as middleware (as it relates to BLT).


grasmash:
Good catch.

PR #432, opened 2016-09-20
Change/remove drupal scaffold

dpagini:
Ok, this is a total shot in the dark... I'm guessing this has maybe already been thought of, but was working on this today and wanted to see if there are any thoughts on this approach.

Ultimately, if we don't keep drupal/core or any contrib modules in our repo, I don't see any reason to keep drupal-scaffold files either (unless you have changes to these, which I do for `docroot/.htaccess` for example). This also seems to me to be an approach that @grasmash slightly agrees with given the changes in PR #391. I will note, this goes against the suggested approach of the [drupal-composer/drupal-scaffold](https://github.com/drupal-composer/drupal-scaffold) project:

> It is assumed that the scaffold files will be committed to the repository, to ensure that the correct files are used on the CI server (see Limitation, above). After running composer install for the first time commit the scaffold files to your repository.

So here is what I'm doing... 

First, with this approach, users will not have the drupal scaffold files in their local installs, as they only get placed by the drupal-scaffold project on updates of drupal/core module. We can get around this by calling the library's functionality manually. Composer has a new command added to it, `$ composer drupal-scaffold`, which downloads and places the scaffold files on every run. This command gets added as a 2nd step to BLT's `setup:composer:install` command, which users will run, and CI would run. This, IMO, takes care of the drupal-scaffold warning around CI, quoted above.
(Side note: it feels like we are running this more than we have to though)

Next, I wanted to, by default, `.gitignore` the scaffold files. Since I also want users to be able to place a scaffold file in their repo (to modify/extend), and not have BLT overwrite it, I also place this .gitignore file in the `scripts/blt/ignore-existing.txt` file.

Now, if as a BLT end user, I want to include `docroot/.htaccess` file in my repo to add changes, I would remove that file's corresponding line from `docroot/.gitignore`, and commit that file to my repo. 
I could probably use some documentation on this approach as well, because by _now having_ this file in my repo, everytime `$ composer update` gets run with an update to drupal/core, it will try to revert my custom changes to `docroot/.htaccess`. However, I can [add a section to `composer.json` "extra/drupal-scaffold/excludes"](https://github.com/drupal-composer/drupal-scaffold#configuration) to tell drupal-scaffold to not overwrite my changes. In this case, you will not necessarily know about upstream changes to these files, but that seems like a choice we can leave to the end user: a) manually ignore upstream changes to drupal-scaffold files, b) have drupal-scaffold leave your file alone, but not easily find out about upstream changes.

Possible ToDo's Here:
1. Should there be an upgrade path for this that deletes or suggests to delete these files if they exist for a user?
2. Additional documentation?

All feedback is welcome. Thanks!

### DIFF
--- a/phing/files/deploy-exclude.txt
+++ b/phing/files/deploy-exclude.txt
@@ -14,6 +14,7 @@
 /docroot/sites/*/files
 /docroot/sites/*/private
 /docroot/themes/contrib
+/docroot/.gitignore
 /drush/contrib
 /drush.wrapper
 /project.yml
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -44,6 +44,7 @@
 
   <target name="setup:composer:install" description="Installs project dependencies, including drupal core and contrib.">
     <exec dir="${repo.root}" command="export COMPOSER_EXIT_ON_PATCH_FAILURE=1; composer install --ansi --no-interaction" logoutput="true" checkreturn="true" level="info" passthru="true" />
+    <exec dir="${repo.root}" command="composer drupal-scaffold --ansi --no-interaction" logoutput="true" checkreturn="true" level="info" passthru="true" />
   </target>
 
   <target name="setup:drupal:settings" description="Create local settings files using default settings files.">
--- a/scripts/blt/ignore-existing.txt
+++ b/scripts/blt/ignore-existing.txt
@@ -5,6 +5,7 @@ project.yml
 composer.json
 README.md
 drush/site-aliases/aliases.drushrc.php
+docroot/.gitignore
 docroot/sites/default/services.yml
 factory-hooks/pre-settings-php/includes.php
 tests/behat/behat.yml
--- a/template/composer.json
+++ b/template/composer.json
@@ -59,8 +59,9 @@
     }
   },
   "scripts": {
-    "install-phantomjs": "PhantomInstaller\\Installer::installPhantomJS",
     "blt-alias": "blt install-alias -Dcreate_alias=true",
+    "drupal-scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
+    "install-phantomjs": "PhantomInstaller\\Installer::installPhantomJS",
     "post-install-cmd": [
       "PhantomInstaller\\Installer::installPhantomJS"
     ],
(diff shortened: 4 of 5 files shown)


dpagini:
Not too familiar with these Travis tests. Maybe this just needs to be restarted? I seem to not have permissions to try that.
It's erroring on a panelizer behat test. I don't _think_ my change would cause that, but not sure. 

Anyways, I'm going to hold off trying to fix that just to get a few thoughts on this change. If it's not the direction the project wants to move in, then I'll close the PR and won't try to fix. If this is something to pursue, I'm happy to help make changes. Hope that's OK? Thanks!

grasmash:
Hm, I do like the idea of keeping unmodified core files out of the repository.

However, I am concerned that with this change, developers will be confused as to how they should commit scaffold files that they have customized. `.htaccess` and .`robots.txt` being the two most commonly modified files. And as you said, when upstream changes to modified files are made, those changes will not be apparent.

dpagini:
@grasmash Yep, I agree on the changes not being apparent.
Do you think this is just a matter of documenting? If we explain somewhere how to make changes to these files (maybe in documentation AND a comment in the `.gitignore` file itself?) it still gives the developer the two options: a) add the file to the repo (by removing a line from the `docroot/.gitignore` file), and know it will be updated/overwritten with every `drupal/core` update, or b) add the file to the repo, and manually ignore the `drupal/core` updates by adding a line to `composer.json`. 
This would, in effect, make the default behavior of a new repo to use the default drupal-scaffold files from the upstream, which I think is the best approach.

As a side note, I noticed an update to `development.services.yml` get merged in. Being a drupal-scaffold file, this is a great example of various approaches. I know that file is _there_ to use in development, but I actually created a sibling file called `twig.development.services.yml` and included that in my local settings, via:
```php
$settings['container_yamls'][] = $dir . '/docroot/sites/twig.development.services.yml';
```
This way, I can let drupal-scaffold continue to manage all my main drupal files (.htaccess, as you mention, is the tough one). Anyways, just wanted to share my approach.

I'm happy to modify this PR further if you would like to see any changes. Otherwise, I can close it if you don't think this is a good addition to the project.

greylabel:
I think this issues could probably be closed if you use this approach: https://github.com/acquia/blt/pull/1647#issuecomment-308277177

grasmash:
This issue has sat open for a long time, and there doesn't seem to be any strong momentum for completing it. I'm inclined to close it for now unless there is a compelling reason to get these files out of version control.

PR #438, opened 2016-09-22
Remove root detection to allow drush to run remotely

### LINKED ISSUE #426: Remote Drush commands from BLT repo root
I am running into a problem with the `/drush.wrapper` file. I'm not sure if there's a good way to fix this, but I wanted to bring it up.

So, `drush.wrapper` will call the `vendor/bin/drush.launcher` executable and pass in the "-r $DOCROOT" (which is the project/local docroot) for what I can tell is all invocations of the drush command.
The problem I am seeing is that calling `$ drush @alias status` from my repo root causes me to see the simple output of status, anytime you run it from a non-Drupal directory.

Here are the commands I'm running, from each place, and the output:
* from `/repo/root/` I run `$ drush status`
 * I get the status of my local site. I believe this is the intended behavior of the drush.wrapper file, to be able to run a drush command from anywhere in the repo.
* from `/repo/root/` I run `$ drush @alias status`
 * I get the _basic_ drush output, like it can't find the Drupal install. I see 3 PHP variable outputs (php exec, config, os), and 5 drush variable outputs (drush script, version, config, temp directory, alias file). This is what I think is _wrong_. 
* from `/repo/root/docroot` I run `$ drush status`
 * Same as first output.
* from `/repo/root/docroot` I run `$ drush @alias status`
 * Same as first output.

When I run the first two commands with the "-debug" flag, it seems that the remote drush operation is being called with the LOCAL root in the alias example, instead of the remote root that is specified in my alias file. I'm not sure why this would be? Running the same command IN the project docroot lets those values be set correctly, though.

Here is the obfuscated output of the alias commands, first in the repo root, then in the docroot.
Command: `/var/www/mysite$ drush @alias status`
````
Calling proc_open(ssh -p 22 mysite@server.acquia.com 'env COLUMNS=124 drush  --config=/var/www/mysite/drush/drushrc.php --alias-path=/var/www/mysite/drush/site-aliases --root=/var/www/mysite/docroot --debug --user --uri=mysite.com --verbose  core-status   --e --g 2>&1' 2>&1);
````
Command: `/var/www/mysite/docroot$ drush @alias status`
````
Calling proc_open(ssh -p 22 mysite.@server.acquia.com 'env COLUMNS=124 drush  --debug --user --root=/var/www/html/mysite/docroot --uri=mysite.com --verbose  core-status   --e --g 2>&1' 2>&1);
````
_Notice in this 2nd command that the `--root` flag is set correctly to the **remote** docroot._

I hope this makes sense? Please let me know if I can clarify anything here...? Is this a necessary evil, or is there anything I can try to work around this?

### PR DESCRIPTION
**Pull request by steveworley**
This fixes #426.

By removing the directory detection and passthrough we remove the ability to run `drush` commands from the BLT project root and this will now need to be run from the `docroot` directory. However this allows Drush to correctly infer the `root` from defined aliases so if you set up a local drush alias you can still run `drush @local` from the BLT project root.

This fixed issues I was having with running remote drush commands and I don't think it's necessary to detect and pass in the root directory with the `-r` option.

This is working with global and local drush installs, ie.

```
./drush.wrapper @local status
./drush.wrapper @dev status
```

and

```
drush @local status
drush @dev status
```

### DIFF
--- a/template/drush.wrapper
+++ b/template/drush.wrapper
@@ -37,15 +37,4 @@ if [ ! -f ${DIR}/vendor/bin/drush ]; then
   composer install
 fi
 
-if [ ! -f ${DIR}/vendor/bin/blt ]; then
-  echo "BLT was not found in this project's vendor/bin directory."
-  echo "Attempting to run composer install. This takes a few minutes."
-  composer install
-fi
-
-DOCROOT=$(${DIR}/vendor/bin/blt echo-property -Dproperty.name=drush.root -S -e | tail -1)
-if [ -z "$DOCROOT" ]; then
-  vendor/bin/drush.launcher --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"
-else
-  vendor/bin/drush.launcher -r $DOCROOT --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"
-fi
+vendor/bin/drush.launcher --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"

### CODE REVIEW

**Comment by grasmash** [comment]
Hm. Very tempting to merge this, but I hesitate to give up this feature. Will need to think on it.

**Comment by steveworley** [comment]
The pull request could be changed to match:

``` shell
DOCROOT=$(${DIR}/vendor/bin/blt echo-property -Dproperty.name=drush.root -S -e | tail -1)
if [[ $1 == @* ]]; then
  vendor/bin/drush.launcher --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"
else
  vendor/bin/drush.launcher -r $DOCROOT --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"
fi
```

Which looks for `@site` aliases and uses that to determine if it needs to provide the docroot. According to Drush the alias needs to be the first param so this shouldn't change. 

My only reservation with it is that it obfuscates what is happening and might lead to some confusion for projects that don't use BLT.

dpagini:
I _sort of_ like that latest comment approach. Could/should there just be a comment in there to just say what this line is doing? Isn't this better than removing the ability to run drush from the root?

```sh
DOCROOT=$(${DIR}/vendor/bin/blt echo-property -Dproperty.name=drush.root -S -e | tail -1)
# check if drush command has an alias specified (and therefore, root is specified in alias file)
if [[ $1 == @* ]]; then
  vendor/bin/drush.launcher --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"
else
  vendor/bin/drush.launcher -r $DOCROOT --config=${DIR}/drush/drushrc.php --alias-path=${DIR}/drush/site-aliases "$@"
fi
```

PR #452, opened 2016-09-23
Pinning Drupal Console to beta4.

grasmash:


grasmash:
Root cause is fixed in console rc1.

Issue #450, opened 2016-09-23
Features to be enabled and reverted before Executing config import


vaibhavjain-in:
The current BLT integration executes Db updates and then enable and revert features.
In "setup:update", should we enable and revert the features before config import and executing Db updates ?
Reason 1 : We are using lightning, and it adds its own set of features, which we do not want, and we are removing it in the config. As of now, configs are imported first, and the Db updates are executes. Post this features are enabled and reverted, which causes some configs created by features to come back, and override what we want. Thus, failing the builds, as our test cases wont pass.
Reason 2 : This might be another reason I would like features to be enabled and reverted before config and db updates. If I add some update hooks to any of the feature, it will not be called, some  changes will not be incorporated during builds.

I am not very sure if the order is done some specific reason.

grasmash:
Sounds fine to me. @danepowell @aweingarten would be most likely to know if this might be problematic.

vaibhavjain-in:
Cool, if we are good to go, here is the PR - https://github.com/acquia/blt/pull/453

greylabel:
One best-practice in D7, which should hold true in D8, is to run database update hooks (`updb`) before running any config revert tasks (`fra`). The reason for this is you may need to make schema changes in `updb` that your configuration changes will depend on before they can be implemented successfully with `fra`. I think the order in which you run `config-import` vs `fra` more depends on which you want to take precedence in your individual use case/dev workflow.

danepowell:
@greylabel nailed it, you always want to run updb _before_ fra. Schema updates must always happen first.

@vaibhavjain-in it sounds like your use case might be better addressed by https://github.com/balsama/acme-lightning-extend

grasmash:
Given feedback from @greylabel and @danepowell, it seems that it would not be prudent to merge this change. This should be handled in project-specific update hook when the need arises.

Closing.

vaibhavjain-in:
Yes, it looks to more of a project specific need, and how you design it your project.
@danepowell correct me If I am wrong here. If we add a module to be installed after lightning is installed (delete additional roles), that may not help here, because, we are doing `fra`, which will revert the changes that we have made after lightning installation, via our custom module.

danepowell:
[...]

PR #453, opened 2016-09-23
450: Moved feature enable and revert before config import.

vaibhavjain-in:


phing/tasks/setup.xml
```diff
@@ -206,6 +206,16 @@
   </target>
 
   <target name="setup:update" description="Update current database to reflect the state of the Drupal file system.">
+    <!-- Enable features module. -->
+    <drush command="en" assume="yes" alias="${drush.alias}">
+      <param>features</param>
```
grasmash: We don't actually have the features module in our composer.json. Will this cause a failure if features doesn't exist? Perhaps we should add it to template/composer.json.

grasmash: Additionally, I think we should run this target in BLT's .travis.yml after `blt acsf:init` to at least verify that it does not return a non-zero exit code.

grasmash:
Closing. See #450.

PR #454, opened 2016-09-23
Make git commit-msg hook configurable.

jrbeeman:
As a developer, I would like to make the commit-msg git hook configurable so that I don't have to provide my own git hook setup process that is specific to my project. In particular, I want to configure the regex used to check the commit message format, and provide an alternative instructional response upon failure.

I think this will require some discussion and work to get a solution that doesn't break current installations.

### DIFF
--- a/template/project.yml
+++ b/template/project.yml
@@ -20,6 +20,16 @@ git:
     # Defining git remotes allows builds deployed via CI.
     - bolt8@svn-5223.devcloud.hosting.acquia.com:bolt8.git
     # - radass4@svn-11692.prod.hosting.acquia.com:radass4.git
+    commit_message:
+      regex: "^$BLT-[0-9]+(: )[^ ].{15,}\."
+      instructions: |
+        Invalid commit message. Commit messages must:
+
+        * Contain the project prefix followed by a hyphen
+        * Contain a ticket number followed by a colon and a space
+        * Be at least 15 characters long and end with a period.
+
+        Valid example: BLT-135: Added the new picture field to the article feature.
 
 drush:
   # You can set custom project aliases in drush/site-aliases/aliases.drushrc.php.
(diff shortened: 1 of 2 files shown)


jrbeeman:
It looks like the regex introduces test failures, but I'm unable to reproduce this issue locally within a project at the moment.

grasmash:
Looks like it's having some issues parsing project.yml:

    Previous IOException: Unable to parse contents of /home/travis/build/acquia/blt-project/project.yml: Unable to parse at line 7 (near "{  }")

grasmash:
I suspect that after running `blt update` the regex may be reformatted in a way that renders it unparsable.

template/project.yml
```diff
@@ -20,6 +20,16 @@ git:
     # Defining git remotes allows builds deployed via CI.
     - bolt8@svn-5223.devcloud.hosting.acquia.com:bolt8.git
     # - radass4@svn-11692.prod.hosting.acquia.com:radass4.git
+    commit_message:
+      regex: "^$BLT-[0-9]+(: )[^ ].{15,}\."
```
grasmash: Maybe using single quotes will address the syntax issue.

jrbeeman:
Still failing after changing to single quote. Perhaps we should tackle the larger discussion about approach from @pinterest-subsites-travis-ci: Is this an appropriate solution to the problem?

steveworley:
I agree that BLT should provide some default functionality for git-hooks but ultimately should allow projects to configure targets to run during each hooks.

I've started a branch to extend the pre-commit task in #474. At this stage, it only allows you to define custom phing targets to run during the pre-commit stage. I think this is okay, projects can defined custom include files which can open up all sorts of Phing magic which can be included in `project.yml`, but definitely interested in the conversation around this.

jrbeeman:
I think we should close this PR in favor of the discussion happening in #628.

grasmash:
Alright then. #628 was closed with the merging of #692.

PR #461, opened 2016-09-27
Temporarily disabling Lightning Behat tests.

grasmash:


grasmash:
Due to upstream failures.

PR #470, opened 2016-09-28
Allow configurable PHPUnit testing

steveworley:
It would be useful to be able to run PHPUnit testing from `pre-commit` so project developers can get an idea of build status before relying on CI.

This change allows you to specify test directories for different environments via a `phpunit` property in **project.yml**. It would look like:

``` yml
phpunit:
   local:
      - path/to/my/module/tests
   deploy:
      - path/to/more/tests
```

This allows custom modules to bundle unit tests in the module directory as per Drupal standards and still have them be executed during builds, if none are defined it will default to . 

```
${repo.root}/tests/phpunit,${blt.root}/tests/phpunit/BltProject
```

It also allows you to remove the BLT unit testing from your project, a project shouldn't really run the BLT unit tests as it is implicit that the BLT project should be green and a project shouldn't make changes.

Another change to allow customisation of the pre-commit hook will follow which will rely on this commit.


phing/build.xml
```diff
@@ -1,4 +1,5 @@
 <project name="blt" default="list" phingVersion="2.8.2">
+  <property environment="env"/>
```
grasmash: Aren't env properties already available under ENV?

phing/tasks/tests.xml
```diff
@@ -83,21 +83,45 @@
   <target name="tests:phpunit" description="Executes PHPUnit tests against custom code.">
     <delete dir="${reports.localDir}/phpunit" quiet="true"/>
     <mkdir dir="${reports.localDir}/phpunit"/>
+    <if>
+      <and>
+        <isset property="phpunit.local" />
+        <not><isset property="env.TRAVIS" /></not>
```
grasmash: I'd prefer not to add any Travis specific logic to our tasks. Pipelines will replace Travis in the future.

phing/tasks/tests.xml
```diff
@@ -83,21 +83,45 @@
...
+    <foreach list="${phpunit.dirs}" param="dir" target="tests:phpunit:dir"/>
+  </target>
 
+  <target name="tests:phpunit:dir" description="Run phpunit against given directories." hidden="true">
     <phpunit printsummary="true" failureproperty="phpunit.failures" errorproperty="phpunit.errors">
       <formatter type="plain" usefile="false"/>
       <formatter type="xml" todir="${reports.localDir}/phpunit"/>
       <batchtest>
-        <fileset dir="${repo.root}/tests/phpunit">
+        <fileset dir="${dir}">
```
grasmash: Rather than specifying dirs for local and deploy as a means of grouping tests, why not use PHPUnit's built-in ability to group tests with `@` tags? We can make use of the [phpunit task's](https://www.phing.info/docs/stable/hlhtml/index.html#PHPUnitTask) group and excludeGroup params to accomplish this.

Following the established naming conventions for BLT targets, we can then have `ci:tests:phpunit` and `local:tests:phpunit` targets that simply pass params value for group and excludeGroup to `tests:phpunit`. 

Users can also then directly call `tests:phpunit -Dphpunit.groups='' -Dphpunit.excludeGroups=''` to set custom values.

Even with this approach, it would still be great to turn phpunit.dirs into a variable and set a default value in build.yml. E.g.,

    phpunit:
      dirs: 
        - ${repo.root}/tests/phpunit

This would allow users to easily override the directories by setting a custom value in project.yml.

phing/build.xml
steveworley: Agreed, I thought this too - but it wasn't available to me during testing unless I added this. According to the docs you need to add this if you want to access `env` https://www.phing.info/trac/ticket/952

phing/tasks/tests.xml
steveworley: @grasmash I like this idea. I've updated the pull request so that it includes a few new targets to give us group support to phpunit.

Following your recommendation you can now define groups in the phpunit hash  in project.yml

```
phpunit:
  local:
     groups:
         - module
```

Unfortunately the PHPUnit phing task doesn't accept all configuration values defined in a phpunit.xml file, some information can be [found here](https://www.phing.info/trac/ticket/1233) and the [phpunit task docs](https://www.phing.info/docs/stable/hlhtml/index.html#PHPUnitTask) which suggest that only bootstrap, processIsolation, stopOnFailure, stopOnError, stopOnIncomplete and stopOnSkipped are supported.

So I couldn't use the phing task to execute the Drupal groups (this requires configuration from phpunit.xml.dist in `docroot/core`. To work around we can execute phpunit directly and pass in the required options.

So the approach is; I've defined 2 targets, `local:tests:phpunit` and `ci:tests:phpunit` which read values from *project.yml* (defined as above). They execute the `tests:phpunit:groups` target which is expecting the `phpunit.groups` and `phpunit.excludeGroups` properties. This allows us to execute group tests like:

```
blt local:tests:phpunit
blt ci:tests:phpunit
blt tests:phpunit:groups -Dphpunit.groups=module -Dphpunit.excludeGroups=module2
```

grasmash:
This must be refactored for Robo port.

PR #471, opened 2016-09-28
Removing hosting flag for acsf.

grasmash:


dpagini:
Is there a new alternative to this? Is "ACSF" hosting still supported?

grasmash:
A recent ACSF update now permits using a custom settings.php file, obviating the need for disparate artifacts for ACSF vs ACE. This PR has not yet been tested on ACSF. Will merge when we confirm that this works as expected.

dpagini:
Oh awesome. I will take a look at this tomorrow and I can try it out as well.

PR #474, opened 2016-09-29
Allow precommit tasks to be defined in project.yml

steveworley:
Updates pre-commit hook to call a BLT task which will look for steps in the project.yml file. This will allow projects to define additional checks to enforce during the precommit stage.

Might need to change the key in `project.yml` from _targets_ currently it will expect a list of targets as they are currently defined in BLT.

``` yml
targets:
  precommit:
    - validate:phpcs:files
    - tests:phplint
```

With this change, I was a little cautious to allow properties to be passed around given the nature. As this means we will need to define a target for each entry point and document which targets get which parameters.


grasmash:
Closing in favor of #692, which allows custom git hooks to be implemented.

PR #502, opened 2016-10-06
Fixes #501: Default Drupal VM to PHP 5.6.

### LINKED ISSUE #501: Default DrupalVM to PHP 5.6
Currently the blt vm target installs a vanilla drupalvm which includes ubuntu 16 and PHP7.  For parity with acquia cloud it should instead install ubuntu 1404 and PHP 5.6.

cc: @geerlingguy

geerlingguy:
See https://github.com/acquia/blt/issues/501

### DIFF
--- a/scripts/drupal-vm/config.yml
+++ b/scripts/drupal-vm/config.yml
@@ -2,6 +2,9 @@
 vagrant_hostname: ${project.local.hostname}
 vagrant_machine_name: ${project.machine_name}
 
+# Use Ubuntu 14.04 LTS so PHP 5.6 can be installed.
+vagrant_box: geerlingguy/ubuntu1404
+
 # Set drupal_site_name to the project's human-readable name.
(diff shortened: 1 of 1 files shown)


danepowell:
This broke VM initialization, because the package php-bz2 doesn't exist for Trusty (14.04). See #521

geerlingguy:
@danepowell - https://github.com/acquia/blt/pull/503 fixes this, but is waiting in a conga line of selenium/phantom-related PRs.

We could hotfix right now by setting `bz2` as a `php_pecl_extension` instead (to make sure the 8.x HEAD branch still builds the VM until/unless selenium replaces Phantom), but we would also need to remove the bcmath installed_extra (because it also doesn't actually exist for PHP 5.6 :).

cc @grasmash — basically, we need to either get #503 merged (after phantom is no longer a requirement), or we can make a separate PR to get things building in the meantime.

PR #503, opened 2016-10-06
Follow-up to #502: Remove bz2 requirement after switch to Selenium2.

geerlingguy:
Bz2 isn't required if we're not using PhantomJS... and it's harder to get it installed cleanly for PHP 5.6 anyways. Therefore I recommend we drop it once #500 gets merged.


geerlingguy:
@bhasselbeck also mentioned bcmath was only added to support PhantomJS usage... so could we maybe drop that too?

geerlingguy:
Rebased and also removed `bcmath` since (a) it's already installed through FPM, and (b) it breaks the VM build with 5.6 :)

grasmash:
Merging prior to Selenium2 because this is already broken due to Drupal VM upgrade.

PR #505, opened 2016-10-06
Fixes #504: Use newer version of Drupal VM, 3.3.x.

### LINKED ISSUE #504: Use newer version of Drupal VM (currently ~3.1)
Drupal VM 3.3.0 was just released, and there are many improvements that would benefit BLT projects; we should update the dependency to 3.3.x instead of 3.1.x. Only major change is the difference in the `drupal_mysql_*` variables being changed to `drupal_db_*` variables (since Drupal VM can now support PostgreSQL or MySQL).

geerlingguy:
See #504


geerlingguy:
Wrong local branch, see https://github.com/acquia/blt/pull/507 instead.

PR #511, opened 2016-10-06
Fixes #506: Include config in deploy artifact.

### LINKED ISSUE #506: SimpleSAMLphp config files not included in deploy artifact
## Problem/motivation
1. When `blt simplesamlphp:init` the simplesamlphp.settings.php file gets copied into the vendor/acquia/blt/settings directory. 
2. When `blt deploy ...`  is run, it checks for the existence of that settings file in  vendor/acquia/blt/settings and if it exists, `simplesamlphp:deploy:config` is called, which copies the config files into the vendor directory

The problem is that since `simplesamlphp:init` isn't run when the deploy artifact dependencies are being built, the settings file doesn't get created in the deploy artifact's vendor directory, and as a result `simplesamlphp:deploy:config` doesn't get called.

## Proposed resolution
When `blt simplesamlphp:init` is run it needs to modify something outside the vendor directory which can be checked during `blt deploy:build`. However the only modifications that are made outside the vendor directory are modifications that have a high likelihood of being changes that would be made by a simplesaml installation that wasn't initiated by BLT. 

So it seems to me that the logical thing to do would be to set a property inside the project.yml file.

dooleymatt:
See #506


grasmash:
I'm going to merge this but I may rename the variable. I'd like to nest features like this under a common array key like "features"--but that's already taken.

dooleymatt:
Okay thanks. Let me know if/when you rename the variable because the simplesaml targets and the BltDoctorCommand will need to be updated.

Issue #509, opened 2016-10-06
Update Documentation for BLT under WSL


geerlingguy:
Running BLT under the Windows Subsystem for Linux (Ubuntu Bash) is a little more complicated than is currently stated in the documentation—though there are some parts that I think we can consolidate and make simpler.

I'm working on rebuilding BLT + Drupal VM a few times on my old Lenovo running Windows 10, and I found a few extra tips and tricks to help get everything working correctly.

I'll submit a PR with proposed changes/additions, and I'll also add a few notes to this issue for posterity/background.

geerlingguy:
Eh... one weird thing—I'm getting the following error message whenever Drupal VM hits the `Run drush to finish setting it up` task:

```
Drush was not found in this project's vendor/bin directory.
```

And it tries to install _all_ the project dependencies (including phantomjs-installer and a zillion other things) instead of just... running `/usr/local/bin/drush`.

The only place on the internet I've seen that message is in here: https://github.com/mariagwyn/mariagwynBLT/blob/master/drush.wrapper — what's this used for?

Also, how is it that if I `vagrant ssh` into the VM and literally run `/usr/local/bin/drush`, Drush seems to work just fine, printing out it's built-in help?

geerlingguy:
Ah, that's just a fork... it's in https://github.com/acquia/blt/blob/8.x/template/drush.wrapper

geerlingguy:
So I guess @grasmash how is this thing supposed to work, and how is it possible that running the command `/usr/local/bin/drush` during Vagrant provisioning is getting wrapped by this wrapper?

grasmash:
@geerlingguy When the `drush` command is run from within the repository, it looks for and runs drush.wrapper. The wrapper script tells the global install of drush to use the repo's vendor/bin/drush as its binary. It essentially overrides global drush. However, we need to make sure that composer dependencies have actually been built before attempting to execute vendor/bin/drush, hence the call to `composer install`. We might be able to change this to`composer install drush/drush`.

geerlingguy:
I'm trying to pull down the latest dev:

```
composer create-project acquia/blt-project test --stability dev --no-interaction
```

But that doesn't seem to work :/

geerlingguy:
Ah, just realized blt-project is not BLT... https://github.com/acquia/blt-project/blob/8.x/composer.json

To get the latest commit, I have to clone BLT manually and run the commands separately.

geerlingguy:
[...]

PR #516, opened 2016-10-07
Issue #509: Use default hostname that works more universally.

geerlingguy:
It looks like `*.localhost` domains don't work on Windows. Under Chrome and IE, with all the default settings, I would get "CONNECTION_REFUSED".

If I switch to anything without `.local` in the domain, things just work. So I figure why not use the pattern `local.[machine-name].com`.

This has two advantages:

  1. It _works_ (I've had issues with .local and .localhost in myriad environments—not just Windows).
  2. If you have a domain like "www.example.com", with a stage env of "stage.example.com", now your local environment would be "local.example.com". So much simpler to reason about and program around if automating anything (e.g. settings.php, build scripts, etc.) since it's just `[environment].example.com`.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,7 +7,7 @@ This document outlines the process for creating a new BLT release.
 In order to use these testing instructions:
 
 * The `blt` alias must be installed.
-* Your LAMP stack must have host entry for `http://blted8.localhost` pointing to `./blted8/docroot`.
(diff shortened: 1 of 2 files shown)


greylabel:
Using a local domain structure that does not resemble your production services can sometimes be a bit of an anti pattern -- potentially does not play as well with cookie sharing, CORS, and other features encountered in enterprise builds.

PR #542, opened 2016-10-11
Include Instructions for VM configuration / customization on Create New Projects Page

mikemadison13:
currently the Create New Project page contains no linkage to the instructions for Drupal VM and/or Acquia Dev Desktop. This change slightly modifies the instructions for setting up a new project to ensure these customizations are completed.


grasmash:
I'd prefer to assume that Drupal VM is being used. We can say something like "If you'd like to use Drupal VM, run the following commands. Otherwise please see [link]."

mikemadison13:
I've updated the text and re-commited to be more inline with the above.

PR #555, opened 2016-10-13
Cleaned up SimpleSAML_php integration.

danepowell:
The drupal yaml update command fails because that command isn't generally available globally.

@grasmash on a tangential note, are we supposed to be submitting PRs against 8.x-dev now? Should we update the contribution guidelines to reflect that?


grasmash:
@danepowell to 8.x-develop, yes. Still verifying the workflow.

PR #563, opened 2016-10-14
Fixes #562. Add a version target to the blt phing file to display version info.

### LINKED ISSUE #562: Add BLT command to show current version
It is only possible to view the current blt version by using composer show command.

Add a phing target that wraps this command to enable 'blt version'. (using the -version shows the phing version).

gollyg:


phing/tasks/blt.xml
```diff
@@ -73,6 +73,10 @@
     <exec dir="${repo.root}" command="${blt.root}/scripts/blt/remove-deprecated.sh" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true" />
   </target>
 
+  <target name="version" description="Display the currently installed version of BLT.">
+      <exec dir="${repo.root}" command="echo 'BLT version:' $(composer show | grep blt | awk '{print $2}')" logoutput="true" checkreturn="true"/>
```
grasmash: This would be a bit faster if you run `composer info acquia/blt` and then grep the version line.

grasmash: I'd also add a `level="${blt.exec_level}"` so that it conforms to the BLT verbosity settings.

gollyg: Changed, and pushed. Thanks for the tips.

PR #579, opened 2016-10-18
Tweak simplesaml config.

danepowell:
Right now, acquia_config.php is basically cruft left over from the config setup. This is a little confusing because some people (i.e. yours truly) might assume that it's being dynamically included somehow.

This makes it a little cleaner.

I still need to test this.

### DIFF
--- a/phing/tasks/simplesamlphp.xml
+++ b/phing/tasks/simplesamlphp.xml
@@ -74,7 +74,7 @@
     ============================================================================
 
       * ${docroot}/.htaccess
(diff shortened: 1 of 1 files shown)


grasmash:
@danepowell tell me when you've tested it :)

danepowell:
Just tested and working great!

@dooleymatt feel free to comment since you wrote this originally but I'm pretty sure it's the right way to go.

dooleymatt:
@danepowell yeah, much better.

PR #589, opened 2016-10-18
Run BLT commands on ACE

danepowell:
If you run `config-import` without a parameter, it defaults to the sync directory, as does BLT. This just makes that parameter configurable.

This is the first step in being able to run `local:update` / `setup:update` as easily on ACE as you can locally, which will make deployments way more reliable.


grasmash:
Hm. Can you add some docs to indicate how this should be used?

danepowell:
Yeah maybe I'll make this a somewhat more comprehensive PR. There's going to be other required changes (BLT needs to be run within a Git repo, requires project.yml, etc...)

danepowell:
I'm going to hold on this until after 8.5.0 is released and I can upgrade existing projects, since it conflicts with other patches we need as part of that release.

danepowell:
This is functional but I might try to add some documentation or expand the scope to include example cloud hooks.

phing/tasks/deploy.xml
```diff
@@ -156,4 +156,12 @@
       </fileset>
     </delete>
   </target>
+
+  <target name="deploy:update" description="Update current database to reflect the state of the Drupal file system; uses local drush alias.">
```
grasmash: This is actually using "self" rather than the local alias, which is drush.aliases.local. Can you also add a comment explaining the need for the vcs value?

grasmash:
Yeah we should add some documentation in http://blt.readthedocs.io/en/8.x/readme/deploy/ outlining how to handle actual deployment to an environment. I'd love for that to eventually include building the artifact, checking out the branch to an environment using `drush @${drush.aliases.remote} ac-code-path-deploy ${deploy.branch}`, and then calling `deploy:update` on the remote, which I guess would be something like `drush @${drush.aliases.remote} ssh '../vendor/bin/blt deploy:update'`.

danepowell:
I've always run post-deploy commands automatically via cloud hooks, rather than via the API. Is it even possible to call the update command in an automated way? I imagine it would have to be from Travis... and I'm not sure how Travis (or any tool) would be able to "wait" following a code push to confirm that the code is deployed before running the updates.

phing/tasks/deploy.xml
grasmash: Ok, good to know is is Acquia Cloud enforced.

danepowell:
@grasmash I think this is ready to go. My only concern is that existing projects updating BLT may be surprised by the post-code-deploy hook suddenly running updates on deployments. My recommendation would be to just call this out in the release notes. I think the CMI, CI, and db updates workflow is something BLT should be rather opinionated about, rather than waiting for people to opt-in. But I'm interested in your thoughts.

template/hooks/common/functions.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
grasmash: If we're going to be opinionated about this, I think we shouldn't put it in the template. It should be in a blt subdir, under vendor/acquia/blt. This will allow us to roll out new functions and updates.

template/hooks/common/functions.sh
```diff
@@ -0,0 +1,38 @@
...
+
+  echo "Update starting..."
+  blt deploy:update
+  if [ $? -ne 0 ]; then
+      echo "Update errored."
+      status=1;
+  fi
+  echo "Updates complete."
+
+  # if [ "$target_env" != "prod" ]; then
```
grasmash: This should probably be split into a separate function like enable_dev_tools()

phing/tasks/deploy.xml
grasmash: Still need the comments regarding VCS value.

tests/phpunit/Blt/DeployTest.php
```diff
@@ -52,7 +52,6 @@ public function testBltDeployBuild() {
     // Ensure non-required files are not in deploy dir.
     $this->assertFileNotExists($this->deploy_dir . '/blt.sh');
     $this->assertFileNotExists($this->deploy_dir . '/build');
-    $this->assertFileNotExists($this->deploy_dir . '/project.yml');
```
danepowell: Fixed.

template/hooks/common/functions.sh
```diff
@@ -0,0 +1,38 @@
...
+  drush_cmd="drush @$site.$target_env"
+  echo $drush_cmd
+
+  # Prep for BLT commands.
+  repo_root="/var/www/html/$site.$target_env"
+  export PATH=$repo_root/vendor/bin:$PATH
+  cd $repo_root
+
+  echo "Update starting..."
+  blt deploy:update
```
danepowell: I'm currently using this on several "vanilla" ACE sites (no pipelines or livedev), and I haven't seen any errors related to Git. Where are you seeing the problem?

timcosgrove: (I'm working with @greylabel on a platform build and we're making use of BLT.)

I see why your example was working: you manually CD to the 'repo root' in the script (in quotes since it's not a proper repo), which means vendor/bin/blt is able to find vendor/acquia/blt/blt.sh, and therefore doesn't fail. That works, but it would be nice to minimally figure out how to make vendor/bin/blt invokable from anywhere in the build.

timcosgrove: You could equivalently identify the location of blt.sh and invoke that, since that's all vendor/bin/blt really does anyway.

danepowell: Ah that makes sense. Yeah, changing the directory before running the command does feel a bit janky, but I must have not been able to come up with a better solution (sorry, my memory is rusty as to why).

timcosgrove: No worries! This is hard :) I've been sort of struggling with how to get BLT happy to operate inside of ACE myself.

template/hooks/common/functions.sh
danepowell: I refactored this file and moved it into BLT core.

template/hooks/common/functions.sh
danepowell: Yeah actually we should just rely on Phing for this... see my last comment here...

template/hooks/common/functions.sh
danepowell: Do you think this is something we need to remediate before this is merged, or can we clean it up later? Like I said, it seems to be working for me as-is, even though I know it's not ideal.

danepowell:
I think I've resolved all of the concerns in this thread, I just need to test the latest refactor around the cloud hooks and will report back later today.

danepowell:
Working great!

danepowell:
Woot, thanks!

timcosgrove:
@danepowell Thank you for this. Already started using it, and it's working beautifully.

PR #591, opened 2016-10-19
Removed duplicate dependencies.

danepowell:
The template composer.json defines a direct dependency on several packages that are already dependencies of other packages:
- Drupal Console (dependency of BLT)
- Drupal Coder (BLT)
- PHPUnit (BLT)
- PHPCS (BLT)
- Composer installers (Drupal Console and Lightning)

Because these dependencies are effectively defined twice, composer throws a warning when you run composer update on a BLT project:

> Dependency "drupal/coder" is also a root requirement, but is not explicitly whitelisted. Ignoring.

Right now I think this is a harmless warning, but it could lead to unpredictable behavior if BLT, Drupal console, or Lightning updates or patches one of these dependencies.

Seems like the best solution is to remove them from the project template composer.json, and just allow upstream projects to manage them.

### DIFF
--- a/template/composer.json
+++ b/template/composer.json
@@ -8,7 +8,6 @@
         }
     ],
     "require": {
-        "composer/installers":                        "^1.0.20",
         "cweagans/composer-patches":                  "dev-master#5456199acdcc16b243814aa45259f4c56272a634 as 1.5.0",
         "drupal/acquia_connector":                    "^8.1.5",
         "drupal/acsf":                                "^8.1.32",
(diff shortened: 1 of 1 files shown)


danepowell:
I've confirmed that removing these dependencies from a generated BLT project doesn't actually change the composer.lock, confirming that they were basically superfluous.

template/composer.json
```diff
@@ -8,12 +8,10 @@
         }
     ],
     "require": {
-        "composer/installers":                        "^1.0.20",
```
grasmash: This is not in BLT's composer.json. Does it get pulled in by drupal scaffold?

grasmash:
Perhaps we should move a few of these to BLT's composer.json:

* composer/installers 
* cweagans/composer-patches

danepowell:
Yeah, composer-installers comes from both Lightning and Drupal Console. But since I think BLT invokes it directly, we should move it to BLT's composer.json.

PR #601, opened 2016-10-24
Adding updater for specific BLT delta migrations.

grasmash:


grasmash:
This is an untested POC.

PR #602, opened 2016-10-25
Added sample patches for SSL and SAML.

danepowell:
I originally was attempting to automate SimpleSAML setup as part of this as well, but couldn't quite get it to work, so I spun it into a new PR. Let's just get the example patches in for now.


danepowell:
I'm still testing the simplesaml patch

danepowell:
I'm sorry to say that this probably won't work... it appears that drupal-scaffold won't let you patch core files like .htaccess: https://github.com/drupal-composer/drupal-scaffold/issues/45

PR #607, opened 2016-10-28
Automated SimpleSAMLphp htaccess patch (DO NOT MERGE)

danepowell:
I'm still testing this, but I've already fixed one of the bugs from last time. I didn't realize that the composer munge command arguments were ordered.


danepowell:
Ugh. It turns out this probably isn't going to work, so I'm closing this PR. We can follow up in #608

Issue #609, opened 2016-10-30
Update blt template files to match Drupal core.


greylabel:
BLT provides in its _template_ copies of several Drupal core files. These files should be updated to match the current release of Drupal (8.2.1).

grasmash:
See #610

PR #610, opened 2016-10-30
Update blt template files to match Drupal 8.2.1.

greylabel:
See: https://github.com/acquia/blt/issues/609


template/docroot/sites/development.services.yml
```diff
@@ -5,9 +5,3 @@
 services:
   cache.backend.null:
     class: Drupal\Core\Cache\NullBackendFactory
-
```
grasmash: I'd like to keep this settings.

grasmash:
I'd like to commit as few of these as possible. I vaguely remember that template/docroot/sites/default/default.settings.php is required for installation in some instances because of the order that some operations are run. Perhaps that can be changed and we can avoid committing even that.

greylabel:
@grasmash I feel it is very valuable to keep these files in the repo as a reference as they are Core files. Much nicer to be able to refer them locally than needed to google every time.

greylabel:
@grasmash Requested changes made and changed the commit msg to match common syntax being used.

grasmash:
> Much nicer to be able to refer them locally than needed to google every time.

Aren't these files downloaded to your local machine after you run `blt setup`?

grasmash:
I've actually removed all scaffold files: #632 

If you'd like local copies for reference and they're not available at a particular stage of development, please file an issue to that effect. We can look into executing the scaffold file download earlier in the development cycle.

Issue #626, opened 2016-11-03
Allow cm.features.bundle to contain multiple feature bundles


timcosgrove:
Our build utilizes multiple Features bundles, and I suspect this will be a common potential scenario. 

The current handling of Features in `setup:update` only accounts for a single bundle. 

I will submit a PR that allows for multiple bundles to be specified in properties.

PR #627, opened 2016-11-03
Allow specification of multiple Features bundles

timcosgrove:
https://github.com/acquia/blt/issues/626

We would like to be able to specify multiple Features bundles.

I would have liked to change the property name to cm.features.bundles but left it as-is for backwards compatibility sake. I've confirmed that the ForEachTask is perfectly happy with a single item.

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -248,13 +248,17 @@
           <param>drush</param>
         </drush>
         <!-- Revert all features. -->
-        <drush command="fra" assume="yes" alias="${drush.alias}">
-          <option name="bundle">${cm.features.bundle}</option>
-        </drush>
+        <foreach list="${cm.features.bundle}" param="bundle" target="setup:update:features-import" />
       </then>
     </if>
     <!-- Rebuild caches. -->
     <drush command="cr" alias="${drush.alias}"/>
+  </target>
+
+  <target name="setup:update:features-import" description="Perform a features import on the specified bundle.">
+    <drush command="fra" assume="yes" alias="${drush.alias}">
+      <option name="bundle">${bundle}</option>
+    </drush>
   </target>
 
 </project>


timcosgrove:
https://github.com/acquia/blt/pull/625 will most likely allow tests on this to pass. Will rebase once that is merged.

grasmash:
Restarting tests. Rebase should not be necessary.

grasmash:
@timcosgrove I believe that this changes requires cm.features.bundle to be an array (it can never b e a string). 

If that is the case, we should change the variable name to cm.features.bundles and change the [example here](https://github.com/acquia/blt/blob/8.x/template/project.yml#L68) to be an array.

phing/tasks/setup.xml
```diff
@@ -248,13 +248,17 @@
           <param>drush</param>
         </drush>
         <!-- Revert all features. -->
-        <drush command="fra" assume="yes" alias="${drush.alias}">
-          <option name="bundle">${cm.features.bundle}</option>
-        </drush>
+        <foreach list="${cm.features.bundle}" param="bundle" target="setup:update:features-import" />
```
grasmash: @timcosgrove I believe that this changes requires cm.features.bundle to be an array (it can never b e a string).

If that is the case, we should change the variable name to cm.features.bundles and change the example here to be an array.

timcosgrove: @grasmash I tested it with yml formatted as such:

```yaml
cm:
  features:
    bundle: my_feature_bundle
```
This worked as expected, feeding into the foreach as an array with a single value.

I think 'bundles' is clearer myself but I wanted to not breaking existing project.yml.

timcosgrove: Testing confirmed the following forms all feed validly to the ForEachTask:

single:
```yaml
cm:
  features:
    bundle: my_feature_bundle
```
comma-delimited:
```yaml
cm:
  features:
    bundle: my_feature_bundle, my_other_feature_bundle, my_last_feature_bundle
```
per-line:
```yaml
cm:
  features:
    bundle: 
      - my_feature_bundle
      - my_other_feature_bundle
      - my_last_feature_bundle
```
My personal preferences is for the last option, but it looks like Phing is happy to take either form.

timcosgrove: And to be clear I think 'bundles' is clearer semantically but again I wanted to make sure this didn't break anyone's build.

grasmash:
K, I'm going to merge as is for the moment but I will likely change to bundles. I am concerned about backwards compatibility--we need a way to roll out delta updates that take care of changes like this.

timcosgrove:
A simple way to handle the property name change would be to make 'bundles' the default, and have a simple check somewhere that sets `bundles` = `bundle` if `bundles` is empty and `bundle` is set.

PR #640, opened 2016-11-08
Added trusted host configuration.

danepowell:
For now, this is mostly useful as a demonstration of how to customize settings for a project, and a reminder to set the trusted host.

Ultimately it would be nice if you could configure environments and domains in project.yml, and have the trusted domains and drush alias uris all be configured automatically. But that might be a lot of work for not a huge return, and it might be hard to accommodate multisite.

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -76,6 +76,7 @@
     <chmod mode="0644" failonerror="false" file="${docroot}/sites/default/settings.php"/>
     <echo>Ensuring that blt.settings.php is required by settings.php</echo>
     <exec dir="${docroot}/sites/default" command="grep vendor/acquia/blt/settings/blt.settings.php settings.php || echo 'require DRUPAL_ROOT . &quot;/../vendor/acquia/blt/settings/blt.settings.php&quot;;' >> settings.php" logoutput="true" checkreturn="true" level="${blt.exec_level}"/>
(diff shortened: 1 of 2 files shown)


phing/tasks/setup.xml
```diff
@@ -76,6 +76,7 @@
     <chmod mode="0644" failonerror="false" file="${docroot}/sites/default/settings.php"/>
     <echo>Ensuring that blt.settings.php is required by settings.php</echo>
     <exec dir="${docroot}/sites/default" command="grep vendor/acquia/blt/settings/blt.settings.php settings.php || echo 'require DRUPAL_ROOT . &quot;/../vendor/acquia/blt/settings/blt.settings.php&quot;;' >> settings.php" logoutput="true" checkreturn="true" level="${blt.exec_level}"/>
+    <exec dir="${docroot}/sites/default" command="grep trusted_host.settings.php settings.php || echo '// require DRUPAL_ROOT . &quot;/sites/default/settings/trusted_host.settings.php&quot;;' >> settings.php" logoutput="true" checkreturn="true" level="${blt.exec_level}"/>
```
grasmash: Hm. I'm not crazy about doing this by default. I love having a very small footprint in settings.php. If we do this, I think we need to add some checks either in the form of PHP Unit tests or BLT doctor assertions.

danepowell: I get that, it's not a huge win on its own. Note that the code I added here is commented out, it has no effect on its own.

I just know that we get a lot of questions along the lines of "how do I add custom configuration for x?" The easiest and most intuitive way of demonstrating this is by doing it. Alternatively, we could just add documentation somewhere.

PR #646, opened 2016-11-09
Updating Console to rc8.

grasmash:


grasmash:
Failure is due to https://github.com/hechoendrupal/DrupalConsole/issues/2905

bkosborne:
Does BLT require the use of Console to execute tasks? I think I was surprised to see it as a dependency. It being locked to beta5 is seemingly conflicting with the latest version of Metatag module (beta11) preventing us from using console at all =/

I see that you're trying to push forward the main blocking issue in Console tho.

grasmash:
@bkosborne Yes we use the yaml:update command in console. Although, I'm considering just porting the command to BLT directly to remove the dependency. We've been blocked on an update for some time now.

geerlingguy:
Also, not having a later version is causing some issues with the current project I'm on; there are bugs we're hitting that have been fixed in newer versions that we can't work around since we're stuck on the older version :(

geerlingguy:
Also, rc11 is out now.

PR #665, opened 2016-11-14
Adding validate:twig target.

grasmash:


bobbygryzynger:
@grasmash - excited to see this go in!! When do you think the next stable release will be?

PR #667, opened 2016-11-15
ReadTheDocs is not good at Markdown

TravisCarden:
GitHub and ReadTheDocs don't parse Markdown the same way. Let's see if this makes ReadTheDocs happier.


grasmash:
Rather than using triple backticks, RTD requires you to simply indent code blocks.

TravisCarden:
Like this, @grasmash?

grasmash:
Probably. If this doesn't work, we can just push your branch to origin and set RTD to build the feature branch so that you can iterate without requiring a merge.

PR #671, opened 2016-11-15
Resolves #670: Toggle modules after config import.

### LINKED ISSUE #670: How to handle D8 config with local/deploy modules
The `local:refresh` task will invoke `setup:update` and pass along a list of modules to uninstall from the project.yml file. This is great, except it doesn't really work properly since immediately after the modules are disabled, the config is imported via drush. This basically performs an "undo" operation since the imported config has those modules enabled.

This is sort of a generic problem I'm facing with D8 but something that's readily apparent in my usage of BLT.

Anyone have this problem and a workaround?

grasmash:


grasmash:
Closing for #674

PR #684, opened 2016-11-16
Removed cloud hook samples.

danepowell:
It seems confusing and unnecessary to me that we maintain local copies of cloud hook samples and templates that are available upstream at https://github.com/acquia/cloud-hooks

If people need these, we have a link to that upstream repo in the docs, it seems like it would be best to just maintain them there.


danepowell:
@grasmash I cleaned up the README a bit. I'm not clear on what kind of additional support we should provide here. It's easy enough to browse the cloud hooks repo and download samples on an as-needed basis. We could add new individual commands to set up things like Slack, Hipchat, or New Relic integration... but it seems like a pretty low ROI.

PR #695, opened 2016-11-21
Fixes #664: Disable cloud hooks on ACSF.

grasmash:

PR #698, opened 2016-11-21
Fixed cloud hooks.

danepowell:
#695 broke cloud hooks on ACE (they would never fire). This fixes them.

### DIFF
--- a/template/hooks/common/post-code-deploy/post-code-deploy.sh
+++ b/template/hooks/common/post-code-deploy/post-code-deploy.sh
@@ -17,11 +17,8 @@ deployed_tag="$4"
 repo_url="$5"
 repo_type="$6"
 
-# 1 if is ACSF, 0 if not.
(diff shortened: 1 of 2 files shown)


template/hooks/common/post-code-deploy/post-code-deploy.sh
```diff
@@ -17,11 +17,8 @@ deployed_tag="$4"
 repo_url="$5"
 repo_type="$6"
 
-# 1 if is ACSF, 0 if not.
-[[ ! -f "/mnt/files/$AH_SITE_GROUP.$AH_SITE_ENVIRONMENT/files-private/sites.json" ]]
```
danepowell: Additionally, I think it's safer / more portable to use `$site.$target_env` rather than `$AH_SITE_GROUP.$AH_SITE_ENVIRONMENT`, but not sure how much you want to take on in this PR...

PR #700, opened 2016-11-21
Toggle modules per environment.

### LINKED ISSUE #668: More granular control of modules per environment
Right now, you can select which modules to enable/disable for local, remote ("deploy"), and ci environments. Many (most?) projects need more granular control than this.

For instance, it's pretty typical to enable development modules in a remote dev environment. You might also want a smaller subset of development modules for the stage environment (for instance, to create dummy users or content for UAT). You wouldn't want either of these in prod.

I'm trying to think of the best way to support these scenarios, since adding even more elements to the module array in `project.yml` seems pretty clunky.

danepowell:
Fixes #668 .

Changes proposed:
- Modules will be toggled based on the value of an "environment" parameter.
- No functional change for the `ci:setup`, `ci:update`, `local:setup`, and `local:update` targets (we are simply consolidating variables)
- For `deploy:update`, you will now need to define modules to toggle per environment. But if you prefer the old style, simply don't pass the environment parameter and it will default to `deploy` like before.

### DIFF
--- a/phing/tasks/ci.xml
+++ b/phing/tasks/ci.xml
@@ -14,8 +14,7 @@
     <phingcall target="setup">
       <property name="drush.alias" value="${drush.aliases.ci}"/>
       <property name="create_alias" value="n"/>
-      <property name="enable_property" value="modules.ci.enable"/>
-      <property name="uninstall_property" value="modules.ci.uninstall"/>
+      <property name="environment" value="ci"/>
     </phingcall>
   </target>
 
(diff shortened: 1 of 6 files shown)


danepowell:
Note that at some point, we could take this even further and have an "environment" array that defines more than just modules per-environment. For example:

````
environment:
  local:
    modules:
      enable:
        - devel
      uninstall:
        - acquia_connector
    drush-alias: self
    uri: my-project.local // Use to set trusted_host pattern, base_url, etc
    caching: disabled // Use to enable/disable caching in a settings file
...
````

We probably don't want to do all of that as part of this PR, but maybe we want to at least rename the array from "modules" to "environments"?

danepowell:
You could do that, it just seems like `local` is a distinct environment. For instance, you don't want shield locally, but you do want it on dev. And I think there's nice symmetry between the environments and `ci:*` and `local:*` wrappers.

At any rate, these are just suggested environments, projects can define whatever environments they want.

Issue #720, opened 2016-12-01
Travis CI Drupal site install failing (as of today)


geerlingguy:
My system information:
* Operating system type: macOS
* Operating system version: 10.12.1
* BLT version: 8.5.2

When I run this command:

- Travis CI build/test.

I get the following output:
```
blt > setup:drupal:install:
     [echo] Installing Drupal.
You are about to DROP all tables in your 'drupal' database. Do you want to continue? (y/n): y
Starting Drupal installation. This takes a while. Consider using the [ok]
--notify global option.
exception 'Drupal\Core\Installer\Exception\AlreadyInstalledException'[error]
with message '<ul>
<li>To start over, you must empty your existing database and copy
<em>default.settings.php</em> over <em>settings.php</em>.</li>
<li>To upgrade an existing installation, proceed to the <a
href="/update.php">update script</a>.</li>
<li>View your <a href="http://default">existing site</a>.</li>
</ul>' in
/home/travis/build/acquia-pso/project/docroot/core/includes/install.core.inc:493
Stack trace:
#0
/home/travis/build/acquia-pso/project/docroot/core/includes/install.core.inc(112):
install_begin_request(Object(Composer\Autoload\ClassLoader), Array)
#1
/home/travis/build/acquia-pso/project/vendor/drush/drush/includes/drush.inc(726):
install_drupal(Object(Composer\Autoload\ClassLoader), Array)
#2
/home/travis/build/acquia-pso/project/vendor/drush/drush/includes/drush.inc(711):
drush_call_user_func_array('install_drupal', Array)
#3
/home/travis/build/acquia-pso/project/vendor/drush/drush/commands/core/drupal/site_install.inc(80):
drush_op('install_drupal', Object(Composer\Autoload\ClassLoader),
Array)
#4
/home/travis/build/acquia-pso/project/vendor/drush/drush/commands/core/site_install.drush.inc(247):
drush_core_site_install_version('lightning', Array)
#5 [internal function]: drush_core_site_install('lightning',
'install_configu...')
#6
/home/travis/build/acquia-pso/project/vendor/drush/drush/includes/command.inc(366):
call_user_func_array('drush_core_site...', Array)
#7
/home/travis/build/acquia-pso/project/vendor/drush/drush/includes/command.inc(217):
_drush_invoke_hooks(Array, Array)
#8 [internal function]: drush_command('lightning',
'install_configu...')
#9
[...]

PR #723, opened 2016-12-01
Issue #719: Fixed failing setup:git-hooks.

### LINKED ISSUE #719: BLT setup:git-hooks fails
My system information:
* BLT version: 8.6.0-beta1

When I run `blt setup:git-hooks`, it fails with this error:

> IOException: FileSystem::chmod() FAILED. Cannot chmod vendor/acquia/blt/scripts/git-hooks/commit-msg. Mode 755.

I suspect this has been failing for a while, and maybe we just didn't notice until this commit increased verbosity: https://github.com/acquia/blt/commit/71c4745e73ec18dabcbcc7c91fd42938b89ff7b5#diff-d6981a61601a737ce906a130556d199fR251

It seems like this chmod should no longer be necessary at all, since the git hooks are distributed by BLT and should already have the correct permissions.

Additionally, this helpful message should be updated to reference the correct location (not `scripts/git-hooks`): https://github.com/acquia/blt/commit/71c4745e73ec18dabcbcc7c91fd42938b89ff7b5#diff-d6981a61601a737ce906a130556d199fR238

danepowell:
Fixes #719 

I'm currently testing this locally.

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -234,26 +234,18 @@
     <if>
       <not><equals arg1="${git.hooks.pre-commit}" arg2="false"/></not>
       <then>
-        <echo level="info">Symlinking ${repo.root}/scripts/git-hooks/pre-commit to ${git.hooks.pre-commit}/pre-commit</echo>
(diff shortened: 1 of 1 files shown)


danepowell:
I've tested and verified this corrects the problem. However Travis is failing because of #720

PR #724, opened 2016-12-01
Issue #704: Fixed update hooks.

danepowell:
Fixes (partly) #704 

`project.yml` was moved to the `blt` subdirectory as part of 8.6.0-beta1, not 8.5.1, which is why I think this update hook wasn't running.

I'm not sure if this is the correct version syntax to apply, still testing locally.

### DIFF
--- a/src/Update/Updates.php
+++ b/src/Update/Updates.php
@@ -38,11 +38,11 @@ class Updates {
 
   /**
    * @Update(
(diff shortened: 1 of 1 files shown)


danepowell:
Now this is working as expected.

grasmash:
Thanks!

Issue #718, opened 2016-11-30
Module schemas are not installed when enabling a module via config when running blt deploy


ba66e77:
## Problem summary
I installed and configured the Redirect module on my local environment, exported config and ran `blt deploy` to push the code up.  The deploy command completed successfully, but the environment throws errors because the table which the module defines is not present in the database.  

The config for the module was imported though, and core.extension shows the module as enabled.  `Drush pmi` shows the module as not installed though.  Attempting run `drush en` says the module is enabled.  Running `drush pm-uninstall` says it is not enabled.

In testing, when I overrode the setup:update target to not run the config-imports and instead ran config-import manually when the deploy was complete, the table was installed and the module worked as expected.

## My system information:
* Operating system type: OSX
* Operating system version:10.11.6
* BLT version: 8.5.2

## Output of `blt doctor`:
```
blt > doctor:

Changed current directory to /home/vagrant/.composer
+----------------------+----------------------------------------------------------------------------------+
|[32m Property             [39m|[32m Value                                                                            [39m|
+----------------------+----------------------------------------------------------------------------------+
| %paths               |                                                                                  |
|  - %root             | /var/www/eebsmith/docroot                                                        |
|  - %site             | sites/default                                                                    |
|  - %modules          | modules                                                                          |
|  - %themes           | themes                                                                           |
|  - %config-sync      | /var/www/eebsmith/config/default                                                 |
|  - %files            | sites/default/files                                                              |
|  - %private          | /var/www/eebsmith/files-private                                                  |
|  - %temp             | /tmp                                                                             |
| admin-theme          | seven                                                                            |
| blt-version          | 8.5.2                                                                            |
| bootstrap            | Successful                                                                       |
| config-sync          | /var/www/eebsmith/config/default                                                 |
| db-driver            | mysql                                                                            |
| db-hostname          | localhost                                                                        |
| db-name              | drupal                                                                           |
| db-password          | drupal                                                                           |
| db-port              | 3306                                                                             |
| db-username          | drupal                                                                           |
| drupal-settings-file | sites/default/settings.php                                                       |
| drupal-version       | 8.2.3                                                                            |
| drush-alias-files    | /var/www/eebsmith/docroot/../drush/site-aliases/aliases.drushrc.php              |
|                      | /var/www/eebsmith/docroot/../drush/site-aliases/example.acsf.aliases.drushrc.php |
|                      | /var/www/eebsmith/docroot/../drush/site-aliases/example.local.aliases.drushrc.ph |
[...]

PR #728, opened 2016-12-02
Downgrading to Drush 8.

grasmash:


grasmash:
Need to wait for next Lightning release before this will pass.

dpagini:
j/w - why the downgrade? Is this related to an issue?

grasmash:
The latest Drush 9 release is actually behind the latest Drush 8 release. It lacks vital bug fixes. See https://github.com/acquia/blt/issues/718 for example.

PR #732, opened 2016-12-02
Change private files path.

dpagini:
This is a proposed change for the private files path in BLT. Acquia hosting puts private files in a `files-private` directory. BLT uses `private` directory (inside the `sites/*/` directory) for the same files. This PR just proposes to align the naming conventions. I know not everyone using BLT uses Acquia, but even within BLT you have `files-private` [mentions](https://github.com/acquia/blt/search?utf8=%E2%9C%93&q=files-private) throughout the code (namely [blt/default.local.settings.php](https://github.com/acquia/blt/blob/8.x/settings/default.local.settings.php#L126))

Changes proposed:
- Change `.gitignore` files
- Exclude folder on creation of artifact

Consideration: do we need some kind of upgrade task for this type of thing? I'm thinking maybe not b/c of the above mentioned default file.

### DIFF
--- a/phing/files/.gitignore
+++ b/phing/files/.gitignore
@@ -5,6 +5,7 @@ local.*
 # Ignore paths that contain user-generated content.
 docroot/sites/*/files
 docroot/sites/*/files-private
+docroot/sites/*/private
 
 # OS X
 .DS_Store
--- a/phing/files/.gitignore
+++ b/phing/files/.gitignore
@@ -4,7 +4,7 @@ local.*
 
 # Ignore paths that contain user-generated content.
 docroot/sites/*/files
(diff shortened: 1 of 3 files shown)


phing/files/.gitignore
```diff
@@ -4,7 +4,7 @@ local.*
 
 # Ignore paths that contain user-generated content.
 docroot/sites/*/files
-docroot/sites/*/private
+docroot/sites/*/files-private
```
grasmash: Why don't we leave both of these lines in place to cover our all of our bases?

dpagini:
Works for me. Fixed as requested, @grasmash.

PR #733, opened 2016-12-02
Fixed errors on ACE deploys.

danepowell:
After updating to 8.6.0-beta1, deployments to ACE throw this error:

> sh: 1: 
> composer: not found

I'm assuming this is due to the addition of this version check, which uses composer: https://github.com/acquia/blt/blob/8.x/phing/tasks/properties.xml#L36

This may not be the most elegant solution, but it eliminates the errors. Another (more performant) solution might be to make the version check a dedicated function that runs only as a dependency of Phing targets that need it (i.e. `update-delta` and `version`)

Still testing this.


grasmash:
Looks like logoutput is set to "false " rather than "false".

danepowell:
Ah good eye! I was wondering why output was still being logged... Should be good to go now.

PR #743, opened 2016-12-05
Fixes #737: Twig validation does not recognize Drupal filters/functions.

### LINKED ISSUE #737: Twig Linting doesn't recognize Drupal functions/filters/etc
My system information:
* Operating system type: OSX
* Operating system version: 10.12.1
* BLT version: 8.6.0-beta1

Output of `blt doctor`:
```
+--------------------------+---------------------------------------------------------------------------------+
| Check                    | Outcome                                                                         |
+--------------------------+---------------------------------------------------------------------------------+
| checkDocrootExists       | Found docroot.                                                                  |
| checkCoreExists          | Drupal core exists                                                              |
| checkSettingsFile        | BLT settings are included in settings file.                                     |
| checkLocalSettingsFile   | Found your local settings file.                                                 |
| checkLocalDrushFile      | Found your local drush settings file.                                           |
| checkUri                 | $options['uri'] is set.                                                         |
| checkUriResponse         | Received a response from site local.mysite.com.                       |
| checkFileSystem:%files   | Public files directory is writable.                                             |
| checkFileSystem:%private | Private files directory is not set.                                             |
| checkFileSystem:%temp    | Temporary files directory is writable.                                          |
| checkDbConnection        | Connected to database.                                                          |
| checkDrupalBootstrapped  | Bootstrapped Drupal via drush.                                                  |
| checkDrupalInstalled     | Drupal is installed.                                                            |
| checkCaching:page        | Drupal cache is disabled.                                                       |
| checkCaching:css         | CSS preprocessing is disabled.                                                  |
| checkCaching:js          | JS preprocessing is disabled.                                                   |
| checkNvmExists           | NVM does not exist.                                                             |
|                          |                                                                                 |
|                          | It is recommended that you use NVM to manage multiple versions of NodeJS on one |
|                          | machine.                                                                        |
|                          | Instructions for installing NVM can be found at:                                |
|                          |   https://github.com/creationix/nvm#installation                                |
| checkCiConfig            | Git remotes are set in project.yml.                                             |
| checkComposer:require    | acquia/blt is in composer.json's require object.                                |
| checkComposer:plugins    | hirak/prestissimo plugin for composer is installed.                             |
| checkBehat:exists        | Behat local settings file exists.                                               |
| checkProjectYml:keys     | project.yml has no deprecated keys.                                             |
| checkPhpDateTimezone     | PHP setting for date.timezone is correctly set                                  |
+--------------------------+---------------------------------------------------------------------------------+
```

When I run this command:
```
git commit -m "Some commit message"
```

I get the following output:
```
      11      */
      12     #}
  >>  13     <img{{ attributes|without('height','width') }} />
  >> Unknown "without" filter. 
      14     
  ERROR  in /Users/dpagini/Websites/mysite/docroot/themes/custom/mytheme/templates/layout/html.html.twig (line 27)
```

This is a valid Drupal twig filters, so they should be recognized as such.

Apologies if this was already reported, or even fixed, but I couldn't find another issue that reported the same thing. The twig linting does not recognize drupal twig function such as `path()` or filters such as `without`. My commit was still successful, so I think this is just for information for now, but for this to be useful, I dont want to see these types of errors, or else the output here will just be ignored.

grasmash:


grasmash:
Unfortunately, we've got to downgrade this feature to a degree. We're switching to pure syntax checking and ignoring things like filter/function checks. 

The previous approach required a real compilation of twig into php, which would have required a full Drupal bootstrap and real implementation of Drupal core's twig extension. That was just a bit too much of a rabbit hole for me to chase. 

However, this could be a good feature for Drupal Console, which is container aware. @jmolivas you might be interested in adding twig validation.

src/Console/Command/TwigLintCommand.php
```diff
@@ -1,28 +0,0 @@
...
-
-  protected function configure() {
-    $this->addArgument('environment-dir', InputOption::VALUE_REQUIRED);
-    parent::configure();
-  }
-
-  protected function execute(InputInterface $input, OutputInterface $output)
-  {
-    $loader = new \Twig_Loader_Filesystem($input->getArgument('environment-dir'));
-    $this->setTwigEnvironment(new \Twig_Environment($loader));
```
grasmash: If you wanted to go all the way here, you'd actually add Drupal core's Twig extension:
```
    $twig = new \Twig_Environment(NULL, array(
      'debug' => TRUE,
      'cache' => FALSE,
      'autoescape' => 'html',
      'optimizations' => 0
    ));
$twig->addExtension(new TwigExtension($renderer);
```

PR #761, opened 2016-12-06
Generating a subtheme in project set-up. Adding default front-end tasks.

aellison:
Fixes # .

Changes proposed:
-
-

### DIFF
--- a/phing/tasks/frontend.xml
+++ b/phing/tasks/frontend.xml
@@ -2,18 +2,37 @@
 
   <target name="frontend" description="Runs all frontend targets"
           depends="frontend:setup, frontend:build">
+    <echo>Running main frontend task.</echo>
   </target>
 
-  <target name="frontend:setup" description="Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build." hidden="true">
-    <phingcall target="target-hook:invoke">
-      <property name="hook-name" value="frontend-setup"/>
-    </phingcall>
+  <target name="frontend:setup" description="Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build.">
+    <if>
+      <available file='${target-hooks.frontend-setup.dir}' type='dir' />
+      <then>
+        <phingcall target="target-hook:invoke">
+          <property name="hook-name" value="frontend-setup"/>
+        </phingcall>
+        <echo>Front-end setup task run.</echo>
+      </then>
+      <else>
+        <echo>Subtheme not yet created.</echo>
+      </else>
+    </if>
   </target>
 
-  <target name="frontend:build" description="Uses project.yml hooks to run custom defined commands to build front end dependencies for custom themes." hidden="true">
-    <phingcall target="target-hook:invoke">
-      <property name="hook-name" value="frontend-build"/>
-    </phingcall>
+  <target name="frontend:build" description="Uses project.yml hooks to run custom defined commands to build front end dependencies for custom themes.">
+    <if>
+      <available file='${target-hooks.frontend-setup.dir}' type='dir' />
+      <then>
+        <phingcall target="target-hook:invoke">
+          <property name="hook-name" value="frontend-build"/>
+        </phingcall>
+        <echo>Front-end build task run.</echo>
+      </then>
+      <else>
+        <echo>Subtheme not yet created.</echo>
+      </else>
+    </if>
   </target>
 
 </project>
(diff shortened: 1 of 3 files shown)


phing/tasks/frontend.xml
```diff
@@ -2,18 +2,37 @@
...
+    <echo>Running main frontend task.</echo>
   </target>
 
-  <target name="frontend:setup" description="Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build." hidden="true">
-    <phingcall target="target-hook:invoke">
-      <property name="hook-name" value="frontend-setup"/>
-    </phingcall>
+  <target name="frontend:setup" description="Uses project.yml hooks to run custom defined commands to setup front end dependencies for frontend:build.">
+    <if>
+      <available file='${target-hooks.frontend-setup.dir}' type='dir' />
```
grasmash: We have different logic that does this same check but explicitly fails when the directory does not exist:
https://github.com/acquia/blt/blob/8.x/phing/build.xml#L99

Can we just avoid calling commands for non-existent dirs?

phing/tasks/frontend.xml
```diff
@@ -2,18 +2,37 @@
...
-      <property name="hook-name" value="frontend-build"/>
-    </phingcall>
+  <target name="frontend:build" description="Uses project.yml hooks to run custom defined commands to build front end dependencies for custom themes.">
+    <if>
+      <available file='${target-hooks.frontend-setup.dir}' type='dir' />
+      <then>
+        <phingcall target="target-hook:invoke">
+          <property name="hook-name" value="frontend-build"/>
+        </phingcall>
+        <echo>Front-end build task run.</echo>
```
grasmash: This echo statement should be moved to target-hook:invoke and simply output the target-hook name dynamically.

phing/tasks/setup.xml
```diff
@@ -289,4 +289,17 @@
     </if>
   </target>
 
+  <!-- Create a subtheme and set it as the site default -->
+  <target name="setup:subtheme" description="Creates a new subtheme of Cog.">
+    <drush command="en cog -y" verbose="false" logoutput="true" />
+    <drush command="cog ${project.subtheme.machine_name} ${project.subtheme.human_name} --description=${project.subtheme.description}" verbose="false" dir="${docroot}/themes/contrib/cog" logoutput="true" />
```
grasmash: It looks like this will run every single time that "setup" is executed. We should probably make this generation a bit more intentional and couple it less tightly to BLT's core targets.

For example, create a `frontend:cog:init` command that will:
* Ensure cog is in composer.json
* Enable cog
* Generate a subtheme
* Enable subtheme
* Set config

This way it can be run once and then not again.

template/blt/project.yml
```diff
@@ -12,6 +12,11 @@ project:
   local:
     protocol: http
     hostname: local.${project.machine_name}.com
+  # These values will be used to generate the theme for this project.
+  subtheme:
```
grasmash: Unless there's a compelling reason to have a human name and a description, I think using simply `subtheme: cogsub` should suffice. However, I might default it to something dynamic like `${project.machine_name}_sub`.

To be honest , project.human_name is fairly useless right now.

template/blt/project.yml
```diff
@@ -39,15 +44,15 @@ target-hooks:
...
-    command: echo 'No frontend-setup configured.'
+    command: 'npm install -g yarn; yarn install;'
   # Executed when front end assets should be generated.
   frontend-build:
     # E.g., ${docroot}/sites/all/themes/custom/mytheme.
-    dir: ${docroot}
+    dir: '${docroot}/themes/custom/${project.subtheme.machine_name}'
     # E.g., `npm run build`.
-    command: echo 'No frontend-build configured.'
+    command: 'gulp'
```
grasmash: This should be executing using an npm script rather than running as a globally available bin.

template/blt/project.yml
```diff
@@ -39,15 +44,15 @@ target-hooks:
...
-    dir: ${docroot}
+    dir: '${docroot}/themes/custom/${project.subtheme.machine_name}'
     # E.g., `npm install` or `bower install`.
-    command: echo 'No frontend-setup configured.'
+    command: 'npm install -g yarn; yarn install;'
   # Executed when front end assets should be generated.
   frontend-build:
     # E.g., ${docroot}/sites/all/themes/custom/mytheme.
-    dir: ${docroot}
+    dir: '${docroot}/themes/custom/${project.subtheme.machine_name}'
```
grasmash: The `frontend:cog:init` theme could just write this value using `drupal yaml:update` so that it doesn't exist prior to the directory existing.

grasmash:
This must now be refactored due to port to Robo.

PR #762, opened 2016-12-06
Updating template/composer.json to use Drupal packagist constraints.

grasmash:


grasmash:
Todo:
- [ ]  Integrate script in update hook
- [ ]  Resolve drupal security advisory conflict (upstream in Lightning)

scripts/blt/composer-version-convert.php
```diff
@@ -0,0 +1,108 @@
...
+    return '*@stable';
+  }
+
+  // ~8.1.0-alpha1 > ~1.0.0-alpha1
+  // ^8.1.0-alpha1 > ^1.0.0-alpha1
+  // ^8.1.0 > ^1.0
+  // 8.1.0-alpha1 > 1.0.0-alpha1
+  // 8.1.0-beta12 > 1.0.0-beta12
+  // 8.12.0-rc22 > 12.0.0-rc22
+  if (preg_match('/^([\^~])?8(\.)?(\d+)?(\.\d+)?(-(alpha|beta|rc)\d+)?(\.\*)?(@(dev|alpha|beta|rc))?/', $version_constraint, $matches)) {
```
grasmash: Need to also cover refs and "dev-" prefixes.

grasmash: Need to support more operators:
https://getcomposer.org/doc/articles/versions.md#range

scripts/blt/composer-version-convert.php
```diff
@@ -0,0 +1,112 @@
...
+  'url' => 'https://packages.drupal.org/8',
+];
+
+$output = json_encode($composer_json, JSON_PRETTY_PRINT|JSON_UNESCAPED_SLASHES);
+print $output;
+
+function convert_version_constraint($version_constraint) {
+
+  // 8.* => *
+  if (preg_match('/8\.\*$/', $version_constraint)) {
```
phenaproxima: There's no anchor at the beginning, so won't this also match something like 1.8.*?

balsama:
We ended up tweaking the output for branches a little bit to:
* Account for the `-dev` coming at the end, e.g: `"drupal/google_analytics": "8.2.x-dev"`
* Ultimately convert all branches to `{#}.x-dev` as that seems to be the only documented format for branches on D.O packagist

See here: https://github.com/acquia/lightning/blob/8.x-1.x/src/UpgradeBuild.php

PR #771, opened 2016-12-07
Removing testProd PHPUnit test.

grasmash:


grasmash:
A change introduced by Drupal 8.2.4 has broken this test. The assertion in this test would better belong in an insight hook anyway. Removing.

PR #780, opened 2016-12-09
issue-777 Make php memory limit settable

nikgregory:
Fixes # 777

Changes proposed:
- adds a php.mem_limit property

Usage:
`blt foo -Dphp.mem_limit=1G`

### DIFF
--- a/phing/tasks/properties.xml
+++ b/phing/tasks/properties.xml
@@ -1,5 +1,14 @@
 <project name="properties" default="build">
 
+  <!-- Set the memory limit for php. -->
+  <if>
+    <isset property="php.mem_limit"/>
+    <then>
+      <echo>Setting memory allocation to ${php.mem_limit}</echo>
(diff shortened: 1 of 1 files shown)


phing/tasks/properties.xml
```diff
@@ -1,5 +1,14 @@
 <project name="properties" default="build">
 
+  <!-- Set the memory limit for php. -->
+  <if>
+    <isset property="php.mem_limit"/>
```
grasmash: Small request, but I'd prefer to have this variable named memory_limit so that it matches the setting in php.ini.

Also, we already have a line in build.xml that modifies the memory_limit:
https://github.com/acquia/blt/blob/8.x/phing/build.xml#L21

We should modify the existing line rather than adding this to a second place in the code base. This should be simple to accomplish:

```
<property name="php.memory_limit" value="1G" override="false" />
<php expression="ini_set('memory_limit', '${php.memory_limit}');" verbose="true" />
```

PR #782, opened 2016-12-09
Issue 779 Separate out composer from drush in acsf:init

nikgregory:
Fixes # 779

Changes proposed:
- create a new acsf:init:drush task that does the drush work

This should allow builds to be more efficient since I think that composer update will get new versions of acsf and there should be no need to run that twice if the project is already 'inited'


grasmash:
This has a merge conflict as a result of another PR being merged. Please rebase.

grasmash:
Wow nevermind, apparently you can now resolve conflicts in the GitHub UI.

PR #797, opened 2016-12-14
Fixes #796: Don't exclude local.* files from build artifact.

### LINKED ISSUE #796: Don't exclude local.* files from build artifact
I spent a good deal of time today trying to figure out why some configuration on my `deploy` environments differed from the configuration in my `local` environment—there was some `config_split` module configuration for the local environment (which needed to be on all environments) that disappeared from dev/stage/prod, causing local build failures.

I was also speaking to another developer, @msherron, who had a similar issue with some local task (e.g. the tabs on a node page) configuration that went missing from build artifacts.

The `local.*` gitignore and `deploy-exclude.txt` default configuration is a bit heavy-handed, as it will block out configuration/feature exports including any file with names like something.local.yml.

geerlingguy:
Fixes #796 .

Changes proposed:
- Don't exclude `local.*` from default gitignore template for build artifacts.
- Don't exclude `local.*` files from default `deploy-exclude.txt` file.

### DIFF
--- a/phing/files/deploy-exclude.txt
+++ b/phing/files/deploy-exclude.txt
@@ -25,7 +25,7 @@
 /.gitignore
 .git
 example.*
-local.*
-*.local.*
+local.settings.php
+local.drushrc.php
 node_modules
 /vendor
--- a/template/.gitignore
+++ b/template/.gitignore
@@ -1,5 +1,6 @@
 # Ignore configuration files that may contain sensitive information.
-local.*
+local.settings.php
+local.drushrc.php
 *.local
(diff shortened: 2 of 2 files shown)


grasmash:
We still need to exclude local.settings.php

geerlingguy:
@grasmash ah, good catch. I'll add that as a separate line then, if that's okay.

geerlingguy:
PR has been updated. I tested with one of my codebases, and it seems to exclude the settings file correctly (but not the default/example one).

PR #814, opened 2016-12-19
Fixed memcache contamination.

danepowell:
We noticed odd behavior on our dev/stage sites (which share the same hardware) where cache keys were very clearly being shared between sites. For instance, we deployed a module to dev, and then got routing errors on stage indicating that Drupal was looking for that module.

My theory is that the memcache module is not properly prefixing keys per host. It appears that the ACE server-side includes do configure these keys correctly, and the memcache store uses them correctly about 99% of the time. But if the key is too long, the memcache module hashes it and I think this is the case where the hostname is ignored and caches can become contaminated across multiple sites on the same server.

This is fixed upstream here, so updating to dev should fix it: https://www.drupal.org/node/2701903

I also requested a new alpha release, so we can get back off of dev ASAP: https://www.drupal.org/node/2837857

### DIFF
--- a/template/composer.json
+++ b/template/composer.json
@@ -15,7 +15,7 @@
         "drupal/cog": "^1.0.0",
         "drupal/core": "^8.0",
         "drupal/features": "^3.0.0",
-        "drupal/memcache": "^2.0.0-alpha2",
+        "drupal/memcache": "2.x-dev",
         "drupal/seckit": "^1.0.0-alpha2",
         "drupal/security_review": "*",
         "drupal/shield": "^1.0.0",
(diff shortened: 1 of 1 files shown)


template/composer.json
```diff
@@ -15,7 +15,7 @@
         "drupal/cog": "^1.0.0",
         "drupal/core": "^8.0",
         "drupal/features": "^3.0.0",
-        "drupal/memcache": "^2.0.0-alpha2",
+        "drupal/memcache": "2.x-dev",
```
grasmash: Please specify a hash since this has a patch applied against it.

danepowell: Hmm I don't think there's a patch against it but I can still lock it to a hash tomorrow.

grasmash: Oh I see. I noticed that a patch had changed in this PR and assumed it was for memcache. Looks like it's actually just a whitespace change.

PR #815, opened 2016-12-19
Added Drush to DrupalVM.

danepowell:
DrupalVM 4.* no longer includes Drush by default, it needs to be specifically included.

Without this, most BLT tasks fail.

cc @geerlingguy


geerlingguy:
@danepowell / @grasmash - One new feature in Drupal VM 4.0 was the ability to not have drush installed globally, but to use the packaged drush that you ship with your project. Just FYI—I still like drush installed globally, but you can do it either way with this version.

PR #827, opened 2016-12-21
Fixes #799: additional support for multisites

### LINKED ISSUE #799: Remote config import throws exception
My system information:
* Operating system version: 12.04.5
* BLT version: 8.x-dev

Output of `blt doctor`:
```
blt > doctor:

sh: 1: composer: not found
+----------------------+----------------------------------------------------------------------------------+
| Property             | Value                                                                            |
+----------------------+----------------------------------------------------------------------------------+
| %paths               |                                                                                  |
|  - %root             | /mnt/www/html/[redacted]dev/docroot                                                |
|  - %site             | sites/default                                                                    |
|  - %modules          | modules                                                                          |
|  - %themes           | themes                                                                           |
|  - %config-sync      | /mnt/www/html/[redacted]dev/docroot/sites/default/files/config_4bf1f02d5081baa3a95 |
|                      | acaf7465bd63b0efb82ed/sync                                                       |
|  - %config-vcs       | /mnt/www/html/[redacted]dev/docroot/../config/default                              |
|  - %files            | sites/default/files                                                              |
|  - %private          | /mnt/files/[redacted].dev/files-private                                            |
|  - %temp             | /mnt/tmp/[redacted]dev                                                             |
| admin-theme          | seven                                                                            |
| blt-version          | 8.x-dev                                                                          |
| bootstrap            | Successful                                                                       |
| config-sync          | /mnt/www/html/[redacted]dev/docroot/sites/default/files/config_4bf1f02d5081baa3a95 |
|                      | acaf7465bd63b0efb82ed/sync                                                       |
| config-vcs           | /mnt/www/html/[redacted]dev/docroot/../config/default                              |
| db-driver            | mysql                                                                            |
| db-hostname          | [redacted]                                                                    |
| db-name              | [redacted]dev                                                                      |
| db-password          | [redacted]                                                                  |
| db-port              | 3306                                                                             |
| db-status            | Connected                                                                        |
| db-username          | [redacted]                                                                           |
| drupal-settings-file | sites/default/settings.php                                                       |
| drupal-version       | 8.2.4                                                                            |
| drush-alias-files    | /mnt/drush/[redacted]//[redacted].aliases.drushrc.php                                |
|                      | /mnt/www/html/[redacted]dev/docroot/../drush/site-aliases/aliases.drushrc.php      |
| drush-conf           | /mnt/www/html/[redacted]dev/drush/drushrc.php                                      |
|                      | /etc/drush/drushrc.php                                                           |
| drush-script         | /mnt/www/html/[redacted]dev/vendor/drush/drush/drush.php                           |
| drush-temp           | /mnt/tmp/[redacted]dev                                                             |
| drush-version        | 8.1.8                                                                            |
| files                | sites/default/files                                                              |
| install-profile      | standard                                                                         |
| modules              | modules                                                                          |
| php-bin              | /usr/local/php5.6/bin/php                                                        |
| php-conf             | /usr/local/php5.6/etc/cli/php.ini                                                |
| php-os               | Linux                                                                            |
| private              | /mnt/files/[redacted].dev/files-private                                            |
| root                 | /mnt/www/html/[redacted]dev/docroot                                                |
| site                 | sites/default                                                                    |
| temp                 | /mnt/tmp/[redacted]dev                                                             |
| theme                | meyers                                                                           |
| themes               | themes                                                                           |
| uri                  | http://default                                                                   |
| user                 |                                                                                  |
+----------------------+----------------------------------------------------------------------------------+
+--------------------------+---------------------------------------------------------------------------------+
| Check                    | Outcome                                                                         |
+--------------------------+---------------------------------------------------------------------------------+
| checkDocrootExists       | Found docroot.                                                                  |
| checkCoreExists          | Drupal core exists                                                              |
| checkSettingsFile        | BLT settings are included in settings file.                                     |
| checkLocalSettingsFile   | Could not find local settings file.                                             |
|                          | Your local settings file should exist at                                        |
|                          | /mnt/www/html/[redacted]dev/docroot/sites/default/settings/local.settings.php.    |
| checkLocalDrushFile      | Local drushrc file does not exist.                                              |
|                          | Create /mnt/www/html/[redacted]dev/docroot/sites/default/local.drushrc.php.       |
|                          | Run `blt setup:drush:settings` to generate it automatically, or run `blt setup` |
|                          | to run the entire setup process.                                                |
| checkFileSystem:%files   | Public files directory is writable.                                             |
| checkFileSystem:%private | Private files directory is writable.                                            |
| checkFileSystem:%temp    | Temporary files directory is writable.                                          |
| checkDbConnection        | Connected to database.                                                          |
| checkDrupalBootstrapped  | Bootstrapped Drupal via drush.                                                  |
| checkDrupalInstalled     | Drupal is installed.                                                            |
| checkCaching:page        | Drupal cache is disabled.                                                       |
| checkCaching:css         | CSS preprocessing is disabled.                                                  |
| checkCaching:js          | JS preprocessing is disabled.                                                   |
| checkNvmExists           | NVM does not exist.                                                             |
|                          |                                                                                 |
|                          | It is recommended that you use NVM to manage multiple versions of NodeJS on one |
|                          | machine.                                                                        |
|                          | Instructions for installing NVM can be found at:                                |
|                          |   https://github.com/creationix/nvm#installation                                |
| checkComposer:require    | acquia/blt is in composer.json's require object.                                |
| checkComposer:plugins    | hirak/prestissimo plugin for composer is not installed.                         |
|                          |   Run `composer global require hirak/prestissimo:^0.3` to install it.           |
|                          |   This will improve composer install/update performance by parallelizing the    |
|                          | download of dependency information.                                             |
| checkBehat:exists        | tests/behat/local.yml is missing!                                               |
|                          |   Run `blt setup:behat` to generate it from example.local.yml.                  |
| checkProjectYml:keys     | project.yml has no deprecated keys.                                             |
| checkAcsfConfig          | BLT settings are included in your pre-settings-php include.                     |
| checkPhpDateTimezone     | PHP setting for date.timezone is correctly set                                  |
+--------------------------+---------------------------------------------------------------------------------+

BUILD FINISHED

Total time: 3.2329 seconds

```

When I run this command (on remote environment):
```
blt setup:config-import 
```

I get the following output:
```
blt > deploy:update:


blt > setup:config-import:

     [echo] Importing configuration...

Fatal error: Uncaught exception 'Symfony\Component\HttpKernel\Exception\BadRequestHttpException' in /mnt/www/html/[redacted]/docroot/core/lib/Drupal/Core/DrupalKernel.php:361
Stack trace:
#0 /mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/DrupalBoot8.php(68): Drupal\Core\DrupalKernel::findSitePath(Object(Symfony\Component\HttpFoundation\Request), true)
#1 /mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/DrupalBoot.php(391): Drush\Boot\DrupalBoot8->conf_path(true, true)
#2 /mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/DrupalBoot.php(333): Drush\Boot\DrupalBoot->bootstrap_drupal_site_validate_settings_present()
#3 /mnt/www/html/[redacted]/vendor/drush/drush/includes/bootstrap.inc(430): Drush\Boot\DrupalBoot->bootstrap_drupal_site_validate()
#4 /mnt/www/html/[redacted]/vendor/drush/drush/includes/bootstrap.inc(471): drush_bootstrap_validate(2)
#5 /mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/BaseBoot.php(59): drush_bootstrap_to_phase(6)
#6 /mnt/www/html/[redacted]/vendor/drush/drush/includes/pr in /mnt/www/html/[redacted]/docroot/core/lib/Drupal/Core/DrupalKernel.php on line 361
Drush command terminated abnormally due to an unrecoverable error.       [error]
Error: Uncaught exception
'Symfony\Component\HttpKernel\Exception\BadRequestHttpException' in
/mnt/www/html/[redacted]/docroot/core/lib/Drupal/Core/DrupalKernel.php:361
Stack trace:
#0
/mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/DrupalBoot8.php(68):
Drupal\Core\DrupalKernel::findSitePath(Object(Symfony\Component\HttpFoundation\Request),
true)
#1
/mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/DrupalBoot.php(391):
Drush\Boot\DrupalBoot8->conf_path(true, true)
#2
/mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/DrupalBoot.php(333):
Drush\Boot\DrupalBoot->bootstrap_drupal_site_validate_settings_present()
#3
/mnt/www/html/[redacted]/vendor/drush/drush/includes/bootstrap.inc(430):
Drush\Boot\DrupalBoot->bootstrap_drupal_site_validate()
#4
/mnt/www/html/[redacted]/vendor/drush/drush/includes/bootstrap.inc(471):
drush_bootstrap_validate(2)
#5
/mnt/www/html/[redacted]/vendor/drush/drush/lib/Drush/Boot/BaseBoot.php(59):
drush_bootstrap_to_phase(6)
#6 /mnt/www/html/[redacted]/vendor/drush/drush/includes/pr in
/mnt/www/html/[redacted]/docroot/core/lib/Drupal/Core/DrupalKernel.php,
line 361
[phingcall] /var/www/html/[redacted].prod/vendor/acquia/blt/phing/tasks/setup.xml:198:32: Drush exited with code 255

BUILD FAILED
/var/www/html/[redacted].prod/vendor/acquia/blt/phing/tasks/deploy.xml:163:22: Execution of the target buildfile failed. Aborting.

Total time: 0.4797 seconds

Update errored.
```

And I expected this to happen: No errors to occur.

bobbygryzynger:
Fixes #799.

Changes proposed:
- Allow multisite-specific configuration to be supplied.
- Perform update and deployment tasks for all defined sites.


phing/build.yml
```diff
@@ -72,11 +72,18 @@ git:
...
     commit-msg: ${blt.root}/scripts/git-hooks
 
+# Site(s) definition.
 multisite:
-  # The docroot/sites/default directory is used by default.
-  name:
-    - default
-    # - another_site
+  default:
+    aliases:
```
grasmash: This seems to conflict with the existing drush.aliases configuration. It's not clear to me how these interact. Why do we have both drush.aliases and multisite.default.aliases?

phing/build.yml
```diff
@@ -72,11 +72,18 @@ git:
...
+      local: self
+      remote: ${project.machine_name}.test
+    uri:
+      local: ${project.machine_name}.local
```
grasmash: This is a lot of config. Why do we need per site AND per env uris? I'd prefer to start with per site uris.

phing/phingcludes/YamlKeysTask.php
```diff
@@ -0,0 +1,167 @@
+<?php
+
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * Class YamlKeysTask
+ *
+ * This task allows keys to be extracted from YAML variables and output to Phing
```
grasmash: Can you add some example usage information?

phing/tasks/acsf.xml
```diff
@@ -10,10 +10,18 @@
     <phingcall target="acsf:init:drush"/>
   </target>
 
+  <!-- Perform acsf:init for each multisite. -->
+  <target name="acsf:init:drush">
+    <!-- Get the mulitsite names. -->
+    <yamlKeys file="${blt.config-files.project}" variable="multisite" outputProperty="multisite.names"/>
+    <foreach list="${multisite.names}" param="multisite.name" target="acsf:init:site:drush"/>
```
grasmash: I didn't realize that acsf had to be initialized per multisite. @nikgregory any concerns here?

phing/tasks/local-sync.xml
```diff
@@ -15,15 +15,21 @@
     </phingcall>
   </target>
 
-  <target name="local:sync" description="Synchronize local environment from remote (remote --> local)."
+  <!-- Sync each local multisite from remote aliases. -->
+  <target name="local:sync" description="Synchronize each local environment from remote (remote --> local).">
+    <yamlKeys file="${blt.config-files.project}" variable="multisite" outputProperty="multisite.names"/>
+    <foreach list="${multisite.names}" param="multisite.name" target="local:sync:site"/>
```
grasmash: Rather than apply this to every multisite every time that this target is run, I'd prefer to run one command per multisite. Why not just allow the user the pass the multisite name as an argument?

Better yet, we can avoid a lot of the configuration nesting if we use multiple config files, one per multi site. Each multisite can have a unique file that defines drush.aliases, project.uri, etc. You can pass a reference to the config file using `-propertyfile <file> -propertyfileoverride`. If that's too much to type, we can wrap *that* with something higher level.

bobbygryzynger: Agreed, this is a bit heavy as I'm seeing it run with my own multisites and often wouldn't be necessary. A target like `local:sync:all` could work as a nice wrapper if you do indeed want to grab everything. 

I like the idea of separating out site config to a file for each site.

phing/build.yml
bobbygryzynger: The reason I have both here right now is for the `local:sync` target. I'm not aware of a way to do a `sql-sync` only specifying `uri`s, but perhaps there's a way I don't know of?

phing/build.yml
bobbygryzynger: I suppose the `drush.aliases` could go away entirely and be replaced with a `sites.default.yml` (or `sites/default.yml`) config file that would be in the template by default and would be loaded if no other config is specified; if this is something you're not adversed to.

blt-project/blt/sites/default.yml
```diff
@@ -0,0 +1,12 @@
+# @file default.yml
+# Site definition file for the default site.
+aliases:
```
grasmash: I'd prefer for the default values to be taken from project.yml. Users can define a separate default.yml if they wish, but for backwards compatibility, let's support the current approach.

blt-project/blt/sites/example.yml
```diff
@@ -0,0 +1,21 @@
+# @file
+# example.yml
+# Site definition file for remote and local environment information.
+#
+# The filename of this file should correspond to docroot/sites/[site-name]
+# (e.g. default.yml for docroot/sites/default). You should create one file in
+# this directory for each site running against your codebase.
+aliases:
```
grasmash: Can we continue to use drush.aliases here rather than introducing something new?

bobbygryzynger: I've made this change and moved the configuration this file supplies to `phing/example.multisite.yml` so it doesn't have to be explicitly ignored during initialization.

blt-project/blt/sites/default.yml
bobbygryzynger: My latest commits have made all configuration under `blt/sites/*` optional, with the values set in `project.yml` being used if no site-specific configuration is supplied.

phing/tasks/acsf.xml
```diff
@@ -7,12 +7,33 @@
...
+    <phingcall target="acsf:init:all" />
+  </target>
+
+  <!-- Perform acsf:init:site for each multisite. -->
+  <target name="acsf:init:all" description="Refreshes the acsf settings and hook files for all acsf sites.">
+    <foreach list="${multisite.names}" param="site.name" target="acsf:init:site"/>
+  </target>
+
+  <!-- Wrapper around acsf:init:site:drush. -->
+  <target name="acsf:init:site" description="Refreshes the acsf settings and hook files for an acsf site.">
```
bobbygryzynger: I'm not certain that this is necessary as I'm not entirely familiar with the requirements of ACSF. This can be reverted if initialization doesn't need to be done on a per-site basis.

bobbygryzynger:
@grasmash - are there any other changes you'd like to see here? 

I'm still unsure about the requirements for initializing ACSF, but perhaps there is someone you could ping regarding what is necessary (per-site or per-environment initialization)?

phing/tasks/vm.xml
```diff
@@ -27,7 +27,7 @@
   </target>
 
 
-  <target name="vm:local:init" depends="setup:drupal:settings" description="Initializes DrupalVM for this machine." hidden="true">
+  <target name="vm:local:init" depends="setup:drupal:settings:all" description="Initializes DrupalVM for this machine." hidden="true">
```
grasmash: This doesn't appear to be a real target any longer?

phing/build.yml
```diff
@@ -31,6 +31,8 @@ blt:
     local: ${repo.root}/blt/project.local.yml
     example-local: ${repo.root}/blt/example.project.local.yml
     schema-version: ${repo.root}/blt/.schema_version
+  config-dirs:
```
grasmash: Is there an instance when this would not be sites?

phing/tasks/setup.xml
```diff
@@ -57,44 +57,60 @@
...
   <!-- We intentionally do not run setup:behat because we'd like to delay setting values in local.yml until just before they are run. -->
-  <target name="setup:settings" depends="setup:drupal:settings, setup:drush:settings, setup:console:settings" description="Creates settings files for Drupal and drush.">
+  <target name="setup:settings" depends="setup:drupal:settings:all, setup:drush:settings:all" description="Creates settings files for Drupal and drush.">
   </target>
 
   <target name="setup:console:settings" description="Initializes Drupal console config.">
     <exec dir="${repo.root}" command="${composer.bin}/drupal init --quiet" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
   </target>
 
   <!-- Run Drupal settings setup for each multisite. -->
```
bobbygryzynger: `setup:drupal:settings:all`, which `vm:local:init` depends on is here.

phing/tasks/vm.xml
bobbygryzynger: See my comment below, this task is in `setup.xml`.

phing/build.yml
bobbygryzynger: At this point, no, but I figured to future-proof this a bit it may make sense to allow additional keys to be added here in the future. But the `sites` key could certainly simply be placed under the existing `config-files` key.

phing/build.yml
```diff
@@ -93,3 +89,5 @@ phpcs:
 project:
   local:
     uri: ${project.local.protocol}://${project.local.hostname}
+  ci:
+    uri: 127.0.0.1:8888
```
grasmash: First of all, thanks for all of your hard work on this. We'll definitely get your work merged in.

But first, I need to get my head around what's happening, so I'll take some notes here for my own usage. On a global level we have:

* project.local.uri
* project.ci.uri
* multisite.names
* site.name
* environment

Then for the active multisite, we generate import variables from the site-specific yaml file into the "site" namespace:

* site.behat...
* site.uri.local
* site.uri.ci
* site.drush.aliases.remote

Couple of things:

1. Why do we use project.ci.uri in one place and site.uri.ci? It seems like we should have a consistent naming scheme.
1. Why import variables into the site namespace at all? It seems simpler to just have the multisite yml file override the global variables, so that, for instance, it changes the values for:
  * project.local.uri
  * project.ci.uri
  * behat.tags
  * drush.aliases.local
  * etc.

I think I may cherry pick some of your code and introduce it in smaller PRs so that we can release it in stages. I'm concerned about too much change too quickly here. I'll get back to you.

bobbygryzynger: @grasmash, no problem and your inclination towards breaking this up a bit sounds good to me. Let me know if there is any way that I can assist.

From my perspective, it would be most helpful to have the multisite support during the deployment and `local:sync`/`local:refresh` stages. The lack of strong support here forces me to run many things manually on each of the sites my codebase supports.

Regarding your two comments:
- I agree the `site` namespace could go away. But I do think it adds some clarity around what exactly is transpiring.

- The project `uri` keys are only included as a fallback if there are no site-specific configurations. As I currently have this implemented, these values are immediately used within the `site` namespace (see my code comment), to provide values if none exist. I'm definitely open to changing this, but I didn't see a way to do so without introducing new configuration keys and breaking existing projects. However, any changes this requires could conceivably be addressed with an update hook.

phing/tasks/properties.xml
```diff
@@ -49,10 +49,47 @@
...
+
+  <!-- Set site aliases if not already set by ${site.name}.yml. -->
+  <property name="site.drush.aliases.local" value="${drush.aliases.local}" override="false"/>
+  <property name="site.drush.aliases.remote" value="${drush.aliases.remote}" override="false"/>
+  <property name="site.drush.aliases.ci" value="${drush.aliases.ci}" override="false"/>
+
+  <!-- Set site uris if not already set by ${site.name}.yml. -->
+  <property name="site.uri.local" value="${project.local.uri}" override="false"/>
+  <property name="site.uri.ci" value="${project.ci.uri}" override="false"/>
+
```
bobbygryzynger: This is the only place that the `project` namespace `uri`s are being used.

phing/build.yml
bobbygryzynger: Also, if you're open to it, perhaps these larger scale changes are best suited to a beta release, during which there's less promise of non-breaking changes and unforeseen issues can be ironed out - just a thought about how to move forward.

phing/build.yml
```diff
@@ -58,9 +60,9 @@ drupal:
 
 drush:
   bin: ${composer.bin}/drush
-  cmd: ${drush.bin} @${drush.alias} -r ${docroot} -l ${multisite.name}
+  cmd: ${drush.bin} @${drush.alias} -r ${docroot} -l ${site.uri.${environment}}
```
danepowell: Instead of manually specifying the site URI here, shouldn't that be defined as part of the Drush site alias? There would need to be a separate Drush alias for each site anyway.

kylebrowning: @danepowell http://github.com/acquia/club will also download all of your aliases for you so they should already exist for you if use that.

danepowell: @kylebrowning I haven't tried it yet, but just looking at the club alias command it doesn't appear to be multisite aware. e.g. it would generate one alias for each combination of subscription and environment, rather than one alias for each combination of _site_ (i.e. database) and environment.

bobbygryzynger: @danepowell - yeah, perhaps the `uri` argument should go away entirely and the alias argument should be multisite aware within BLT (something along the lines of `@site.drush.aliases.${environment}`). This is the approach that I've tried to take in managing previous multisite projects (i.e. an alias for each site environment) but currently BLT seems to favor using the `uri` argument. This works fine except when trying to do a `sql-sync ` which requires two aliases if you want to sync remote to local. 

This being the case, the PR tries to keep both intact, but making alias configurations responsible for this may be the cleaner approach. 

What do you think?

danepowell: Yeah I definitely think we need to rely on aliases and get rid of URI. I've already added some documentation encouraging folks to use aliases: #1049

bobbygryzynger: @grasmash - what are you thoughts on this? Any reservations? I'd be happy to modify/open a new PR to make this change.

danepowell: I can't speak for @grasmash but I suspect if you were able to break this up into smaller independent PRs it'd be a lot easier to review and merge.

bobbygryzynger: @danepowell - #1023 is a swipe at breaking it up and reducing the scope to just setup tasks, but at this point it keeps the `uri` argument in place but this PR is probably the best place to begin phasing its use out.

bobbygryzynger:
Closing this as #1092 addresses many aspects of this. Thanks @danepowell.

danepowell:
No problem @bobbygryzynger , thank you for the inspiration and hard work.

PR #860, opened 2016-12-24
Update bash console with new blt command

kieranacquia:
You need to run source ~/.bash_profile on OSX to update you console to use the blt command

Fixes # .

Changes proposed:
-
-


grasmash:
Good catch.

PR #861, opened 2016-12-24
specify directory where to find project.yml

kieranacquia:
be more specific about the path for new users trying to figure everything out

Fixes # .

Changes proposed:
-
-


grasmash:
Somehow, this change has already been committed.

PR #875, opened 2016-12-27
If you tap cask you also need to install cask

jakubsuchy:


jakubsuchy:
I was wrong about this. I was missing a brew update

PR #884, opened 2016-12-28
Preventing Site UUID conflicts during install.

grasmash:


phing/tasks/setup.xml
```diff
@@ -181,6 +181,8 @@
       <option name="account-name">"${drupal.account.name}"</option>
       <option name="account-pass">"${drupal.account.password}"</option>
       <option name="account-mail">"${drupal.account.mail}"</option>
+      <!-- Using ${cm.core.config-dir} does not seem to work here. Drush issue? -->
+      <option name="config-dir">${repo.root}/config/default</option>
```
grasmash: @danepowell Does this cause an issues with ${cm.core.config-dir} usage?

danepowell: The bigger problem here is that if you pass the `config-dir` parameter Drush performs a full import, i.e. _completely_ blows away existing configuration and replaces it with what's in config/default. This goes against BLT's current pattern of only running a partial import and would break a lot of existing sites.

References:
* https://github.com/acquia/blt/blob/8.x/phing/tasks/setup.xml#L219
* https://github.com/acquia/blt/pull/69

PR #886, opened 2016-12-29
Support different behat tags for deploy, pr, cron, local

aweingarten:

### DIFF
--- a/phing/build.yml
+++ b/phing/build.yml
@@ -17,6 +17,10 @@ behat:
     - ${docroot}/profiles
     - ${repo.root}/tests/behat
   tags: '~ajax&&~experimental&&~lightningextension'
+  tags-by-build-type:
+    pr: '~ajax&&~experimental&&~lightningextension'
+    deploy: '~ajax&&~experimental&&~lightningextension'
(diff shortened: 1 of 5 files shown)


grasmash:
Looks good. Can you add some documentation about this feature? It should be mentioned in ci.md.

aweingarten:
@grasmash  I just added documentation.  

This code has been merged to CB.  Going to test there for the 2-3 days before giving you the green light to merge.  It looks like the PR builds are working great.  Want to see a cron run and a deploy.

phing/build.yml
```diff
@@ -17,6 +17,10 @@ behat:
     - ${docroot}/profiles
     - ${repo.root}/tests/behat
   tags: '~ajax&&~experimental&&~lightningextension'
+  tags-by-build-type:
```
grasmash: Looking at this with fresh eyes, it seems like we should allow for more than dynamic behat tests. We could, for instance, allow a user to just define a bash script that gets called.

grasmash:
Closings pull request as per the latest comment. I'd like to support nightly builds, but I think that we need to broaden the scope beyond Behat tags.

PR #890, opened 2016-12-29
Fixing Ansible version documentation

### LINKED ISSUE #889: Ansible needs to be at least 2.2
Installation mentions Ansible but default brew installs Ansible 2.0 and 2.2 is required. Update install.md to require a minimal Ansible version.

jakubsuchy:
Fixes #889 

Changes proposed:
-
-

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -30,9 +30,10 @@ If you'd like to create a VM with BLT, you will require the following additional
 
         brew tap caskroom/cask
         brew install ansible
+        brew upgrade ansible
         brew cask install virtualbox vagrant
 
-The minimum required versions are VirtualBox 5.1.x and Vagrant 1.8.6. 
+The minimum required versions are VirtualBox 5.1.x, Vagrant 1.8.6 and Ansible 2.2.x. 
 
 The local PHP environment should also have a memory limit of at least 2G for BLT to initialize. You can modify your PHP CLI's memory limit by editing php.ini. You can use the following command to open the correct php.ini in TextEdit. Set `memory_limit = 2G`.
 


INSTALL.md
```diff
@@ -30,9 +30,10 @@ If you'd like to create a VM with BLT, you will require the following additional
 
         brew tap caskroom/cask
         brew install ansible
+        brew upgrade ansible
```
grasmash: You shouldn't need to upgrade ansible if you just installed it?

jakubsuchy: I am unsure how to do this right - i installed ansible and it gave me ansible 2.0, i had to run brew upgrade ansible to get 2.2. Is there a way to request a minimal 2.2 version?

grasmash: Hm. You probably needed to update brew itself first to get the latest formulaes.

INSTALL.md
```diff
@@ -30,9 +30,10 @@ If you'd like to create a VM with BLT, you will require the following additional
...
         brew cask install virtualbox vagrant
 
-The minimum required versions are VirtualBox 5.1.x and Vagrant 1.8.6. 
+The minimum required versions are VirtualBox 5.1.x, Vagrant 1.8.6 and Ansible 2.2.x.
```
geerlingguy: Ansible isn't technically required at all—if not installed locally, it will be installed at the right version inside the VM (and this is preferable on non-unixy environments). I'd rather not add it as a requirement in the BLT docs.

INSTALL.md
geerlingguy: It looks like in this case, you might have installed Ansible a while back, and just need to update it on your system (same as you'd need to do with VirtualBox or Vagrant).

Homebrew usually has the latest version within a day or less of official release... and checking now, it looks like this is the version you'll get if you install `ansible`:

```
$ brew info ansible
ansible: stable 2.2.0.0 (bottled), HEAD
```

grasmash:
Agreeing with @geerlingguy here.

PR #905, opened 2017-01-04
Ignore local.services.yml

danepowell:
As a result of #797, a lot of files that should be ignored are no longer ignored. One of those is local.services.yml.

@dooleymatt can you double check this meets our requirements?


dooleymatt:
@danepowell `box/local.config.yml` is also being tracked after the most recent update.

PR #906, opened 2017-01-04
Making search for blt bin a bit better.

grasmash:


grasmash:
This will be superseded by move to Robo.

PR #322, opened 2016-08-23
Add configuration for filesystem settings. (#211)

pixlkat:

PR #920, opened 2017-01-05
Ignore private files.

danepowell:
When creating a new project with BLT following [these instructions](https://github.com/acquia/blt/blob/8.x/readme/creating-new-project.md), after the `local:setup` step I get a /files-private directory in the repo root that should be gitignored.

I'm not quite sure _why_ this directory is created in the repo root instead of in the Drupal docroot. This PR seems like it would put it into the docroot: #322 :man_shrugging: 

Regardless, I'm definitely not seeing it at docroot/sites/*/files-private, which is where BLT apparently expects it to be following #732, so I figure we should change to the gitignore to match reality.

@dpagini under what circumstances were you seeing files show up at docroot/sites/*/files-private? I'm guessing it was for an older version of BLT?

### DIFF
--- a/phing/files/.gitignore
+++ b/phing/files/.gitignore
@@ -5,7 +5,7 @@ project.local.yml
 
 # Ignore paths that contain user-generated content.
 docroot/sites/*/files
-docroot/sites/*/files-private
+/files-private
 docroot/sites/*/private
 
 # Ignore build artifacts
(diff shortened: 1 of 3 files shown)


dpagini:
I think I was operating off an incorrect assumption, actually. This was because the [default.local.settings.php](https://github.com/acquia/blt/blob/8.x/settings/default.local.settings.php#L126) file sets the private directory locally. Since the gitignore already had the entry for `docroot/sites/*/private`, I was thinking this meant the expected place for local private files was that directory.
This makes more sense to me... but then I just wonder what that entry for `docroot/sites/*/private` is for at all?

PR #923, opened 2017-01-06
Remove factory-hooks from default template.

### LINKED ISSUE #599: Remove `factory-hooks` from default template
Shouldn't the `factory-hooks/` folder (and subsequent settings include file) be created as part of the `$ blt acsf:init` command?
Happy to attempt a PR for this, but wanted to get some feedback before I did anything.

dpagini:
Fixes #599.

Changes proposed:
- Delete `template/factory-hooks` directory, as not all users of BLT are on Site Factory.
- Move the hooks file to `settings/acsf/`.
  - This location could change if preferred. I originally was going to put these at `scripts/acsf/` and move the one file manually, but then I opted to move an entire folder so that other settings files could be loaded in the future.
- Create a new phing target `acsf:init:hooks` which creates the new folder, and prompts the user to add the folder to their repo.

All feedback is welcome!


phing/tasks/acsf.xml
```diff
@@ -20,4 +20,10 @@
     <echo>An example alias file for ACSF is located in /drush/site-aliases/example.acsf.aliases.drushrc.php.</echo>
   </target>
 
+  <target name="acsf:init:hooks">
```
grasmash: Please add a description and `hidden="true"`. Thanks!

dpagini:
Added as requested.

PR #930, opened 2017-01-06
DB updates should always be run before config imports.

danepowell:
This is sort of a follow-up to #911, since I just saw an example of it failing in the wild.

As I mentioned in that ticket, db updates _must_ be run before configuration imports so that any dependent modules can be enabled. Otherwise there's no way to add a configuration with a dependency on a new module.

Of course it's possible that something in your update hook might depend on a configuration change, in this case you should manually import just the necessary configuration as part of your update hook. This was the best practice with Features in D7 and I don't see any alternative in D8.


grasmash:
@danepowell Is there any sort of documentation change that should be made to accompany this?

danepowell:
We have several lines of documentation before the update step, and I added one more:
https://github.com/acquia/blt/pull/930/files#diff-d6981a61601a737ce906a130556d199fR218

I think that should be sufficient, unless you want to move all of it to a separate readme.

PR #946, opened 2017-01-09
Adding behat.extra property for arbitrary CLI args.

grasmash:


swichers:
This doesn't work as expected. Wouldn't tests.xml need to be updated as well?

PR #947, opened 2017-01-09
Set PHP memory limit for Travis.

danepowell:
We just had a deploy fail because update hooks that passed on Travis failed on ACE, due to memory exhaustion.

Right now Travis is using some default PHP memory limit, I think either 512M or 1024M. This sets it to 256M, which is what DrupalVM uses by default and is the upper limit of what you'd see on ACE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,8 @@ before_install:
   - echo 'variables_order = "EGPCS"' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   # Ensure that always_populate_raw_post_data PHP setting: Not set to -1 does not happen.
   - echo "always_populate_raw_post_data = -1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
-  # Increase PHP memory limit.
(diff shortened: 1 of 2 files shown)


grasmash:
Hm. The notable difference is that we do things on Travis CI that we wouldn't do on ACE. E.g., we don't even `composer install` on ACE.

danepowell:
True, although I just tried this on a project and had no problem with the decreased memory limit. And consider that we run the same memory limit in DrupalVM every day and have no problem there.

You're probably right that it doesn't make sense for TravisCI to have a limit as low as ACE, but it _should_ at least match the limit used by DrupalVM, which is what this PR does.

grasmash:
This won't ever cause a BLT build to fail, so let's add this to /.travis.yml too. That should provide a canary.

danepowell:
It looks like the memory was originally upped as part of #364 because the `acsf:init` task needed 1GB+ to run `composer require drupal/acsf:~8`. That seems a bit ridiculous, hopefully it was just a memory leak with a older version of composer. 

If you want to be totally sure, I'd clear the Travis caches and then re-run the test on my latest commit.

PR #957, opened 2017-01-10
Modify blt alias and bin/blt to allow for nested repo scenarios.

timcosgrove:
This is more of a proposal. Stating up front that the BLT core contributors may have thought of this scenario and there may be good reasons to not do this.

One thing that is causing my particular build a fair amount of pain is that BLT's alias and vendor/bin commands rely on being in the same repo as BLT's vendor directory. This precludes submodule scenarios and Drupal-specific VCS package scenarios where we want a package-in-progress (say, a custom profile that is shared among multiple builds) to exist within a BLT project. 

In our particular case, we use a VCS package in our BLT-based project's composer.json to include a profile. We store the profile in a separate repository. When composer installs this, it puts it in docroot/profiles/contrib, and the profile contains its .git directory. We .gitignore the profile. This allows us to work in a single codebase (from an IDE perspective at any rate) but to keep the code bases separate.

BLT currently cannot work properly inside our profile directory because it is its own git repo, and therefore the BLT alias cannot find the appropriate vendor directory. We've been working around this, but https://github.com/acquia/blt/pull/922 has thrown a wrench into this.

The proposal is to make the blt alias and bin/blt both recursive functions that look for acquia/blt in the vendor directory, as before, but if it is not found, go up one directory, and try again. This continues until BLT is found or we are no longer in a  git repo.

Appreciate any feedback. Thank you!

### DIFF
--- a/bin/blt
+++ b/bin/blt
@@ -5,7 +5,7 @@ set -o pipefail
 # set -o xtrace
 
 function run_blt() {
-  # If we are in a git repository, find the repo root.
+  # If we are in a git repository, go to the repo root.
   if [ "`git rev-parse --show-toplevel 2> /dev/null`" != "" ]; then
     GIT_ROOT=$(git rev-parse --show-toplevel)
     cd $GIT_ROOT
@@ -14,8 +14,8 @@ function run_blt() {
       ./vendor/acquia/blt/blt.sh "$@"
       exit 0
     else
-      # Go up one directory, 
-      # which will either take us into a parent repo or out of any Git repo.
+      # Go up one directory, which will either take us
+      # into a parent repo or out of any Git repo.
       cd ../
       # Recurse.
       run_blt "$@"
@@ -27,5 +27,5 @@ function run_blt() {
   fi
 }
 
-# Invoke the function
+# Invoke the function.
 run_blt "$@"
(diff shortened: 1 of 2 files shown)


grasmash:
I have definitely been wanting to make this change and I just haven't done it yet. Thanks for the contribution!

grasmash:
One concern I have here is that many people already have the blt alias in their .bash_profile files locally. I think we'll need to create an update script for this, or else try to limit to vendored files. I'd love to minimize the logic that is stored on each user's machine.

timcosgrove:
Yeah, I didn't try to address that. I know there are various mechanisms that BLT uses to update existing files...I didn't know if something existed for modifying the alias.

I can take a pass at that if you like, though it would essentially be bash grep voodoo.

grasmash:
@timcosgrove I created something akin to `hook_update_n()`, feel free to take a stab at it:
https://github.com/acquia/blt/blob/8.x/src/Update/Updates.php#L120

We can probably just look for the alias in an array of possible file, and if we find it, replace it. Would take a little bit of regex and string manipulation to replace only the right portion of the file. Might be nice to split into a dedicated function so that it's available if the alias needs to be changed again in the future.

timcosgrove:
Apologies I haven't been able to provide an update just yet. I should have some time with the weekend now here.

grasmash:
Closing. 8.x of BLT will attempt to find climb upwards 10 directories to discover the repo root.

PR #960, opened 2017-01-11
Fixes #649: Adding Pipelines build to BLT testing.

### LINKED ISSUE #649: Execute Pipelines build during BLT tests
BLT should execute a build on Pipelines in addition to or in place of the Travis CI build that is run for each pull request and 8.x commit.

Requirements:

- [x] Download Pipelines client
- [x] Kickoff Pipelines build using template .acquia-pipelines.yml

Implementation:

This is tricky. BLT is not actually what we want to test--we want to test a project created with BLT. Ideally, this would mean actually using the acquia-pipelines.yml file that a BLT project would use, but that doesn't seem possible. Unfortunately, it means we will have a root acquia-pipelines.yml file, and another one that is used for new projects, located in `scripts/pipelines`. We need to find a way to address this issue--we need to test the acquia-pipelines.yml file that will actually be used by BLT projects.

One possibility would be to request that acquia-pipelines.yml adds an 'import' feature. This would allow one yml file to import other "stock recipes" which actually could be quite useful for other users. This feature is used in Behat configuration: http://docs.behat.org/en/v2.5/guides/7.config.html#imports

grasmash:


grasmash:
Closing in favor of https://github.com/acquia/blt/pull/1039

PR #962, opened 2017-01-11
Auto start VM.

danepowell:
About twice a day I run `local:refresh` while forgetting to spin up the DrupalVM first. Kind of annoying because it takes a while to fail hard and you end up losing 5-10 minutes if you're not paying attention.

This simply runs the VM at the start of the refresh target if a Vagrantfile is detected.

If you like this idea, there would be ways to improve it... such as actually checking for the presence of the stack (whether VM or not), and also putting a property into project.yml so that we can look for the "right" stack a little more intelligently and also allow people to override it locally if they can't use the project default for whatever reason.

### DIFF
--- a/phing/tasks/local-sync.xml
+++ b/phing/tasks/local-sync.xml
@@ -6,9 +6,9 @@
     </phingcall>
   </target>
 
-  <target name="local:refresh" description="Refreshes local environment from upstream testing database." depends="setup:build, local:sync, local:update"/>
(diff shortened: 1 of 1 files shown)


danepowell:
Eh... this adds a few seconds to every BLT command which I'm not thrilled about, so I'd rather do this a more efficient away if we do it at all. Unfortunately even `vagrant status` takes 2-3 seconds to run on a fast machine so I'm not sure how to do this without impacting performance.

PR #964, opened 2017-01-11
Fixes #963: behat.extra is not actually used.

### LINKED ISSUE #963: New behat.extra option is not actually passed along.
* BLT version: 8.6.10

When I run this command:
```
blt tests:behat -Dbehat.extra="-dl"
```

I get the following output:
```
Normal test output
```

There should be a Behat definitions list.

swichers:
Fixes #963 .


grasmash:
I think we need to add `behat.extra: ''` to build.yml

grasmash:
Replaced by #970.

Issue #965, opened 2017-01-11
Fixes #1138: Support Config Split for environment-specific Core CMI


geerlingguy:
This is a feature request, and it's something I'm still working on conceptually—it could also dovetail into a more nuanced approach in general to how BLT handles different Configuration Management scenarios.

Currently the default is to either do:

  - Core CMI with `--partial` (which results in difficulty deleting fields and content types, etc.—you have to use a lot of update hooks and APIs to do cleanup work)
  - Features + Core CMI with `--partial` (same caveats, but often better for multisite-style projects since feature 'packages of configuration' can be enabled or disabled per-site).

I've been building a site using the [Configuration Split](https://www.drupal.org/project/config_split) module for Drupal, and it seems to give me the best of both worlds for many aspects (block content management is still an issue... but that's a separate thing for a separate day):

  - Works just like Core CMI (export and import everything. No need to deal with features workflow/creation/management issues, or the rogue `drush fra` ghost that requires extra `drush cc drush`es to clear up. Makes managing massive changes (e.g. deleting lots of fields and adding others, same with other config entities) a lot easier).
  - Allows 'splits' for different environments (e.g. enable Shield in Dev and Stage, enable Devel and Stage File Proxy in Local only, don't enable certain modules in CI, etc., and have some configuration differences per environment (e.g. different Google Analytics settings per environment without hacking settings.php to kingdom come).

To use Configuration Split, besides the normal setup (documented on the module page), I needed to make a few modifications to a few of BLT's Phing build targets. So far I've done the following:

First, change `import` inside `blt/project.yml` to add my custom Phing target overrides:

```
import: '${repo.root}/blt/build.xml'
```

Second, create `blt/build.xml` with said overrides for the `setup:update` and `setup:drupal:install` targets (**Note**: The things I added or changed in these targets are marked by a comment line with `Addition`):

> See this gist: https://gist.github.com/geerlingguy/1499e9e260652447c8b5a936b95440fa

The `setup:update` customized target basically swaps out `config-import --partial` for `config-split-import --split=[environment-here]`.

The `setup:drupal:install` customized target just adds in one line to pass the `--config-dir` option to Drush, so fresh site installs will handle existing UUIDs properly (this makes it so multisite installs or CI installs still work even with full exported CMI data).

geerlingguy:
[...]

PR #966, opened 2017-01-11
Connects to #965: Add --config-dir option to Drush site-install command.

geerlingguy:
Relates to Issue #965.

Changes proposed:
- Add `--config-dir` option to Drush `site-install` command.
- Empower sites using full CMI exports (either via core CMI or with Config Split) to be installed in CI or other environments (otherwise the install fails on UUID issues).

Note: This should be tested on sites that are currently _not_ using full CMI export/import to make sure nothing is broken in the build in that situation.

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -179,6 +179,7 @@
       <option name="account-name">"${drupal.account.name}"</option>
       <option name="account-pass">"${drupal.account.password}"</option>
       <option name="account-mail">"${drupal.account.mail}"</option>
+      <option name="config-dir">../config/default</option>
       <param>"${project.profile.name}"</param>
       <param>"install_configure_form.update_status_module='array(FALSE,FALSE)'"</param>
     </drush>


grasmash:
@geerlingguy just FYI, you need to use one of the following keywords to "connect" an issue in GitHub to another:

close
closes
closed
fix
fixes
fixed
resolve
resolves
resolved

E.g., "Fixes #965" rather than "Issue #965"

geerlingguy:
@grasmash This woudn't close that issue though... it's just related. E.g. the first little bit that would go a ways towards helping two use cases. So in this case I still hope the PR can get merged, but I don't want it to close #965.

geerlingguy:
Error is:

```
Import the listed configuration changes? (y/n): y
Drupal\Core\Config\ConfigImporterException: There were errors        [error]
validating the config synchronization. in
Drupal\Core\Config\ConfigImporter->validate() (line 728 of
/home/travis/build/acquia/blt-project/docroot/core/lib/Drupal/Core/Config/ConfigImporter.php).
The import failed due for the following reasons:                     [error]
This import is empty and if applied would delete all of your
configuration, so has been rejected.
```

Hmm... so our automated tests would only work if you have at least a _tiny_ bit of config exported.

grasmash:
@geerlingguy Re:keywords. Fair enough. In that case, let's use Waffle.io's keywords so that issues are visually connected on our Kanban board:

connect to
connects to
connected to
connect
connects
connected

geerlingguy:
We have a kanban board??

geerlingguy:
@grasmash - I updated the commit message at least—haven't looked into making Travis happy.

grasmash:
@geerlingguy:

> The import failed due for the following reasons:                     [error]
This import is empty and if applied would delete all of your
configuration, so has been rejected.

geerlingguy:
@grasmash - Yeah... not sure how to deal with that—this is necessary _after_ you've done your first config dump, but it doesn't work until that's been done. Chicken and egg issue.

Maybe we can have a flag/if/else thing that can say "enable this in the build" and have it off by default?

grasmash:
I think this is basically a duplicate of https://github.com/acquia/blt/pull/884

geerlingguy:
@grasmash - Yeah, basically. It would be nice if we had a flag that you could set to enable this option once the project is up and running—and only if you choose to use full CMI (and not `--partial`).

geerlingguy:
That would also allow us to default to the current behavior, which can pass tests here in Travis since there isn't any config to kick off a brand new site/project.

geerlingguy:
Closing since https://github.com/acquia/blt/pull/1080 got in.

PR #978, opened 2017-01-12
Ignore compiled theme CSS and other artifacts

danepowell:
Cog is about to remove their own .gitignore, and in general it seems that it's the responsibility of the parent project to ignore compiled CSS: https://github.com/acquia-pso/cog/issues/29

Once this and the Cog issue are merged, Cog-based themes should work out of the box with any BLT project (including deployment of CSS).

### DIFF
--- a/template/.gitignore
+++ b/template/.gitignore
@@ -23,8 +23,10 @@ docroot/profiles/contrib
 docroot/libraries
 drush/contrib
 
-# Ignore custom theme node_modules folders
+# Ignore custom theme build artifacts
 docroot/themes/custom/*/node_modules
+docroot/themes/custom/*/css
+docroot/themes/custom/*/styleguide
 
 # Ignore build artifacts
 /deploy


geerlingguy:
The only problem here is that there are some projects which want to compile everything and commit it to their repo, so we'll have the opposite problem. But, as long as they can customize their own .gitignore files, it might not be an issue.

sarahjean:
@geerlingguy @danepowell Maybe a comment in there to that effect?

danepowell:
This is the same paradigm that we follow for all sorts of other build artifacts (contributed packages, front-end libraries, etc...) so I'm not sure it makes sense to document the exception inline here. I think it would be better to have separate documentation on how to commit source and build artifacts together (although I'm not sure why you'd want to use BLT at all in that case)

geerlingguy:
@danepowell - True... in one case we have a theme that _used_ to use CSS and is now using SASS, but during the transitional period, to minimize weird issues, we're committing compiled CSS to the theme (until we get all the automation working correctly).

For any new projects (and as a general rule) there shouldn't be a need for an exception.

geerlingguy:
But it would be good to at least warn people strongly in the CHANGELOG/release notes for the next release, in case they're in the same situation I'm in for that particular site.

danepowell:
Definitely noting it in the release notes would be wise.

PR #996, opened 2017-01-18
Downgrade PHPUnit to version used by Drupal core (~4.8)

TravisCarden:
PhpStorm seems to be incompatible with PhpUnit 5 (or some of its dependencies). Can we downgrade to the version used by Drupal core (~4.8)? Tracking core on that point seems like the safest thing to do anyway, since many people will want to write unit tests that will work on the Drupal.org testbot and some people may even want to run core's unit tests; and the safest thing in both those cases is to use the same version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "drupal/console":                   "^1.0.0-rc13",
         "phing/phing":                      "^2.15.0",
         "php":                              ">=5.6",
-        "phpunit/phpunit":                  "^5.5.4",
+        "phpunit/phpunit":                  "^4.8",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         "drupal/console":                   "^1.0.0-rc13",
         "phing/phing":                      "^2.15.0",
         "php":                              ">=5.6",
-        "phpunit/phpunit":                  "^5.5.4",
+        "phpunit/phpunit":                  "~4.8",
```
grasmash: Please specify this as a minimum version. E.g., ^4.8

TravisCarden: @grasmash Done.

PR #1005, opened 2017-01-21
Add multisite support for setup tasks

bobbygryzynger:
Addresses aspects of #799.

Changes proposed:
- Allow multisite configurations to be supplied in per-site yaml files.
- Perform setup tasks for each multisite.

### DIFF
--- a/phing/tasks/properties.xml
+++ b/phing/tasks/properties.xml
@@ -69,9 +69,9 @@
       <propertyregex subject="${multisite.names}" pattern=",$" replace="" property="multisite.names" override="true"/>
       <!-- Load default site properties if overrides are specified. -->
       <if>
-        <available filepath="${blt.config-dirs.sites}" file="${site.name}.yml" />
+        <available filepath="${blt.config-dirs.sites}" file="project.${site.name}.yml" />
         <then>
-          <property file="${blt.config-dirs.sites}/${site.name}.yml" prefix="site" logoutput="false"/>
+          <property file="${blt.config-dirs.sites}/project.${site.name}.yml" prefix="site" logoutput="false"/>
         </then>
       </if>
     </then>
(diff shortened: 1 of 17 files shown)


bobbygryzynger:
@grasmash - I've taken a stab here at breaking up #827 a bit, let me know what you think.

phing/tasks/properties.xml
```diff
@@ -56,7 +56,43 @@
...
-  <echo level="verbose">Executing commands against multisite "${multisite.name}"</echo>
+  <!-- Set the default site to operate against. -->
+  <property name="site.name" value="default" override="false"/>
+
+  <!-- Load multisite-specific overrides. -->
+  <if>
+    <available file="${blt.config-dirs.sites}"/>
+    <then>
+      <!-- Build the site list from *.yml files in the blt/sites directory. -->
+      <exec dir="${blt.config-dirs.sites}" command="ls *.yml | tr -d '\n' | sed s/\.yml/,/g" outputProperty="multisite.names"/>
```
grasmash: I think we may need a more blt-specific pattern, since this could pick up unrelated yml files. Perhaps project.site.yml.

grasmash:
Thanks for breaking this into something smaller. I think that the last step before merging is to ensure that there's some type of test coverage. Perhaps we can add a default project.site.yml to a directory during BLT's .travis.yml execution and assert that it has been parsed and generated the expected settings files.

kylebrowning:
👍

bobbygryzynger:
@grasmash - working on adding some tests here, but ran into #1011.

bobbygryzynger:
@grasmash - I've put some additional testing in place here. I wasn't entirely sure how to go about testing that site properties are being correctly parsed, but take a look at `PropertiesTest.php` and see if you feel like it's a reasonable approach.

PR #1018, opened 2017-01-27
Fixes #1017: Running BLT commands in Drupal VM requires bz2; update to Drupal VM 4.2.x.

### LINKED ISSUE #1017: Running BLT commands within Drupal VM requires bz2 extension
My system information:
* Operating system type: Windows
* Operating system version: 10, Anniversary Edition
* BLT version: 8.6.10

When I run this command:
```
composer blt-alias
```

I get the following output:
```
Something to the effect of "bz2 extension is required".
```

Copying from the Windows laptop to this issue is excruciating, so trust me on it.

* The BLT alias should be installed successfully.

To fix, we need to make sure the bz2 extension is available inside the VM (for the benefit of Windows users and others running stuff inside their VM instead of on the host machine). PR incoming.

geerlingguy:


geerlingguy:
Still testing... now I'm getting "../acquia/blt/bin/blt: not found" :/

geerlingguy:
Weird, it seems all kinds of `blt` commands are trying to find paths like `../phing/phing/bin/phing` or `../acquia/blt/bin/blt`, but when I check, all those folders and executables exist and work independently in the right paths inside the `vendor` directory.

This PR is still needed... but I'm hitting a wall at this point in my quest to set up instructions for doing literally everything (from no project to full BLT project) within Windows only.

geerlingguy:
Turns out I had to reboot entirely, and now I can delete that folder from within Ubuntu Bash, using `sudo rm -rf vendor`.

geerlingguy:
And I wasn't running Cmder as administrator, so symlinks were all catawampus.

geerlingguy:
And I finally have a fully functional huge set of instructions from going to no project to fully-functional BLT project with Drupal VM on Windows that I'll do a screencast of, then post to my blog. I didn't know if it would be possible, but you can actually do every single thing inside Windows 10 native environment.

geerlingguy:
(As long as this PR is merged, that is).

geerlingguy:
Tagged a new Drupal VM release so this change works correctly.

geerlingguy:
...and here's a blog post on the entirety of the process, along with a splendiferous video: https://www.jeffgeerling.com/blog/2017/set-blt-and-drupal-vm-entirely-within-windows-10

Issue #977, opened 2017-01-12
Require blt validate:phpcbf


jayakrishnanj:
Its not a bug, requesting an enhancement.

Now we have blt validate:phpcs for validating, which we can use locally to validate the Code. Similarly if we have blt validate:phpcbf to fix the CS issue, it would help us all.

grasmash:
Yeah I agree, I've wanted to create this feature for some time.

dooleymatt:
@grasmash, @jayakrishnanj I'd be happy to take this on if you'd like.

grasmash:
@dooleymatt Yes please!

TravisCarden:
I, too, would like to see this command. @dooleymatt, I merely request that 1) it doesn't go under `validate` because it's not merely a validator--it's a fixer. And 2) for the same reason, it doesn't run automatically. Some of us would riot in the streets if our pre-commit hook one day started changing our code without asking. :)

dooleymatt:
@TravisCarden sounds good. Thanks for the input. 

I've been struggling to figure out which files should be fixed with this function. So far, my thinking has been that the most common use of this function is probably going to be to fix files that are flagged with PHPCS violations in the pre-commit hook. Which, to me means that the files that are fixed should probably default to the files that are staged for commit. I agree that it shouldn't run automatically, however, I feel like there should either be an interactive prompt to the user asking if they want to automatically fix violations when they are found, or a message that just says run `blt phpcbf` to fix the violations. 

Additionally, I think that there should be PHPCBF tasks that would mirror the PHPCS tasks. So, one that would fix all custom code (with the same filters as are used in the PHPCS task), one that would fix a specific fileset, and one that would fix a single file. But, I feel like the default behavior should be to only fix staged files. 

@TravisCarden, @grasmash, @jayakrishnanj do you guys agree, or do you have any other thoughts on this?

TravisCarden:
In the spirit of the Unix philosophy, I would argue for the primary use case being to run on an arbitrary list of files and directories provided by command line argument. That way it has the most generic interface possible, which supports both developer use and piping out-of-the-box. From there it would be easy to write integrations that, for example, pipe in a list of uncommitted files on a given hook. If you wanted to, you could default the scope if no arguments are provided, and cover more or less all your bases.

dooleymatt:
> you could default the scope if no arguments are provided

@TravisCarden I think I agree completely, but, could you explain what you mean by this? Are you saying that the files argument should default to the list of staged files?

[...]

PR #1019, opened 2017-01-28
Feature/blt phpcbf

dooleymatt:
Fixes # 977.

Changes proposed:
- Adds `fix:phpcbf` target to automatically fix violations found by phpcs.

### DIFF
--- a/tests/phpunit/BltProject/FixTest.php
+++ b/tests/phpunit/BltProject/FixTest.php
@@ -10,14 +10,67 @@ use Acquia\Blt\Tests\BltProjectTestBase;
  * Verifies that fix related tasks work as expected.
  */
 class FixTasksTest extends BltProjectTestBase {
-  
+
+  private $return_var;
+  private $output = [];
+
+  /**
+   * Tests for existence of phpcbf script.
+   */
+  public function testFixPhpcbfExists() {
+    $this->executePhpcbfFileListScript();
+    $this->assertNotContains("PHP Code Beautifier was not found in this project's bin directory. Please run composer install.", $this->output);
+    $this->assertEquals(1, $this->return_var);
+  }
+
+  /**
+   * Tests the handling of a missing file parameter.
+   */
   public function testFixPhpcbfExitsWhenMissingArguments() {
+    $this->executePhpcbfFileListScript();
+    $this->assertContains("Missing file list parameter.", $this->output);
+    $this->assertEquals(1, $this->return_var);
+  }
+
+  /**
+   * Tests that a file list is generated.
+   */
+  public function testFixPhpcbfWithArguments() {
+    $report = $this->createTestReport();
+    $this->executePhpcbfFileListScript($report);
+    $this->assertNotContains("Missing file list parameter.", $this->output);
+    $this->assertContains("Files that can be fixed by PHPCBF:", $this->output);
+    $this->assertContains("{$this->drupalRoot}/foo.php", $this->output);
+    $this->assertEquals(0, $this->return_var);
+  }
+
+  /**
+   * Helper method that xecutes the phpcbf-file-list.sh script.
+   *
+   * @param string $file
+   *  The full path to the file to be processed.
+   */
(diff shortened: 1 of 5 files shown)


dooleymatt:
@grasmash this is ready to be reviewed. What action should I take to indicate that?

grasmash:
@dooleymatt Please add a test for this feature. Ideally, you should have a tests that asserts both positive and negative cases.

dooleymatt:
Thanks @grasmash. I took a look at the existing tests to see how to write one for this, but I'm not really sure what I should I be testing.  It seems like most of the existing tests check for files that have been added, removed, or changed, but I'm not sure what similar tests could be performed for this new task. I am hoping either you or @TravisCarden could offer some direction. Thanks!

TravisCarden:
I'd be happy to chat about it, @dooleymatt.

dooleymatt:
@grasmash Test has been added.

tests/phpunit/BltProject/FixTest.php
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Helper method to create a temp file, formatted as a phpcs csv report.
+   */
+  protected function createTestReport() {
+    $tmp_file = tempnam(sys_get_temp_dir(), '');
+    if (!$tmp_file) {
+      throw new \Exception("Unable to create temporary file.");
+    }
+    $file_contents = "\"{$this->projectDirectory}/docroot/foo.php\",1,1,error,\"Line indented incorrectly; expected 2 spaces, found 4\",Drupal.WhiteSpace.ScopeIndent.IncorrectExact,5,1";
+    $command = "echo '$file_contents' > {$tmp_file}";
```
TravisCarden: @dooleymatt [`file_put_contents()`](http://php.net/manual/en/function.file-put-contents.php) would be simpler, since we're writing PHP. (I realize I put you on the wrong scent by using the word "echo".)

tests/phpunit/BltProject/FixTest.php
```diff
@@ -0,0 +1,73 @@
...
+    $report = $this->createTestReport();
+    $this->executePhpcbfFileListScript($report);
+    $this->assertNotContains("Missing file list parameter.", $this->output);
+    $this->assertContains("Files that can be fixed by PHPCBF:", $this->output);
+    $this->assertContains("{$this->drupalRoot}/foo.php", $this->output);
+    $this->assertEquals(0, $this->returnVar);
+  }
+
+  /**
+   * Helper method that xecutes the phpcbf-file-list.sh script.
```
TravisCarden: Typo: "xecutes".

tests/phpunit/BltProject/FixTest.php
```diff
@@ -0,0 +1,73 @@
...
+    $this->executePhpcbfFileListScript();
+    $this->assertContains("Missing file list parameter.", $this->output);
+    $this->assertEquals(1, $this->returnVar);
+  }
+
+  /**
+   * Tests that a file list is generated.
+   */
+  public function testFixPhpcbfWithArguments() {
+    $report = $this->createTestReport();
```
TravisCarden: @dooleymatt I'm confused: Don't you want to be creating a test file rather than a test report? A PHP file with standards violations in it so you can assert that PHPCBF correctly finds it?

dooleymatt:
@TravisCarden Thanks for reviewing this. I initially didn't include a test like you described because I felt like it might have been outside the scope of this to test whether or not phpcbf actually does what it's supposed to do. However, I've updated the test to include that now, and have also updated the way that the phpcs csv file report is generated.

grasmash:
@dooleymatt This looks great. My only request is that we put this in the `validate:` namespace rather than creating a new `fix:` namespace.

dooleymatt:
@grasmash I based that change on Travis's request here: https://github.com/acquia/blt/issues/977#issuecomment-274567320

grasmash:
Eh, alright. The argument makes sense.

PR #1023, opened 2017-02-01
Connects to #799: add multisite setup tasks

bobbygryzynger:
Connects to #799.

Changes proposed:
- Allow multisite configurations to be supplied in per-site yaml files.
- Perform setup tasks for each multisite.
- Add testing for multisite properties and settings files.


bobbygryzynger:
Closing this as #1092 addresses many aspects of this.

PR #1024, opened 2017-02-01
Add phing pre-config setup task

adelina-t:
Changes proposed:
- add task for setup before importing configuration


phing/tasks/setup.xml
```diff
@@ -202,7 +202,14 @@
 
   </target>
 
-  <target name="setup:config-import" description="Import configuration from the config directory.">
+  <target name="setup:pre-config" description="Setup before importing configuration.">
```
grasmash: Could we name this pre-config-import?

phing/tasks/setup.xml
```diff
@@ -202,7 +202,14 @@
...
+    <echo>Pre config setup... </echo>
```
grasmash: This echo statement should not be necessary, as the hook invokation itself echoes similar information.

phing/tasks/setup.xml
grasmash: I don't think this needs to be a separate target, it can just be rolled into setup:config-import.

phing/tasks/setup.xml
adelina-t: Agreed

phing/tasks/setup.xml
adelina-t: I would keep it separate, it seems clearer to me.

grasmash:
This has been merged in #1095. Thank you for contributing!

PR #1025, opened 2017-02-01
Fixes #1014: Don't run drush entity-updates during config-import

### LINKED ISSUE #1014: Don't run drush updb with the entity-updates flag during setup:config-import, it's dangerous
When BLT executes the `setup:config-import` task, it will run drush updb after importing configuration. It passes the `--entity-updates` flag to the updb.

I think this should be removed.

This [stack overflow answer](http://drupal.stackexchange.com/questions/221215/what-is-the-purpose-of-drush-entity-updates) indicates that it's really a development tool and shouldn't be run under normal circumstances.

There's also a [detailed change record](https://www.drupal.org/node/2554097) regarding the removal of the automated entity updates system from update.php and explains why the automated update system is dangerous and shouldn't be used normally. Basically modules should handle the entity updates themselves in update hooks.

Here's an [additional change record](https://www.drupal.org/node/2554101) that explains the user-facing changes of the removal of the automated system.

Given that BLT's acquia cloud hooks will run the setup:config-import task during deploys, it seems dangerous to include the `--entity-updates` flag in the updb command.

bkosborne:
Fixes #1014

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -216,9 +216,7 @@
     <!-- This must happen before features are imported or configuration is imported. -->
     <!-- For instance, if you add a dependency on a new extension to an existing configuration file, you must enable that extension via an update hook before attempting to import the configuration. -->
     <!-- If a db update relies on updated configuration, you should import the necessary configuration file(s) as part of the db update. -->
-    <drush command="updb" assume="yes" alias="${drush.alias}" passthru="false">
-      <option name="entity-updates"></option>
-    </drush>
+    <drush command="updb" assume="yes" alias="${drush.alias}" passthru="false"/>
     <!-- Import configuration. -->
     <drush command="config-import" assume="yes" alias="${drush.alias}" passthru="false">
       <option name="partial"></option>


grasmash:
@danepowell I think you may have initially added the entity update flag, any objections?

danepowell:
Thanks for bringing this up. I wasn't aware that best practice was to manually apply entity updates in individual update hooks, I thought everyone was expected to run `updb --entity-updates`.

I'd like to follow best practices, but it's worth calling out that this is a big change in workflow. We've used `--entity-updates` as long as it's existed and never had any problems. The alternative (handling updates manually) is pretty hideously complex, I doubt most developers are aware of it, and there's a risk of unknowingly allowing entity updates to stack up over time.

If we remove this, we should definitely document it, and ideally maybe we could have an automated test that checks for pending entity updates and throws an error if they exist.

bkosborne:
> It sounds like this transition could also solve the problem of not being able to update fields with existing content. If that's the case, it would be a huge win and I'd love to see it documented.

Well, it's still up to the "owner" of the field that changes to write the update hook to handle it that scenario, which I think is still a complete pain. Some examples I've seen have you capturing all the field data into memory, truncating the table, applying the schema update (which normally fails if there's content), then re-adds the content from memory.

grasmash:
@geerlingguy @danepowell What would you like to do about this? Been open a long time, I'd like to defer decision to one of you two since you have more experience with this.

geerlingguy:
@grasmash - Personally, I never run entity-updates unless the status report page yells at me and tells me to do it. I don't know how dangerous it is to do it by default... but I _haven't_ done it on most of my sites, ever, and only have done it manually once or twice when the status report tells me to do so.

danepowell:
I would say merge this. It's not a complete solution but it's a step in the right direction.

Just looking at this from a practical angle, the current system isn't very effective anyway, since we are running `updb --entity-updates` _before_ importing new configuration. For it to apply updates made by configuration changes, we would have to be running updb _after_ importing configuration as well. So the change in this PR at least makes things more internally consistent.

I still don't really understand the interplay been configuration updates and entity updates. Our switch to config_split throws another wrinkle into this. Is applying entity updates even necessary when you are using full config imports like with config_split? I'd like to see some follow up discussion about this.

geerlingguy:
> Is applying entity updates even necessary when you are using full config imports like with config_split?

In a word, yes... but I don't know why that can be the case; it seems there are some situations where the config is imported but something in the database still needs to be updated. Running the entity import after a deployment is over fixes it, though.

That's why I always have "look at the site status report page" as the last step of any deploy I run; sometimes something odd pops up there, and then I fix it.

bkosborne:
I agree that this whole aspect of D8 is confusing.

Here's three resources that may help you understand what's at play:

- http://drupal.stackexchange.com/questions/221215/what-is-the-purpose-of-drush-entity-updates
- https://www.drupal.org/node/2554097
- https://www.drupal.org/node/2554101

grasmash:
Closing in favor of rebased #1293.

danepowell:
Thanks. I'm still not totally clear on how this plays with the configuration management system. Let's say I have a field with no data in it, and I change its storage definition locally, export it, and then import it remotely (`drush config-import`). Will the storage type get fully changed by the configuration system, or do I then have to apply the entity updates manually (either via `drush updb --entity-updates` or my own update hook)?

The one case that is clear to me now is if you want to preserve data in a field while changing its storage definition... in that case, the examples in this article you linked are extremely helpful: https://www.drupal.org/node/2554097

danepowell:
I did some testing on this. Simple changes to a field (such as changing cardinality) don't seem to require any entity updates, the core configuration system handles them completely when you deploy them using config-export / config-import.

But when I tried to delete a field and recreate it as a different type (image -> decimal), very strange things happened. The change _seemed_ to take effect, but the status report shows pending entity updates, and what's really weird is that the pending updates have nothing to do with the field that was changed. Unfortunately, there doesn't seem to be any sort of "diff" mechanism to see exactly what Drupal thinks needs to be updated.

I've created an issue to follow up and try to get clarity on this: #1297

danepowell:
I've done more digging and will post more in that follow-up issue, but just to comment on the question of whether entity updates or dangerous not, which seems most relevant to this issue...

Yes, running entity updates can definitely be dangerous. For instance, if you were to change a field storage uuid, and then import the changed field storage definition, everything would appear to work fine. But after the next cron run, if you ran entity-updates if would delete and recreate this field, destroying any existing data in the process. So not only is it destructive, it's also a bit of a time bomb, at least the way that we were using it before. So removing it was probably the correct choice.

Issue #678, opened 2016-11-16
Why does BLT assume the usage of a "vcs" config directory?


bkosborne:
BLT assumes the usage of a "vcs" config directory, I think because Acquia sets this in the ACE config file that's included on all ACE sites.

I don't really understand why we should have both the "vcs" config directory and the default "sync" directory that Drupal assumes. With the "vcs" approach, I have to specify that directory when running drush config-import and config-export, otherwise drush assumes the default of "sync".

Maybe at the very least this value can be changed to use a variable instead, so we can override it in our project.yml file? https://github.com/acquia/blt/blob/8.x/phing/tasks/deploy.xml#L167

grasmash:
paging @danepowell. I think I had the same question.

danepowell:
Hi @bkosborne , BLT generally assumes that you are storing config in Git, i.e. at `/config/default`. BLT defines that to be the `sync` directory in [local.settings.php](https://github.com/acquia/blt/blob/8.x/settings/default.local.settings.php#L30), while ACE calls that the `vcs` directory.

Honestly, I'm not sure why either ACE or BLT settled on the convention they did. I suspect BLT chose the `sync` key just for ease of use, because Drupal generally assumes you are working with the `sync` directory unless you specify something else. ACE probably chose `vcs` because Drupal by default puts the sync directory into the files directory (which conflicts with how we typically use it).

I could see arguments for any of the following:

1. Leave everything as it is :)
2. Use `vcs` across the board, even locally, which would be more consistent and less confusing, but would require anyone running drush config-import/export to manually pass `vcs` as an argument. We could write a wrapper command to make that easier.
3. Make the deploy/local default config directory key configurable, to support other conventions and other deployment environments besides ACE.

bkosborne:
Thanks @danepowell 

Well I guess for now it makes sense to leave it as it is, I was mostly just confused

grasmash:
@weitzman Could you shed some like on this? I'm frankly still not sure why Drush and Acquia Cloud use a `vcs` dir rather than a `sync` dir.

geerlingguy:
I did a little more research, and it seems there's no particular reason why `vcs` was chosen... It just sort-of happened, and it's the default on Acquia Cloud. It seems to have been made the default around 8.0.0-rc2, and besides a few mentions in the documentation, there's no other public reasoning behind the ACE environments using `vcs` instead of the default `sync` directory used by Drupal core.

weitzman:
[...]

PR #1029, opened 2017-02-02
Don't force the config directory to be "vcs" during deploys

### LINKED ISSUE #1013: Make the cm.core.config-dir property configurable
Ok, so I thought I could deal with this, but every time I'm prompted for a config directory when I do a config import or export I die a little inside :)

Here's the main issue where I brought this up: https://github.com/acquia/blt/issues/678

Basically, Drupal wants there to be a config directory labeled 'sync' (it complains if it doesn't exist), but [BLT wants there to be one labeled 'vcs'](https://github.com/acquia/blt/blob/c6b27ddf0b16b49bbc1632a56fb7c11f946a8201/phing/tasks/deploy.xml#L224).

My site only uses one actual config directory. As a result I have this snippet at the very bottom of my site's settings.php file:

```
$config_dir = dirname(DRUPAL_ROOT) . "/config/$site_dir";
$config_directories['sync'] = $config_dir;
$config_directories['vcs'] = $config_dir;
```

That way they are both defined and BLT and Drupal are happy.

But because of that, whenever I do a config export or config import, I'm prompted to chose one of those, even though they are the same.

What I suggest doing here is changing [this line](https://github.com/acquia/blt/blob/c6b27ddf0b16b49bbc1632a56fb7c11f946a8201/phing/tasks/deploy.xml#L224) so that it accepts a parameter instead of hardcoding "vcs". Then I could configure it on my project.yml file.

bkosborne:
Fixes #1013

### DIFF
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -219,9 +219,6 @@
     <phingcall target="setup:update">
       <param name="environment" value="deploy"/>
       <property name="drush.alias" value="self"/>
-      <!-- Most sites store their version-controlled configuration in /config/default. -->
-      <!-- ACE internally sets the vcs configuration directory to /config/default, so we use that. -->
-      <property name="cm.core.config-dir" value="vcs"/>
     </phingcall>
   </target>
 


danepowell:
I'm fine with making the config directory configurable here, but I don't think we can remove it because deployments to ACE will start failing. On ACE it does need to be `vcs`.

bkosborne:
Hmm yeah I understand. Don't want an unexpected change to break deployments.

Note that I'm just removing the property _override_ here. The `cm.core.config-dir` property still exists and is used by the config import target. [It defaults to the 'sync' directory.](https://github.com/acquia/blt/blob/a1668961af1b10186ad31e943f28b5c1ca1da239/phing/build.yml#L39)

I wonder how people on ACE handle the annoyance of having two config directories defined. I don't think Acquia should have defined this 'vcs' directory in their config file to begin with.

danepowell:
Apparently you're not the only one :)

https://github.com/acquia/blt/issues/678

geerlingguy:
This can be closed now, in light of https://github.com/acquia/blt/pull/1080, correct @grasmash ?

grasmash:
Yes!

PR #1041, opened 2017-02-03
Add setup:toggle-modules test

bobbygryzynger:
Changes proposed:
- Add testing for module toggling behavior


.travis.yml
```diff
@@ -68,7 +68,7 @@ script:
   # Call targets in the new 'blt-project' project.
   - blt ci:build:validate:test -Dcreate_alias=false -Dbehat.run-server=true -Dbehat.launch-phantom=true -Dblt.verbose=true
   # Run 'blt-project' tests.
-  - phpunit ../blt/tests/phpunit --group blt-project
+  - phpunit ../blt/tests/phpunit --group blt-project --bootstrap ../blt/tests/phpunit/bootstraps/ci.php
```
grasmash: Let's put this in a phpunit.xml.dist rather than passing it as an argument

.travis.yml
```diff
@@ -68,7 +68,7 @@ script:
   # Call targets in the new 'blt-project' project.
   - blt ci:build:validate:test -Dcreate_alias=false -Dbehat.run-server=true -Dbehat.launch-phantom=true -Dblt.verbose=true
   # Run 'blt-project' tests.
-  - phpunit ../blt/tests/phpunit --group blt-project
+  - phpunit ../blt/tests/phpunit --group blt-project -c ../blt/tests/phpunit/phpunit.xml.dist
```
bobbygryzynger: @grasmash - I've added the `phpunit.xml.dist` file, but since `phpunit.xml` takes precedence, it still needs to be provided as a cli argument.

bobbygryzynger: @grasmash - I've added the `phpunit.xml.dist` file, but since `phpunit.xml` take precedence, this still needs to be provided as a cli argument. Was this your intention?

grasmash: @bobbygryzynger oh, no. I guess just adding to phpunit.xml. Forgot there was one!

bobbygryzynger:
@grasmash - turns out here (after a closer read) that the `phpunit.xml` has to be in the `cwd` to be read for test configuration:

> If phpunit.xml or phpunit.xml.dist (in that order) exist in the current working directory and --configuration is not used, the configuration will be automatically read from that file.

Since this is the case, how would you prefer to handle this? Provide the cli the path to `phpunit.xml` or cd into the `tests/phpunit` directory to execute the tests?

grasmash:
@bobbygryzynger provide the CLI path rather than cd into dir, if we can.

bobbygryzynger:
@grasmash - okay, I've updated the command to include the path to the xml configuration file.

bobbygryzynger:
@grasmash this looks to be [working as intended](https://travis-ci.org/acquia/blt/jobs/201305432#L1586), anything else you'd like to see here?

tests/phpunit/BltProject/ToggleModulesTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Verifies the modules for a given environment were enabled as expected.
+   *
+   * In the event no environment is specified, this test will be skipped.
+   *
+   * @group blt-project
+   */
+  public function testModulesEnabled() {
+    if (defined('BLT_ENV')) {
```
grasmash: Unit tests really shouldn't have conditionals in them. Will this ever be undefined?

bobbygryzynger: Only if the proper variables aren't set via `phpunit.xml`, at which point the tests will be skipped but won't produce an outright failure

bobbygryzynger: Would you rather this just assume the environment is `ci` and use the default alias?

grasmash: Yeah I think so. We should _always_ use phpunit.xml, right?

bobbygryzynger: Yeah, I agree with you there. I just didn't want to cause unexpected results here since this is the only test invokation that has the `phpunit.xml` being properly passed in at this point. But if we can assume that all phpunit tests will have access to the configuration I see no problem with removing the conditionals. 

Perhaps an issue should be opened to ensure that the travis scripts have this configured?

tests/phpunit/BltProject/ToggleModulesTest.php
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Verifies the modules for a given environment were enabled as expected.
+   *
+   * In the event no environment is specified, this test will be skipped.
+   *
+   * @group blt-project
+   */
+  public function testModulesEnabled() {
+    $modules = $this->config['modules'][BLT_ENV]['enable'];
```
bobbygryzynger: @grasmash - okay, no more conditionals.

grasmash:
@bobbygryzynger I rebased your PR and resubmitted here: #1098

PR #1053, opened 2017-02-08
TravisCI MySQL errors

### LINKED ISSUE #1036: Travis MySQL errors
This isn't a bug with BLT per se, but since many BLT projects rely heavily on Travis it's worth implementing a workaround.

Ever since moving to container-based builds, we frequently see builds failing with this error:

> ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)

According to Travis support, this could be due to MySQL not being fully initialized before we attempt to connect to it, and they suggest adding a sleep statement, i.e.:

```yaml
before_install:
  # Allow MySQL to finish starting.
  - sleep 3
```

I'm testing this on a few projects but since it's such an intermittent error it would be good to get more eyes on this so we can verify the workaround and incorporate it into BLT.

danepowell:
Fixes #1036

### DIFF
--- a/scripts/travis/before_install
+++ b/scripts/travis/before_install
@@ -2,6 +2,8 @@
 
 set -ev
 
+# Allow MySQL to finish starting
+sleep 3
 composer selfupdate
 # Disable xdebug.
 phpenv config-rm xdebug.ini
--- a/scripts/travis/before_install
+++ b/scripts/travis/before_install
@@ -2,7 +2,7 @@
 
 set -ev
 
-# Allow MySQL to finish starting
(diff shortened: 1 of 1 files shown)


geerlingguy:
@grasmash - Sadly, it looks like this may not be a workable solution... see https://github.com/travis-ci/travis-ci/issues/6842#issuecomment-278671414 specifically (from @danepowell )

danepowell:
Yeah I can confirm this didn't fix it. It's pretty harmless though, so I've been waiting to take it out until I find a fix.

PR #1054, opened 2017-02-08
Update lightning to 2.0.3

balsama:


grasmash:
This PR requires a new update hook in https://github.com/acquia/blt/blob/8.x/src/Update/Updates.php for 8.6.12.

It just needs to remove the patch from the composer.json of existing projects:
https://github.com/acquia/blt/blob/8.x/src/Update/Updates.php#L36

PR #1066, opened 2017-02-09
Removing asm89/twig-lint.

grasmash:


dpagini:
j/w - how come the twig linting is getting removed?

grasmash:
Just using a different library for it.

PR #1069, opened 2017-02-09
Adding vm:nuke command.

grasmash:


phing/tasks/vm.xml
```diff
@@ -127,4 +127,21 @@
...
+  <target name="vm:nuke" description="Destroys a Drupal VM and all of its associated configuration.">
+    <propertyprompt propertyName="vm.nuke" useExistingValue="true"
+                    promptText="This will destroy your VM and all of its configuration!!! Continue?"
+                    promptCharacter=":"
+                    defaultValue="y"/>
+    <if>
+      <equals arg1="${vm.nuke}" arg2="y" casesensitive="false"/>
+      <then>
+        <exec dir="${repo.root}" command="vagrant destroy" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
+        <exec dir="${repo.root}" command="rm -rf Vagrantfile box blt/project.local.yml drush/site-aliases/aliases.drushrc.php" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
```
geerlingguy: @grasmash - This is slightly more dangerous than I think it need be...

  - The user could have other customizations in the `box` dir that are useful to preserve (e.g. scripts, local.config.yml, etc.).
  - The `blt/project.local.yml` can also have helpful additions for that particular user too, which would need to be recreated.

I don't know if there's a better way to _guarantee_ that a new `blt vm` works correctly though :/

PR #1079, opened 2017-02-13
Fixing PHPCS filset reference.

dpagini:
I'm not sure if this is correct, but it seems like build.yml has a mistake referencing a fileset for code sniffing. The `files.frontend.custom.themes` fileset contains javascript, bower, npm, etc. files, not PHP code. Seems odd to send that through code sniffing. Should this be the `files.php.custom.themes` fileset instead?

### DIFF
--- a/phing/build.yml
+++ b/phing/build.yml
@@ -92,6 +92,7 @@ phpcs:
     - files.php.custom.modules
     - files.php.tests
     - files.php.custom.themes
+    - files.frontend.custom.theme
   haltonerror: true
   haltonwarning: true
 
--- a/phing/build.yml
+++ b/phing/build.yml
@@ -92,7 +92,7 @@ phpcs:
     - files.php.custom.modules
     - files.php.tests
     - files.php.custom.themes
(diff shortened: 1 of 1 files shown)


grasmash:
JS and CSS code can be sniffed via PHPCS using Drupal coding standards.

dpagini:
Ah ha, interesting. Should this have both lines then? Shouldn't `files.php.custom.theme` be added as well?

grasmash:
Yes I think that's right!

dpagini:
Updated. Thanks @grasmash.

phing/build.yml
```diff
@@ -91,7 +91,8 @@ phpcs:
   filesets:
     - files.php.custom.modules
     - files.php.tests
-    - files.frontend.custom.themes
+    - files.php.custom.themes
+    - files.frontend.custom.theme
```
grasmash: Looks like you accidentally removed the s.

dpagini: Doh! Fixing now.

PR #795, opened 2016-12-14
BLT-794: updating documentation to work around acquia cloud email san…

mikemadison13:
Changes proposed:
- Updated documentation to reflect work around given no ability to create extra cloud accounts with + in the email address.
- Added clarification regarding TA account access and Build failures
- Added clarification for drush sql-sync and rsync requirements in known host configuration

Issue #794, opened 2016-12-14
Acquia Cloud Account Creation + CI Instructions


mikemadison13:
The current instructions on https://github.com/acquia/blt/blob/8.x/readme/ci.md#setting-up-travis-ci-for-automated-deployments are currently not accurate, as the Acquia Cloud no longer allows you to register an account using an email address with a + in it.

mikemadison13:
see https://github.com/acquia/blt/pull/795

grasmash:
Merged #795

PR #1082, opened 2017-02-14
Update ci.md

danepowell:
This was last touched by @mikemadison13 in #794 / #795, which recommended using a personal account because Cloud no longer allowed plus signs. However, using a personal account is not very secure since it would allow anyone on a project to bypass access control and access any subscription associated with your account. It also means that deployments will begin to fail unexpectedly if you are removed from the project.

I just verified that you can create new Cloud accounts using emails with plus signs, as well as invite these accounts to any team, so I think we should go back to recommending the use of a dedicated account.

Note that we have internal tickets to track improvements to this process (NN-12088, CSPF-191).

### DIFF
--- a/readme/ci.md
+++ b/readme/ci.md
@@ -97,7 +97,7 @@ To set up the [workflow described earlier](#workflow), you must configure Acquia
 
          blt ci:travis:init
 
-1. Generate an SSH key locally for Travis / Build use. (e.g.)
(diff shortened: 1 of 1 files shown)


mikemadison13:
@danepowell the cloud team recently fixed this, I believe

danepowell:
Awesome that would make sense. Thanks.

PR #1092, opened 2017-02-14
Support for multisite deploys

danepowell:
This makes two significant changes:

1. The deploy site alias is changed from "self" to "sites". This will cause Drush to run the deploy:update commands against _all_ sites, not just the default site.
2. The `multisite.name` variable is currently overloaded. Sometimes it is used as an array, and sometimes as a single-value string. The only reason it works at all is because the output is identical in both cases for single-install sites. But for multisite projects, we need to decouple this in some places.

### DIFF
--- a/phing/build.yml
+++ b/phing/build.yml
@@ -67,9 +67,9 @@ drupal:
 
 drush:
   bin: ${composer.bin}/drush
-  cmd: ${drush.bin} @${drush.alias} -l ${multisite.name}
+  cmd: ${drush.bin} @${drush.alias} -l default
   dir: ${docroot}
-  uri: ${multisite.name}
+  uri: default
   assume: yes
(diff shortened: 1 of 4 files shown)


phing/build.yml
```diff
@@ -67,9 +67,9 @@ drupal:
 
 drush:
   bin: ${composer.bin}/drush
-  cmd: ${drush.bin} @${drush.alias} -l ${multisite.name}
+  cmd: ${drush.bin} @${drush.alias} -l default
   dir: ${docroot}
-  uri: ${multisite.name}
+  uri: default
```
danepowell: This change should have no effect for single-site projects. Multisite projects should pass either a unique cloud alias or override `drush.uri`, for instance via example.multisite.yml.

PR #1102, opened 2017-02-15
Adding support for config_split.

grasmash:


phing/tasks/setup.xml
```diff
@@ -245,21 +246,49 @@
...
+
+  <target name="setup:config-import:core-only">
+    <drush command="config-import" assume="yes" alias="${drush.alias}" passthru="false">
+      <param>${cm.core.config-dir.key}</param>
+    </drush>
+  </target>
+
+  <target name="setup:config-import:config-split">
+    <drush command="en" assume="yes" alias="${drush.alias}" passthru="false">
+      <param>config-split</param>
```
geerlingguy: This command doesn't seem like it would work...?

grasmash: Oh yeah, need an underscore rather than a hyphen.

geerlingguy:
See also: https://github.com/acquia/blt/issues/1138

danepowell:
I'm working to clean up documentation for this here, I'll open a PR when it's ready:
https://github.com/acquia/blt/blob/danepowell-patch-3/readme/configuration-management.md

Note that this replaces the features-workflow doc.

phing/tasks/setup.xml
```diff
@@ -245,21 +246,50 @@
...
+      <param>config_split</param>
+    </drush>
     <if>
-      <equals arg1="${cm.core.partial}" arg2="true"/>
+      <isset property="env.AH_SITE_ENVIRONMENT"/>
       <then>
-        <drush command="config-import" assume="yes" alias="${drush.alias}" passthru="false">
-          <option name="partial"></option>
-          <param>${cm.core.config-dir.key}</param>
+        <mkdir dir="${cm.core.path}/${env.AH_SITE_ENVIRONMENT}"/>
```
danepowell: Can you do this in a Cloud environment? Wouldn't the filesystem at `${cm.core.path}` be read-only?

danepowell:
We should bump the Drush version to `^8.1.10` since that's the minimum supported version with config split: https://github.com/acquia/blt/blob/8.x/template/composer.json#L21

danepowell:
This looks good to me except for my comments above.

grasmash:
This PR is currently failing because we need some example config_split config to use during CI testing.

phing/tasks/setup.xml
grasmash: Yeah I guess we can't do it in a cloud env. Really, the user should be creating this directory manually. Not sure what happens when this doesn't exist. This may be problematic for ODEs...

geerlingguy: > perhaps we should just skip import of the dir doesn't exist?

I like that idea.

phing/tasks/setup.xml
```diff
@@ -245,21 +246,50 @@
...
+        <drush command="config-split-import" assume="yes" alias="${drush.alias}">
+          <!-- Cloud environments require vcs parameter. -->
+          <param>vcs</param>
+          <option name="split">${environment}</option>
```
geerlingguy: Note—the 'split' is now a `param` and not an `option`—so it should be:

```
<param>${environment}</param>
```

phing/tasks/setup.xml
```diff
@@ -245,21 +246,50 @@
...
         </drush>
       </then>
       <else>
-        <drush command="config-import" assume="yes" alias="${drush.alias}" passthru="false">
-          <param>${cm.core.config-dir.key}</param>
+        <drush command="config-split-import" assume="yes" alias="${drush.alias}">
+          <option name="split">${environment}</option>
```
geerlingguy: Note—the 'split' is now a `param` and not an `option`—so it should be:

```
<param>${environment}</param>
```

geerlingguy:
@grasmash - Note also, it looks like beta4 might have a bug where it doesn't combine the split with `config/default` correctly, either. See [Split no longer merges from default.](https://www.drupal.org/node/2861266)

arknoll:
@geerlingguy @grasmash See a PR where I got the new version of config_split/config_filter working with Canary. Specifically:

https://github.com/acquia-pso/canary/pull/25/files#diff-e56cbd5daab22087fd0e4849dc38055dR14

https://github.com/acquia-pso/canary/pull/25/files#diff-2737999ec8b2368b193b40b46c5679c2R758

phing/tasks/setup.xml
arknoll: You actually shouldn't even put ${environment} as an option. The environment needs to be specified in settings.php
See
@geerlingguy @grasmash See a PR where I got the new version of config_split/config_filter working with Canary. Specifically:

https://github.com/acquia-pso/canary/pull/25/files#diff-e56cbd5daab22087fd0e4849dc38055dR14

https://github.com/acquia-pso/canary/pull/25/files#diff-2737999ec8b2368b193b40b46c5679c2R758

danepowell:
Yeah it looks like it's not really a bug in beta4, just a change in how config_split works. It now relies on overriding the configuration via settings.php (like Alex did for Canary) rather than passing the command line argument.

phing/tasks/setup.xml
grasmash: @arknoll Does it have to be done that way? I'd much rather use a param than write to settings.php.

arknoll: It does appear that you have to use the settings.php approach.

Based on the maintainers statement it appears it was by design:
"I think it is better to simplify the options of the command so that it is easier for everybody to use and understand. In my opinion and my experience it is better and safer to define which splits should be imported and exported in settings.php rather than scripts." -- https://www.drupal.org/node/2861266#comment-11990329

settings/config.settings.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+// Ensure the appropriate config split is enabled.
+$config['config_split.config_split.local']['status'] = FALSE;
+$config['config_split.config_split.dev']['status'] = FALSE;
+$config['config_split.config_split.stage']['status'] = FALSE;
+$config['config_split.config_split.prod']['status'] = FALSE;
+$config['config_split.config_split.ci']['status'] = FALSE;
```
geerlingguy: @grasmash - There are no extra settings for RA (many subscriptions have an `ra` environment as well). Do you think that might be something to add? Or just note that people should add an ra-specific split in their own settings.php if they need it?

One other note—for my sites I'm currently using a `test` split (not `stage`) to match the fact that Acquia Cloud still uses `test` as the machine name for the `Stage` environment... for those projects who are already set up using this kind of naming convention, I'll try to document how to switch the split name correctly. Just wanted to put that out there.

danepowell: I'm also using `test` (mainly because that's what canary does... I haven't tested to see how it works yet).

Besides RA, a project could have any number of custom environments (typically `dev1`, `dev2`, etc... but they could be named anything). I think it would be awesome if we could create this configuration automatically, or as part of a wizard (i.e. "here are the typical environments, do you have any others?"). Short of that, just documenting it would be good.

geerlingguy: @danepowell - Also, ODE could throw a wrinkle depending on how you structure your config... it might be nice to say "any non-prod environment gets split `xyz`"...

geerlingguy: Also see my notes-in-progress here: https://github.com/acquia/blt/issues/965#issuecomment-288171604

PR #1108, opened 2017-02-15
Add post-config-import hook.

bobbygryzynger:
Changes proposed:
- Adds a post configuration import hook.

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -288,6 +288,9 @@
     </if>
     <!-- Rebuild caches. -->
     <drush command="cr" alias="${drush.alias}" passthru="false"/>
+    <phingcall target="target-hook:invoke">
+      <property name="hook-name" value="post-config-import"/>
+    </phingcall>
   </target>
 
   <target name="setup:git-hooks" description="Installs git hooks to local .git/hooks directory from version controlled scripts/git-hooks directory.">
--- a/template/blt/project.yml
+++ b/template/blt/project.yml
@@ -50,6 +50,10 @@ target-hooks:
   pre-config-import:
     dir: ${docroot}
     command: echo 'No pre-config-import configured.'
+  # Executed after configuration is imported.
+  post-config-import:
+    dir: ${docroot}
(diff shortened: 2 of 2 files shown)


danepowell:
Makes sense. I'm curious how it would be used though.

bobbygryzynger:
@danepowell - my team would like to use it to toggle local modules.

For instance, when developing sites we use to `devel` (who doesn't). 

When we run 
```sh
blt setup:toggle-modules -Dmultisite.name=<site> -Denvironment=local
``` 
this enables `devel` for us, but then if we run 
```sh
setup:config-import -Dmultisite.name=<site>
``` 
this will fail because our committed code doesn't list `devel` as enabled in `core.extension.yml`.

If we had a `post-config-import` hook we could attach the module toggling to setup:config-import allowing us a bit more consistency in out internal PRs ensuring our devs only toggle development modules after production configuration has been imported and save us all a few key strokes. 

Has your team run into this issue? I'd be interested to hear how you've been solving it if so.

danepowell:
There seem to be two general approaches to CM right now.

The first is to use Features or partial config imports, which implies _not_ using `core.extensions.yml` and instead leaving the enabling / disabling of extensions to update hooks and `blt setup:toggle-modules`. I personally don't use `core.extensions.yml` because back in the day there were some pretty nasty bugs related to it (installing corrupt or non-existent table schemas, etc), and I'm not 100% confident they've been resolved.

The second approach is to do full config imports, but use config_split to support varying configuration (including installed modules) by environment. As an example, see @geerlingguy's [blog post](https://www.jeffgeerling.com/blog/2017/adding-configuration-split-drupal-site-using-blt-and-acquia-cloud) (especially the first user comment)

bobbygryzynger:
Thanks @danepowell, these alternative approaches are definitely something our teams needs to evaluate.

grasmash:
Yeah I'd also like to see a strong use case presented for this. I'm wary of introducing too many new target-hooks.

bobbygryzynger:
@grasmash I agree, feel free to close this, really what is needed is a solution to #1113.

PR #1117, opened 2017-02-17
Fixes #1113: Add support for multisite sync.

### LINKED ISSUE #1113: Local synchronization for multisites
One hole that still remains in BLT's support of multisite development are the local synchronization tasks. 

Currently, they can only be run easily against the `default` site, even if run as, for instance, 
```sh
blt local:sync -Dmultisite.name=not_default
``` 

The biggest stumbling block seems to be this section of `local-sync.xml`:

```xml
<drush command="sql-sync" alias="">
    <option name="structure-tables-key">lightweight</option>
    <option name="create-db"/>
    <option name="sanitize"/>
    <param>@${drush.aliases.remote}</param>
    <param>@${drush.aliases.local}</param>
</drush>
```

At present, there is no way to configure the aliases without overriding the `drush.aliases` variables directly. E.g, 
```sh
blt local:sync -Dmultisite.name=not_default -Ddrush.aliases.remote=notdefault.remote -Ddrush.aliases.local=notdefault.local
```

While this works, it is cumbersome and error prone. 

I think it would be better if BLT could facilitate defining different aliases for non-default sites (with the fallback being that everything is run against the default site).

#827 attempted to do some of this, but never got to the point that it could be incorporated. I'd be interested in hearing from others (paging @danepowell) who are using BLT in a multisite context. What would you see as a good solution to this issue? @grasmash suggested site-by-site files for defining values (which I'm not opposed to), but if all that is required are aliases, perhaps this isn't necessary. 

I'd love to hear some thoughts as perhaps there's a solution to this issue that I'm overlooking.

bobbygryzynger:
Fixes #1113.

Changes proposed:
- Allows site-specific config to be placed in `docroot/sites/*/site.yml`
  - Allows `local:sync` and `local:refresh` tasks to run against a multisite:
```sh
blt local:refresh -Dmultisite.name=secondsite
```

### DIFF
--- a/phing/tasks/local-sync.xml
+++ b/phing/tasks/local-sync.xml
@@ -17,26 +17,6 @@
 
   <target name="local:sync" description="Synchronize local environment from remote (remote --> local)."
           depends="setup:drupal:settings">
-    <if>
-      <!-- There is multisite config. Allow it to override values for local:sync:site. -->
-      <available file="${blt.config-files.multisite}" />
-      <then>
-        <phingcall target="local:sync:site">
-          <property file="${blt.config-files.multisite}" logoutput="false"/>
-          <!--
-            We need to override drush.uri, which defaults to 'default'
-            in order to run commands against ${multisite.name}.
-          -->
-          <property name="drush.uri" value="${multisite.name}"/>
-        </phingcall>
-      </then>
-      <else>
-        <phingcall target="local:sync:site"/>
-      </else>
-    </if>
-  </target>
-
-  <target name="local:sync:site" hidden="true">
     <drush command="cc drush"/>
     <drush command="sql-drop"/>
     <if>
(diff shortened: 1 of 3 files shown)


phing/tasks/local-sync.xml
```diff
@@ -17,6 +17,26 @@
 
   <target name="local:sync" description="Synchronize local environment from remote (remote --> local)."
           depends="setup:drupal:settings">
+    <if>
+      <!-- There is multisite config. Allow it to override values for local:sync:site. -->
+      <available file="${blt.config-files.multisite}" />
```
grasmash: Rather than doing this only in the context of local:sync, it seems like we should just check for this file and load its properties for every blt call. We'd add this to property.xml.

bobbygryzynger:
Ah, yeah that would be better. I'll make that change.

danepowell:
Could you please document this in https://github.com/acquia/blt/blob/8.x/readme/multisite.md ?

bobbygryzynger:
@danepowell - not problem, take a look at #1124.

PR #1119, opened 2017-02-21
Fixes #1087: Auto-discovery of multisite.name.

### LINKED ISSUE #1087: Auto-discovery of multisite.name
Currently, multisite projects must manually override the `multisite.name` array to add the name of each multisite: https://github.com/acquia/blt/blob/8.x/phing/build.yml#L87

It seems like it would be pretty easy to auto-discover sites in the `sites` directory and build this array automatically, making for an easier and less error-prone setup process.

bobbygryzynger:
Fixes #1087.

Changes proposed:
- Create `multisite.name` list automatically if not provided by user.

### DIFF
--- a/tests/phpunit/phpunit.xml
+++ b/tests/phpunit/phpunit.xml
@@ -5,6 +5,7 @@
     <php>
         <const name="BLT_ENV" value="ci"/>
         <const name="BLT_ALIAS" value="self"/>
+        <const name="BLT_MULTISITE_NAME" value="default,g"/>
     </php>
 
 </phpunit>
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ script:
   - source ${BLT_DIR}/scripts/travis/run_tests
   - source ${BLT_DIR}/scripts/blt/ci/internal/doctor.sh
   - phpunit ${BLT_DIR}/tests/phpunit --group blt-project -c ${BLT_DIR}/tests/phpunit/phpunit.xml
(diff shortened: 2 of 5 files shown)


bobbygryzynger:
I'm getting some slightly unexpected output [here](https://travis-ci.org/acquia/blt/builds/203681887#L2060). It looks like the ACSF task adds an additional site. Should the expected value then be `default,g`?

grasmash:
@bobbygryzynger yeah that's right, acsf does add a "g' site.

bobbygryzynger:
No problem, this is passing now [on travis](https://travis-ci.org/acquia/blt/builds/203920388#L2050) with `default,g` as the expected value.

tests/phpunit/phpunit.xml
```diff
@@ -5,6 +5,7 @@
     <php>
         <const name="BLT_ENV" value="ci"/>
         <const name="BLT_ALIAS" value="self"/>
+        <const name="BLT_MULTISITE_NAME" value="default,g"/>
```
grasmash: Why is this defined as a constant in the phpunit.xml rather than just hardcoded in the test?

bobbygryzynger: I thought it would be better, should this value change, that the test itself wouldn't need to be updated. Also, implementing projects would have an easier time overriding this should they wish to run the tests against their own configuration. 

If you'd rather this be hardcoded, I'm fine with making that change, just let me know.

danepowell:
Woot!

PR #1132, opened 2017-02-23
Allow project.prefix to use overridden value

bobbygryzynger:
Changes proposed:
- Allow the commit message hook to get `project.prefix` from BLT so that an overridden value may be used (such as one in `project.local.yml`).

### DIFF
--- a/scripts/git-hooks/commit-msg
+++ b/scripts/git-hooks/commit-msg
@@ -6,7 +6,7 @@
 # 15 characters.
 
 ROOT_DIR="$(pwd)"
-COMMAND="$ROOT_DIR/vendor/bin/yaml-cli get:value $ROOT_DIR/blt/project.yml project.prefix"
+COMMAND="$ROOT_DIR/vendor/bin/blt echo-property -Dproperty.name=project.prefix -silent -emacs"
 ERROR=$($COMMAND 2>&1 > /dev/null)
 PREFIX=$($COMMAND)
 RED='\033[0;31m'


grasmash:
This makes sense to me, but it further slows down an already very slow set of git hooks. I wonder if we could use a faster approach than running a BLT command?

bobbygryzynger:
We could run `yaml-cli get:value $ROOT_DIR/blt/project.yml project.prefix` and then check if `$ROOT_DIR/blt/project.local.yml` exists and has a value for `project.prefix`. 

I don't know that this ends up being substantially faster and I don't like that it essentially duplicates BLT's internal overriding process.

Did you have any other ideas?

bobbygryzynger:
@grasmash - another approach I've thought about here is making `project.prefix` an array. 

The reason I'd like to be able to use an overridden value or multiple values is that in my multisite context, there are multiple projects, each with their own project project prefix. I'm finding myself forced to update my `project.yml` in order to commit changes and then revert this change to `project.yml` afterwards (and hope I don't forget and accidentally commit the change later on). 

Ideally, `project.prefix` could be overridden in `sites/*/site.yml` but I could settle for multiple values being passed to the regular expression in `commit-msg`. 

Would you be open to this change rather than going through BLT itself?

grasmash:
@bobbygryzynger Can't you just use `site.yml` right now to override project.prefix on per-site basis?

grasmash:
If not, yes I'd prefer this approach:

> We could run yaml-cli get:value $ROOT_DIR/blt/project.yml project.prefix and then check if $ROOT_DIR/blt/project.local.yml exists and has a value for project.prefix.

Even though it's duplicative. It should be much faster.

bobbygryzynger:
@grasmash unfortunately no, you can't use `site.yml` because those overrides are loaded by BLT. 

Any solution allowing this to be used would require retrieving `project.prefix` from BLT. But even that is not entirely viable because during a commit you wouldn't be able to specify what site the commits are for in order to load the correct overrides. 

For now, just checking values within `project.local.yml` is the simplest solution. I'll update the PR.

bobbygryzynger:
Closing see #1259.

PR #1133, opened 2017-02-23
Fixes #1128: Added vm:bb:init command.

### LINKED ISSUE #1128: Add beetbox support
Wanted to start an issue before any PR but would you consider adding  [`beetbox`](https://packagist.org/packages/beet/box) as a secondary option for VM.

This could be added as a option `vm:bb:init` where DVM is the default vm.

The project was originally started to replicate an Acquia environment and is currently pinned to PHP 5.6 / ubuntu 14.04.

It makes use of the same ansible roles as DVM however is all pre-provisioned so the build/rebuild time is greatly reduced.

Plus leveraging linked clones the per project size is much smaller as each instance only has a differential disk of the base box.

We can also manage a blt profile from within the project by adding a profile - for example https://github.com/thom8/beetbox/blob/blt_profile/provisioning/ansible/config/profiles/acquia/blt.config.yml

thom8:
PR for https://github.com/acquia/blt/issues/1128

Adds `vm:bb:init` command to initialise beetbox.
Config is using `acquia/blt` profile.

### DIFF
--- a/phing/tasks/vm.xml
+++ b/phing/tasks/vm.xml
@@ -6,7 +6,7 @@
     <if>
       <not><available file="${repo.root}/Vagrantfile"/></not>
       <then>
-        <phingcall target="vm:project:init"/>
+        <phingcall target="vm:dvm:init"/>
       </then>
     </if>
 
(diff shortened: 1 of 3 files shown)


thom8:
No info on CI fail -- `failed_by_user`

phing/tasks/vm.xml
```diff
@@ -85,6 +85,30 @@
     <echo>vagrant up</echo>
   </target>
 
+  <target name="vm:bb:init" description="Initializes beetbox for this project." hidden="true">
```
grasmash: We need to add a substantial amount of documentation regarding this support.

grasmash:
Closing this and #1128 due to inactivity. I'd be happy to merge the feature, but it will require some documentation and automated testing.

thom8:
@grasmash probably easier to integrate once the `vm` command has been abstracted to support multiple projects.

thom8:
Anyway, anyone can run -- `composer require --dev beet/box` to have the project available :)

thom8:
This + robo and there's not much point in a hacky project like BLT

PR #1139, opened 2017-02-27
Fix misleading "Your composer.json file was modified by BLT" error message

TravisCarden:
If BLT modifies your `composer.json` file during an upgrade, it currently emits the following message:

> Your composer.json file was modified by BLT, you MUST run "composer update" to update your composer.lock file.

This is misleading, because `composer update` does not update your `composer.lock` file--it _upgrades all_ your dependencies, which would completely negate the effect of upgrading with `composer update acquia/blt --with-dependencies`. What is actually wanted is `composer update --lock`, which _does_ just update your `composer.lock` file.

### DIFF
--- a/src/Composer/Plugin.php
+++ b/src/Composer/Plugin.php
@@ -143,7 +143,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface {
       $post_composer_json = md5_file($this->getRepoRoot() . DIRECTORY_SEPARATOR . 'composer.json');
 
       if ($pre_composer_json != $post_composer_json) {
-        $this->io->write('<error>Your composer.json file was modified by BLT, you MUST run "composer update" to update your composer.lock file.</error>');
+        $this->io->write('<error>Your composer.json file was modified by BLT, you MUST run "composer update" to re-process and update dependencies.</error>');
       }
     }
     else {


grasmash:
@TravisCarden Sometimes, running `composer update --lock` will not work. 

For instance:
* You run `composer update acquia/blt --with-dependencies`
* BLT adds a new package to your `require-dev` array
* You run `composer update --lock`
* The new package is not installed

Here's another problematic scenario that isn't even BLT specific:
* You run `composer update acquia/blt --with-dependencies`
* The new version has a conflict with another dependency of yours
* Composer _doesn't_ update BLT, and also doesn't throw an error.
* BLT is not updated

Unfortunately, the only foolproof way to ensure that BLT actually installs all of the intended packages is to run `composer require acquia/blt:^[latest-release] --no-update && composer update`.

TravisCarden:
Okay, @grasmash. Then why don't we just update the message to reflect the real intent, so it doesn't look like a mistake?

> Your composer.json file was modified by BLT, you MUST run "composer update" to re-process and update dependencies.

PR updated.

grasmash:
@TravisCarden yeah we can do that, but I also think a docs change is in order.

danepowell:
I might prefer that we recommend using plain old `composer update` at all times (never specifically whitelisting `acquia/blt` or using `--with-dependencies`), and then point out that if this didn't update BLT it's likely because of conflicting dependencies. Just for consistency, and so folks only have to learn a single command.

Obviously we also need to caveat that this will update any Drupal modules or other packages at the same time. My own practice is to update any individual packages _before_ updating BLT, so that I can at least track the changes independently.

TravisCarden:
I definitely wondered what the point was of updating BLT only just so we could turn around and update everything. An alternative line of questioning is to ask why we would modify the composer.json manually in the first place (thus necessitating the second `composer update`) rather than using the CLI to add new dependencies (which would perform the necessary re-parsing automatically). Then the lock file would never get out of sync in the first place.

grasmash:
@TravisCarden I don't think I understand the approach you're proposing.

> An alternative line of questioning is to ask why we would modify the composer.json manually in the first place (thus necessitating the second composer update) rather than using the CLI to add new dependencies (which would perform the necessary re-parsing automatically)

How would BLT use the CLI to add new dependencies? Are you suggesting that BLT not ship any dev dependencies, and instead instruct users to manually add them on their own? Or, are you suggesting that BLT executes shell commands of its own during a `composer update` ?

grasmash:
I'm going to spitball a few different approaches: 

1. When "update" is called on the `acquia\blt` package, prompt the user "BLT has updates for your composer.json file. To apply these updates, Composer must re-solve your dependency tree. Do you want to apply these automatically?" If yes, then BLT modifies composer.json and then executes `composer update` afterwards. If `--no-interaction` is used, BLT relies on value of `extras.blt.update` to answer this question.
1. Modify BLT's Composer plugin logic such, rather than the current "modify composer.json and require a manual `composer update`" approach, it dynamically requires and updates various packages during the initial `composer update` call. This might be really hard to do.
1. Remove the automated updates to composer.json and create a new `blt composer:json:update` command that must be run manually. Add a message to `composer update` output indicating when this command needs to be run.

grasmash:
Eh, I opened #1140 to address this. Seems like a better approach than my other ideas.

danepowell:
I'm not too picky about how BLT updates `composer.json`, my only concern is that we consistently tell folks to use `composer update` (without any arguments), since the more "clever" approaches that we used to recommend (`composer update --with-dependencies acquia/blt`) can have inconsistent and utterly baffling results.

PR #1140, opened 2017-02-28
Changing default composer config to use wikimedia/composer-merge-plugin.

grasmash:


TravisCarden:
Oooh... I like this.

grasmash:
Yeah the only thing it doesn't address is merging the `scripts` key. Once https://github.com/wikimedia/composer-merge-plugin/pull/136 is merged that should be taken care of.

grasmash:
Replaced by #1165

Issue #1141, opened 2017-03-01
Deploy updates not applied to multisites


danepowell:
As part of #1092, I thought we had `deploy:update` running on multisite installs on ACE, but this turned out to be something of a red herring. Using the `@sites` alias was a good start, but because we also specify drush.uri (https://github.com/acquia/blt/blob/8.x/phing/build.yml#L73), updates only  applied to the default site. We need to remove the `drush.uri` property for this to work.

bobbygryzynger:
@danepowell - just throwing this out there. Would it suffice to override `drush.uri` during `update:deploy`:
```xml
<property name="drush.uri" value=""/>
```

danepowell:
Yeah actually it might be that simple. I'll open a PR. Thanks!

grasmash:
Fixed in #1146

PR #1146, opened 2017-03-04
Issue #1141: Fixes deploy updates only applied to default site.

danepowell:
Because build.yml sets the URI to "default", updates only get applied to the default site, even though we are using the `@sites` alias.


grasmash:
Hrm, this causes an update on 'g' for ACSF sites, which fails on ACE.

PR #1151, opened 2017-03-06
Issue #813: Clear twig caches on deployments.

### LINKED ISSUE #813: Clear Twig caches on deployment
We hit a bug after making a change to a Twig template and deploying to production where page requests would randomly pick up the new or old template, depending on which web head served the request. It turns out that `drupal_flush_all_caches` and its ilk (`drush cr`) does not clear the Twig template cache across multiple web servers.

This issue is described here https://docs.acquia.com/article/drupal-8-cache-backend

Until this is fixed in Drupal core, BLT seems like the best place to fix this, maybe via a custom function in a post-code-deploy / post-code-update cloud hook.

danepowell:
Fixes #813 

Changes proposed:
- Add `deploy:id` step to deployments that writes current deploy tag (if present) to a file in the deploy dir.
- Modify settings.php to read this new tag from disk (if it exists) and set `$settings['deployment_identifier']` accordingly.
- Patch core to use `$settings['deployment_identifier']` to handle stale Twig caches on multiple web heads.

I've tested this on a real project and confirmed that the identifier is set correctly, although I haven't been able to test that this actually fixes the problem with stale Twig templates, since that involves a production deploy. I'm fairly confident it will work though.

### DIFF
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -125,7 +125,7 @@
   </target>
 
   <target name="deploy:build" description="Generates a deploy-ready build in deploy.dir."
-          depends="frontend, setup:hash-salt, deploy:copy, deploy:composer:install, deploy:sanitize">
+          depends="frontend, setup:hash-salt, deploy:copy, deploy:composer:install, deploy:id, deploy:sanitize">
     <!--Allow custom commands to be run before commit.-->
     <phingcall target="target-hook:invoke">
       <property name="hook-name" value="post-deploy-build"/>
@@ -136,6 +136,16 @@
         <then>
           <phingcall target="simplesamlphp:deploy:config"/>
         </then>
+    </if>
+  </target>
+
+  <target name="deploy:id" description="Writes deployment id information to artifact.">
+    <if>
+      <isset property="deploy.tag"/>
+      <then>
+        <!-- Write tag name for the benefit of environments like ACE that can't access it. -->
+        <exec dir="${deploy.dir}" command="echo '${deploy.tag}' > deploy_id.txt" logoutput="true" checkreturn="true" level="${blt.exec_level}"/>
+      </then>
     </if>
   </target>
 
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -144,7 +144,7 @@
       <isset property="deploy.tag"/>
       <then>
         <!-- Write tag name for the benefit of environments like ACE that can't access it. -->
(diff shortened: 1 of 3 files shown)


settings/blt.settings.php
```diff
@@ -80,6 +80,17 @@
...
 
+/**
+ * Deployment identifier.
+ *
+ * Drupal's dependency injection container will be automatically invalidated and
+ * rebuilt when the Drupal core version changes. When updating contributed or
+ * custom code that changes the container, changing this identifier will also
+ * allow the container to be invalidated as soon as code is deployed.
+ */
+$deploy_id = file_get_contents(DRUPAL_ROOT . '/../deploy_id.txt');
```
grasmash: I think we should use a file_exists() check here.

phing/tasks/deploy.xml
```diff
@@ -139,6 +139,16 @@
     </if>
   </target>
 
+  <target name="deploy:id" description="Writes deployment id information to artifact.">
+    <if>
+      <isset property="deploy.tag"/>
+      <then>
+        <!-- Write tag name for the benefit of environments like ACE that can't access it. -->
+        <exec dir="${deploy.dir}" command="echo '${deploy.tag}' > deploy_id.txt" logoutput="true" checkreturn="true" level="${blt.exec_level}"/>
```
grasmash: Why not use the commit hash so that this works for non-tags as well? E.g., `git rev-parse HEAD` rather than `echo '${deploy.tag}'`.

grasmash: I suggest that we name this file `deployment_identifier` rather than `deploy_id.txt` for consistency.

phing/tasks/deploy.xml
```diff
@@ -139,6 +139,16 @@
     </if>
   </target>
 
+  <target name="deploy:id" description="Writes deployment id information to artifact.">
+    <if>
+      <isset property="deploy.tag"/>
```
grasmash: Why only do this for tags? I seems like we'd want it for all build artifacts.

danepowell:
I renamed the file to deployment_identifier and added a file_exists to catch notices.

The only time you'd need to set this deployment identifier is for production deploys, since dev/stage don't have multiple web heads. And you'd only ever deploy tags to production. So this seemed like a simpler solution than commit hashes. I also thought commit hashes would be weird because the deployment_identifier would be the _previous_ commit hash; it wouldn't match the current commit because writing the identifier obviously happens prior to the commit. I don't see why that would be bad, it's just weird.

Let me know if you'd rather use the commit hash, that's fine with me.

grasmash:
Makes sense.

PR #1156, opened 2017-03-08
Ignore platform requirements when running composer for a build artifact.

bkosborne:
When creating a build artifact, I don't think composer needs to consider platform requirements, such as PHP versions or PHP extensions being available. It's most likely that the server creating the build artifact is not the one running the code.

### DIFF
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -162,7 +162,7 @@
             <include name="composer.lock"/>
           </fileset>
         </copy>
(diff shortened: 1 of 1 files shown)


grasmash:
I'm a little nervous about committing this. I concede that often, requirements are not actually being built on the same server that will be hosting the application. 

However, adding this flag will delay the discovery of a missing extension or incorrect PHP version until the application is deployed. It seems worthwhile to enforce the presence of the same platform requirements on the build server.

If you really need this, then I'd prefer to introduce a mechanism for customizing the flags used for this command.

bkosborne:
Well I imagine anyone using BLT is also using a dev/stage environment where they test their builds before deploying them, but I see your point.

I'll see if I can adjust the PR to make this configurable. For now I am just overriding the task.

grasmash:
Thanks. In that case, I will close this PR.

PR #1157, opened 2017-03-08
Issue #608: Improved patch documentation.

### LINKED ISSUE #608: Automate / allow overrides of .htaccess for SSL and SAML
As part of #602 we added sample patches to add SSL and SAML support to .htaccess. I thought these worked at the time, but now it seems like they aren't working. I suspect it's because drupal-scaffold doesn't support patching these files: https://github.com/drupal-composer/drupal-scaffold/issues/45

I think we should remove the sample patches, and somehow work with drupal-scaffold to patch .htaccess and keep it from being clobbered on updates.

danepowell:
Fixes #608 

This also removes the sample .htaccess patches because they caused endless confusion, with everyone expecting they could be applied via composer.


mike-potter:
Yes, this is a good improvement to the docs, thanks!

seanhamlin:
This is 100% clearer, thanks!

grasmash:
Well done!

PR #1158, opened 2017-03-08
Fixes #1152: Allow for screenshots in Behat test runs.

### LINKED ISSUE #1152: Add ability to save screenshot and/or dump of html for failed behat tests.
This would help to understand why a test is failing as behat may be seeing things _slightly_ different compared to your browser.

See:
https://github.com/Behat/Behat/issues/649
https://blogs.library.ucsf.edu/ckm/2014/05/14/see-your-failures-taking-screenshots-with-phantomjs-running-behat-tests/

geerlingguy:
Fixes #1152.

### DIFF
--- a/scripts/drupal-vm/config.yml
+++ b/scripts/drupal-vm/config.yml
@@ -58,6 +58,8 @@ installed_extras:
 php_version: "5.6"
 php_packages_extra:
   - "php{{ php_version }}-bz2"
+  - "php{{ php_version }}-imagick"
+  - imagemagick
 
 # XDebug configuration.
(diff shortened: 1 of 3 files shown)


geerlingguy:
Note: Imagemagick is only required if you're switching the `screenshot_taking_mode` to something like `failed_scenarios` (which uses `imagick` to concatenate multiple screenshots). At it's default value it will just dump the image without needing a special PHP image toolkit.

So technically we could just ship with the first commit and not add Imagemagick to Drupal VM. But having the ability to get screenshots for _all_ steps in a failed scenario is actually very useful, as I just discovered (the impetus for this entire PR).

PR #1185, opened 2017-03-14
Add post-config-import hook

rjgwiz:
Changes proposed:
- Add a target hook invoke at the end of `setup:config-import` to complement the existing `pre-config-import` hook.


grasmash:
What is your use case for this feature?

rjgwiz:
Ability to add local only commands in `project.local.yml` under `target-hooks` to run after module enable & config import/fra, such as `drush user-password` to reset U1 pw

PR #1209, opened 2017-03-19
Add Cloud hooks for ACSF 

### LINKED ISSUE #853: Cloud hooks for ACSF
I want Cloud hooks for ACSF

lcatlett:
Fixes #853 

Changes proposed:
- Provides Cloud Hooks support on ACSF for BLT deploy and setup tasks by looping through available site uris on the factory. 
- Enables cloud hooks on ACSF by default for post-code-update and post-code-deploy hooks. 
- Allows for managing ACSF environment, cm, and deployment config via project.yml
- Adds a drush utility which is already in use on many ACSF projects. It provides commands to get information about Factory sites and enables running drush commands against all sites on a Factory.
- The acsf-tools-ml command requires the user to confirm with 'ok' as it assumes that you are SSHed into the Factory and manually running drush commands which is discouraged since those changes are not tracked in VCS and likely not tested. 

To give some background, the method that cloud hooks functionality has previously been enabled on ACSF projects is via a modified version of the acsf-tools-ml command which removed the 'ok' confirmation requirement, but this only allowed for running single drush commands against all sites and there was no integration with the BLT deploy and setup tasks. This required manually re-writing the deploy and setup tasks into a separate script and there was no support for tasks managed via project.yml such as toggle-modules, etc. This PR adds full support for BLT deploys to ACSF to mirror the ACE deployment process. It is enabled by default for new projects and can be added to existing projects by modifying the ACSF exclusion in the post-code-deploy and post-code-update hooks. In the future, it may make sense to exclude running the deploy updates on the 01devup, 01testup, and 01update environments in the post-code-update hook, as site databases are only synced to these environments on Factory code deploys.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ The new `blt-project` directory will have a composer dependency on your local cl
 
 While Phing and the Symfony Console can both accomplish some of the same tasks, they are different tools with different intended purposes. When developing functionality for BLT we are careful to choose the right tool for the right job.
 
-Phing is intended to be a build tool. It is particularly good at stringing together multiple commands and tasks into a single target which can then be executed procedurally. We use Phing when are requirements are well suited to this strength.
+Phing is intended to be a build tool. It is particularly good at stringing together multiple commands and tasks into a single target which can then be executed procedurally. We use Phing when our requirements are well suited to this strength.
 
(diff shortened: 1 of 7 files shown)


grasmash:
@lcatlett wow this looks great! Have you tested this on a real site?

lcatlett:
@grasmash I have only tested this for one project but would be happy to help out with testing on other projects

PR #1260, opened 2017-03-27
Fixes #1232: provide common variables to hook scripts

### LINKED ISSUE #1232: Scripts called from BLT's hooks don't have access to internal variables
When calling a script via BLT's hooks, scripts don't have access to BLT's internal variables. While this is something that currently works as designed, I think it would be a useful feature if target hooks within BLT were supplied with relevant variables - possibly via sourcing them prior to calling the target script. 

For instance, `pre-config-import` could have a variable `site_name` sourced before execution of the target script and have this variable set to `${multisite.name}` locally and `${deploy.site.name}` remotely (or at least during deployments).

For this feature to work well, each target hook would need a well-documented set of variables that are available during each hook and what one should expect their values to represent.

bobbygryzynger:
Fixes #1232.

Mostly just a proof of concept here, but something along these lines would allow a common set of variables to be passed to all of BLT's hook invocations and would allow specific variables to be declared and passed during specific hook invocations.

Changes proposed:
- Create an interface for common internal variables to be passed to hook scripts automatically
- Alter hook variable values based on what environment BLT is run in (when necessary)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,7 @@ script:
   - source ${BLT_DIR}/scripts/blt/ci/internal/doctor.sh
   # Execute PHP Unit tests.
   - phpunit ${BLT_DIR}/tests/phpunit --group blt-project -c ${BLT_DIR}/tests/phpunit/phpunit.xml
+  - phpunit ${BLT_DIR}/tests/phpunit --group blt-hooks -c ${BLT_DIR}/tests/phpunit/phpunit.xml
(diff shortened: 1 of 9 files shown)


grasmash:
Frankly, I'm a little hesitant to introduce this new feature for a few reasons:

1) It increases the complexity of the target hook system.
2) I believe that will deprecate target-hooks once we move to Robo. That port will begin soon, for 8.8.*.

bobbygryzynger:
@grasmash - no problem, I agree that the way this can be accomplished with phing is somewhat roundabout. I'll close this and wait to see if post-Robo port it is still difficult to get variables into hooks.

PR #1265, opened 2017-03-28
Fixes #915: Sync drushrc.php with template drushrc.php via update hook.

### LINKED ISSUE #915: .gitignore customizations overridden during BLT upgrade
My system information:
* Operating system version: 10.12.2
* BLT version: 8.6.6

When I run this command:
```
composer update acquia/blt --with-dependencies
```

I get the following:

  - All my additions to `.gitignore` are removed.

  - My additions (e.g. theme-related paths and files) should remain.

I have two .gitignore additions that I need to make sure developers don't accidentally commit theme-related things to our repo:

```
prepros.cfg
docroot/themes/custom/project_theme/css/*
```

Is there a different way we should manage `.gitignore` customizations for the project? Or is it a bug that BLT is overwriting these changes?

malikkotob:


src/Update/Updater.php
```diff
@@ -479,6 +479,30 @@ public function moveFile($source, $target, $overwrite = FALSE) {
...
+   *   FALSE if nothing happened.
+   */
+  public function syncWithTemplate($filePath, $overwrite = FALSE) {
+    $sourcePath = $this->getBltRoot() . '/template/' . $filePath;
+    $targetPath = $this->getRepoRoot() . '/' . $filePath;
+    $fileCopied = FALSE;
+
+    if ($this->getFileSystem()->exists($sourcePath)) {
+      $this->getFileSystem()->copy($sourcePath, $targetPath, $overwrite);
+      if ($overwrite) {
```
grasmash: If the file was copied but not overwritten, won't `$fileCopied` be FALSE?

malikkotob: @grasmash that's true. Good catch. Looking back at the [fileSystem copy method](http://symfony.com/doc/current/components/filesystem.html#copy), I realized we actually don't have a way of knowing if the file actually copied if `$overwrite` is `FALSE` without us manually comparing the last modified date on the target file and the source file. I can either modify the logic to check that information (which feels a bit excessive) or I can simply not return anything - which is what the `fileSystem` copy method is currently doing (doesn't feel great either). Thoughts?

grasmash: Let's not return anything. An exception should be throw if it fails

PR #1267, opened 2017-03-29
Instruct local files sync to exclude image styles as well as the css/js cache folders

bkosborne:
This seems reasonable to me?


grasmash:
I think so...

PR #1278, opened 2017-03-31
Fixes #1276: Ensure there are no uncommitted changes before deploy.

### LINKED ISSUE #1276: Deploy can be run with uncommitted changes
In version `8.6.14` I am able to run `blt deploy` with uncommitted changes and these changes will be included in my deploy. This is a potentially dangerous capability as it could cause a lot of unintentional changes to end up on a remote environment. 

Ideally, BLT would check the working tree and enforce that it is clean before allowing a deploy to proceed.

bobbygryzynger:
Fixes #1276.

Changes proposed:
- Check for a clean git status before deploys


grasmash:
Good call.

PR #1306, opened 2017-04-04
BLT-1304: Fixes #1304 default no cache setting for SAML.

### LINKED ISSUE #1304: SimpleSAMLPHP defines NO_CACHE by default.
SimpleSAMLPHP is by default setting a NO_CACHE=1 cookie.

This is a cookie for Acquia Cloud that allows you to bypass varnish. 

My understanding is that it fixes an issue that lets users, who logged into the same SAML provider on another site, be automatically authenticated on this site.

That works great if your site has 1000's of logged in users, but for a completely anonymous site, with 10-20 content editors, this breaks varnish and causes performance implications.

cc @dooleymatt

kylebrowning:
Fixes #1304 

Changes proposed:
- Remove default disabling of cache.

### DIFF
--- a/scripts/simplesamlphp/acquia_config.php
+++ b/scripts/simplesamlphp/acquia_config.php
@@ -21,8 +21,13 @@ $config['technicalcontact_email'] = "your_email@yourdomain.com";
 // Change these for your installation
 $config['secretsalt'] = 'y0h9d13pki9qdhfm3l5nws4jjn55j6hj';
 $config['auth.adminpassword'] = 'mysupersecret';
-// Prevent Varnish from interfering with SimpleSAMLphp.
-setcookie('NO_CACHE', '1');
+
+// Uncomment this line to allow users to be automatically logged in if they
+// signed in via the same SAML provider on another site.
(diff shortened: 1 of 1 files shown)


scripts/simplesamlphp/acquia_config.php
```diff
@@ -21,8 +21,13 @@
 // Change these for your installation
 $config['secretsalt'] = 'y0h9d13pki9qdhfm3l5nws4jjn55j6hj';
 $config['auth.adminpassword'] = 'mysupersecret';
-// Prevent Varnish from interfering with SimpleSAMLphp.
-setcookie('NO_CACHE', '1');
```
greylabel: @kylebrowning I remember at one point seeing requests to `/saml_login/*` being cached unintentionally in Varnish and causing issue. Have we verified this works as expected? The NO_CACHE cookie could also be scoped to a specific path.

kylebrowning: @greylabel Yes it could! If thats the case, maybe we should do that instead.

PR #1313, opened 2017-04-04
Fix config imports

danepowell:
BLT version 8.7.0-beta3+several patches for more recent commits

setup:config-import:config-split is currently broken because for whatever reason Phing doesn't like the relative path used in cm.core.dirs.sync.path and cm.core.path. Simply changing this to an absolute path fixes it.

### DIFF
--- a/phing/build.yml
+++ b/phing/build.yml
@@ -42,7 +42,7 @@ cm:
   strategy: config-split
   core:
     # The parent directory for configuration directories, relative to the docroot.
-    path: ../config
+    path: ${repo.root}/config
     # The default config key to use for imports. This is the key used in Drupal's global $config_directories variable.
     # E.g., $config_directories['sync']. It must have a corresponding key in cm.core.dirs. E.g., `cm.core.dirs.sync`.
     key: sync


danepowell:
Ugh. This doesn't work because the config-dir argument to drush si only accepts relative paths.

phing/build.yml
```diff
@@ -42,7 +42,7 @@ cm:
   strategy: config-split
   core:
     # The parent directory for configuration directories, relative to the docroot.
-    path: ../config
+    path: ${repo.root}/config
```
grasmash: This was done to ensure that a command run from outside of Drupal VM would work within Drupal VM. Are you sure this doesn't work? @arknoll recently made a related change: https://github.com/acquia/blt/pull/1305/files

Also, this does seem to work in our CI process: https://github.com/acquia/blt/blob/8.x/scripts/blt/ci/internal/run_tests.sh#L30

danepowell: Ah, let me try that. The commit message on that implies that it's just fixing a test, so I didn't bother patching for it :)

danepowell: Yep that worked.

arknoll: @danepowell yeah, in retrospect I could have used a word other than "test"   :)

PR #1316, opened 2017-04-05
Updating config split docs.

grasmash:


danepowell:
Looks good, although we could simplify or even get rid of the setup section now that Config Split is used by default: #1319

PR #1323, opened 2017-04-05
Issue #53: Added support for custom git username and email

### LINKED ISSUE #53: manual build and deploy process fails if git user.useConfigOnly option is set
When git's useConfigOnly option is set and the user.name and user.email is not set globally, running the process to create the build artifact and deploy it as defined in [the deploy instructions](https://github.com/acquia/bolt/blob/8.x/template/readme/deploy.md#create-and-deploy-the-build-artifact) fails with the error below.

> project > deploy:artifact:commit:
> 
> fatal: user.useConfigOnly set but no name given fatal: user.useConfigOnly set but no name given

Setting the user config in the deploy directory does not successfully work around the problem since the deploy directory is wiped out each time as the first step in the deploy:artifact:prepare-dir step.

danepowell:
Fixes #53 

This adds a set of hidden properties for git identification (git.user.name and git.user.email). If these properties are set, the deploy repo will be initialized with them.

You can set these either in `project.yml`, `project.local.yml`, or via the command line (e.g. `blt deploy -Dgit.user.name=Joe`)

I didn't document these because it seems like an edge case, but one that's worth supporting. Maybe it would be good to document them somewhere? I don't know that we have a comprehensive list of Phing properties anywhere...

### DIFF
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -217,6 +217,17 @@
     <exec command="git init ${deploy.dir}" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="false"/>
     <exec dir="${deploy.dir}" command="git config --local core.excludesfile false" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
     <echo>Global .gitignore file is being disabled for this repository to prevent unexpected behavior.</echo>
+    <!-- Add the git username and email if necessary. -->
+    <if>
+      <and>
+        <isset property="git.user.name"/>
+        <isset property="git.user.email"/>
+      </and>
+      <then>
+        <exec dir="${deploy.dir}" command="git config --local --add user.email ${git.user.email} user.name ${git.user.name}" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
+      </then>
+    </if>
+
   </target>
 
   <target name="deploy:push-all" hidden="true">


ba66e77:
@danepowell, apologies for the delay testing this patch out.  I'm getting failures from it.

In my project.local.yml, I added
```
git:
  user:
    name: 'Willy Wootness'
    email: 'WillyWootnessi@example.com'
```
Running the deploy gets the following:
> > $ blt deploy -Ddeploy.branch=master -Ddeploy.commitMsg='WW-123: The commit message.'
> blt > deploy:
> blt > deploy:prepare-dir:
>    [delete] Deleting directory /Users/barrett.smith/Desktop/bswaterwheel/deploy
>      [exec] Initialized empty Git repository in /Users/barrett.smith/Desktop/bswaterwheel/deploy/.git/
>      [echo] Global .gitignore file is being disabled for this repository to prevent unexpected behavior.
> error: wrong number of arguments
> 
> 
> git:
> usage: git config [<options>]
> 
> Config file location
>     --global              use global config file
>     --system              use system config file
>     --local               use repository config file
>     -f, --file <file>     use given config file
>     --blob <blob-id>      read config from given blob object
> 
> Action
>     --get                 get value: name [value-regex]
>     --get-all             get all values: key [value-regex]
>     --get-regexp          get values for regexp: name-regex [value-regex]
>     --get-urlmatch        get value specific for the URL: section[.var] URL
>     --replace-all         replace all matching variables: name value [value_regex]
>     --add                 add a new variable: name value
>     --unset               remove a variable: name [value-regex]
>     --unset-all           remove all matches: name [value-regex]
>     --rename-section      rename section: old-name new-name
>     --remove-section      remove a section: name
>     -l, --list            list all
> 
> 
> git:
>     -e, --edit            open an editor
>     --get-color           find the color configured: slot [default]
>     --get-colorbool       find the color setting: slot [stdout-is-tty]
> 
> Type
>     --bool                value is "true" or "false"
>     --int                 value is decimal number
>     --bool-or-int         value is --bool or --int
>     --path                value is a path (file or directory name)
> 
> Other
>     -z, --null            terminate values with NUL byte
>     --name-only           show variable names only
>     --includes            respect include directives on lookup
>     --show-origin         show origin of config (file, standard input, blob, command line)
> 
> [phingcall] /Users/barrett.smith/Desktop/bswaterwheel/./vendor/acquia/blt/phing/tasks/deploy.xml:259:8: /Users/barrett.smith/Desktop/bswaterwheel/./vendor/acquia/blt/phing/tasks/deploy.xml:265:203: Task exited with code 129
> 
> BUILD FAILED/Users/barrett.smith/Desktop/bswaterwheel/./vendor/acquia/blt/phing/tasks/deploy.xml:101:43: Execution of the target buildfile failed. Aborting.
> ; 2.7168 seconds

What's odd is that if I comment out *either* of the two lines in the project.local.yml, git doesn't complain ( until it gets to the part about not being able to commit because one of the config options isn't set, as I reported in #53).  That is, the error reported here doesn't occur in either of the cases below:

**this works**
```
git:
  user:
    name: 'Willy Wootness'
#    email: 'WillyWootnessi@example.com'
```
**this also works**
```
git:
  user:
#    name: 'Willy Wootness'
    email: 'WillyWootnessi@example.com'
```

danepowell:
try #1402

ba66e77:
That did the trick. Thanks!

PR #1326, opened 2017-04-06
Change suggested packages to suggest from require.

### LINKED ISSUE #1325: "vendor/acquia/blt/composer.suggested.json" causing issues with composer
My system information:
* Operating system type: OSX
* Operating system version: Sierra
* BLT version: 8.7.1

When I am updating lightning like this:
```
composer require acquia/lightning:^2.1.0
rm -rf vendor && composer install
```
I got all the correct updated dependencies.
Then I ran:
```
composer update
```
and it downgraded me to older versions. For instance:
```
Updating drupal/core (8.3.0 => 8.2.7) Loading from cache
```
Then I ran:
```
composer update
```
again. And nothing was updated.
Then I removed from my composer.json:
```
"vendor/acquia/blt/composer.suggested.json"
```
And ran composer update and I got all the correct dependencies again.

Need to figure out why the suggested json is causing this problem.

arknoll:
Fixes #1325.

### DIFF
--- a/composer.suggested.json
+++ b/composer.suggested.json
@@ -1,5 +1,5 @@
 {
-  "require": {
+  "suggest": {
     "acquia/lightning": "^2.0.3",
     "drupal/acquia_connector": "^1.5.0",
     "drupal/cog": "^1.0.0",


arknoll:
I am assuming due to the naming convention of the file that these packages should really just be suggestions rather than require.

arknoll:
Tested. this does fix the issue.

grasmash:
This functionality changes the composer.suggested.json file. The intention is to provide defaults that can be easily removed. This change would no longer include the packages. It would simply output a message indicating that these are suggested. 

I guess the real problem is that this changes the UX such that, to get a fully functioning code base with all suggested packages, the user would now need to run many composer require commands to download the suggested packages after creating the blt-based project.

grasmash:
I'm closing this in favor of the solution in #1325.

PR #1331, opened 2017-04-06
Fixes #1324: Add ah_other config split.

### LINKED ISSUE #1324: What to do if using Config Split with the RA environment?
Since 8.7.0 now includes `config-split` support (which seems to work well in most cases), and our site has an `ra` (Remote Administration) environment, what's the best way to handle an `ra` split?

One option might be to just not create and `ra` split at all... I haven't yet tested this, but it could work.

Another option would be to add to the site's `settings.php` at the bottom:

```php
require DRUPAL_ROOT . "/sites/default/settings/config_split.settings.php";
```

Then add this in that `config_split` settings file:

```php
<?php

/**
 * @file
 * Config Split custom settings for BEA.
 */

// RA Environment settings.
$is_ah_ra_env = ($ah_env == 'ra');
$config['config_split.config_split.ra']['status'] = FALSE;
if ($is_ah_ra_env) {
  $split = 'ra';
  if (file_exists("$split_filepath_prefix.$split.yml")) {
    $config["$split_filename_prefix.$split"]['status'] = TRUE;
  }
}
```

We don't particularly use the RA environment, so I'm wondering if maybe the best practice would just be "don't add a split for RA"?

grasmash:


settings/config.settings.php
```diff
@@ -13,6 +13,7 @@
 $config['config_split.config_split.prod']['status'] = FALSE;
 $config['config_split.config_split.ci']['status'] = FALSE;
```
geerlingguy: I think you also need to add:

```
$config['config_split.config_split.ah_other']['status'] = FALSE;
```

grasmash: yes.

PR #1341, opened 2017-04-07
Split deploy git variables

### LINKED ISSUE #1340: Split git config user email and name and document it
As a follow-up for #1323 I think the configuration is independent between `git.user.name` and `git.user.email`, so it should not be required to have both.

Also, I do not know the right place to document that, but I think it should.

PR will follow.

marvil07:
Fixes #1340.

Changes proposed:
- Split git config overrides on deploy git directory, so both of them are not required.
- Documents git.user.name and git.user.email blt options.

### DIFF
--- a/phing/tasks/deploy.xml
+++ b/phing/tasks/deploy.xml
@@ -236,13 +236,14 @@
     <echo>Global .gitignore file is being disabled for this repository to prevent unexpected behavior.</echo>
     <!-- Add the git username and email if necessary. -->
     <if>
-      <and>
-        <isset property="git.user.name"/>
-        <isset property="git.user.email"/>
-      </and>
-      <then>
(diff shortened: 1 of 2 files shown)


marvil07:
This expands #1323.

phing/tasks/deploy.xml
```diff
@@ -236,13 +236,14 @@
...
-        <isset property="git.user.name"/>
-        <isset property="git.user.email"/>
-      </and>
-      <then>
-        <exec dir="${deploy.dir}" command="git config --local --add user.email ${git.user.email} user.name ${git.user.name}" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
-      </then>
+      <isset property="git.user.name"/>
+    <then>
+      <exec dir="${deploy.dir}" command="git config --local --add user.name ${git.user.name}" logoutput="true" checkreturn="true" level="${blt.exec_level}" passthru="true"/>
+    </if>
```
grasmash: XML tags are not properly closed here.

grasmash:
Closing due to syntax error. Waiting on response in #1340.

PR #1344, opened 2017-04-08
Implements #1343 by adding a new command.

### LINKED ISSUE #1343: Add a command to print out available Behat definitions.
We previously added a way for someone to manually print out the definition list using the `behat.extras` config item. It would be nice if this was bundled into a command so someone can execute it directly.

swichers:
Fixes #1343 .

Changes proposed:
- Add a new `blt tests:behat:definitions` command.


grasmash:
tests.xml has been removed. Can you please re-submit as Robo command?

PR #1346, opened 2017-04-08
Implements #1345 by making delay configurable.

### LINKED ISSUE #1345: Make Selenium and PhantomJS startup delay configurable.
When running Behat tests there is a significant delay (10-15 seconds) while BLT waits for the Selenium or PhantomJS servers to become available. It would be nice if this was configurable for  impatient folks like myself.

swichers:
Fixes #1345  .

Changes proposed:
- Adds configurable delay instead of a static one.

### DIFF
--- a/phing/build.yml
+++ b/phing/build.yml
@@ -11,6 +11,10 @@ behat:
   launch-selenium: true
   # If true, PhantomJS GhostDriver will be launched with Behat.
   launch-phantomjs: false
+  # The delay to wait for the Selenium server to start.
+  wait-selenium: 15
+  # The delay to wait for the PhantomJS GhostDriver to start.
+  wait-phantomjs: 10
   # An array of paths with behat tests that should be executed.
   paths:
     # - ${docroot}/modules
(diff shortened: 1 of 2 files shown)


grasmash:
The tests commands are being moved to https://github.com/acquia/blt/blob/8.x/src/Robo/Commands/Tests/BehatCommand.php. The tests.xml file will be removed before the next release. Could you please submit a PR with a modification to the Robo file?

swichers:
Correct me if I am wrong, but the new implementation doesn't have a static delay. Instead it waits for a URL, no?

```
  $this->getContainer()->get('executor')->waitForUrlAvailable($this->seleniumUrl);
```

If so then that renders this change unnecessary.

grasmash:
Well it does max out at 15 seconds: https://github.com/acquia/blt/blob/5dda56486b8b50a5750876d43d273b02c25b6976/src/Robo/Common/Executor.php#L124

swichers:
Functionally before you had to wait 15 seconds no matter what, with this it has the chance to return much sooner with a cap of 15. That sounds good to me. I don't think this change adds much to the new code, but I can update if you want. Should I continue?

grasmash:
Nah, I'm fine leaving it how it is.

PR #1349, opened 2017-04-10
Fixes #1328: Add support for directories with spaces.

### LINKED ISSUE #1328: Spaces in project paths break BLT
My system information:
* BLT version: 8.7 and earlier

When I run this command:
```
  blt doctor
  <or any blt command that passes or uses raw paths>
```

I get the following output:
```
BUILD FAILED<path with space>/.vendor/acquia/blt/phing/build.xml:25:51: <path with space>/.vendor/acquia/blt/phing/tasks/properties.xml:42:175: Task exited with code 1; 0.2125 seconds

Note: Line number and file may vary depending on command used.
```

There are several places where paths are passed raw to the shell without escaping or quoting, and this causes the command to fail.

malikkotob:


grasmash:
I'd like to suggest that we fix this in Robo rather than Phing.

PR #1353, opened 2017-04-10
Update notifications not disabled during Drupal 8.3.x installation

danepowell:
BLT disables update notifications during install time via a command-line argument to Drush. This is nice because it prevents Drupal from sending an email that can cause problems on systems without a local mail transport agent.

The problem is that Drupal changed the update notification configuration form in 8.3.0: http://cgit.drupalcode.org/drupal/commit/?h=8.3.x&id=b97e6a73ab0593efa99c84a36c635bbe3fe9f4e6

This caused this parameter to stop working, and for some users errors due to their system being unable to send the email.

This fixes it by adding the new parameter (leaving the old parameter for 8.2.x users shouldn't hurt anything).

I also considered making this a Phing property / variable to reduce duplication, let me know if you'd prefer that.

### DIFF
--- a/phing/tasks/drupal.xml
+++ b/phing/tasks/drupal.xml
@@ -24,7 +24,7 @@
                   <option name="account-mail">"${drupal.account.mail}"</option>
                   <option name="locale">"${drupal.locale}"</option>
                   <param>"${project.profile.name}"</param>
-                  <param>"install_configure_form.update_status_module='array(FALSE,FALSE)'"</param>
+                  <param>"install_configure_form.update_status_module='array(FALSE,FALSE)';install_configure_form.enable_update_status_module=FALSE;install_configure_form.enable_update_status_emails=FALSE"</param>
               </drush>
           </then>
           <else>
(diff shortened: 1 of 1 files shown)


geerlingguy:
+1 also started getting the following in Travis CI builds:

```
Unable to send email. Contact the site administrator if the problem  [error]
persists.
```

geerlingguy:
If you need a quick hacky workaround for this problem in the Travis CI environment, stick the 'sendmail_path' line into your `.travis.yml` file as shown below:

```yaml
before_install:
  # Disable xdebug.
  - phpenv config-rm xdebug.ini
  - echo 'sendmail_path = "/bin/true"' | sudo tee "/home/travis/.phpenv/versions/`php -i | grep "PHP Version" | head -n 1 | grep -o -P '\d+\.\d+\.\d+.*'`/etc/conf.d/sendmail.ini"
  - composer self-update
  - composer install
```

It basically tells any PHP process trying to send an email that the send was successful no matter what. Don't ever use this hack on a live site :)

danepowell:
@dreamfony I've tested this and the email sending was definitely disabled. I'm not sure how to reconcile that with your comment. It's possible I changed something else at the same time without realizing it, but I'm pretty sure that's not the case.

danepowell:
@dreamfony thanks a lot for catching this, I think I have the correct fix in #1360

PR #1356, opened 2017-04-10
Fixes #1355: Make Drush alias for 'local' work in VM guest or host by default.

### LINKED ISSUE #1355: Make Drush alias for 'local' work in VM guest or host by default
My system information:
* Operating system version: 10.12.4
* BLT version: 8.7.0

I am noticing more and more developers choosing to operate on the project entirely inside Drupal VM (instead of running things like composer, BLT, and drush commands on their host), and it made me think... why don't we make the `@project.local` Drush alias work outside (the current default) _and_ inside the VM?

I think there's a flag I can use in the `aliases.drushrc.php` file to do this.

geerlingguy:
Fixes #1355.

### DIFF
--- a/scripts/drupal-vm/drupal-vm.aliases.drushrc.php
+++ b/scripts/drupal-vm/drupal-vm.aliases.drushrc.php
@@ -1,11 +1,16 @@
 // [vagrant_machine_name].local
 $aliases['${project.machine_name}.local'] = array(
-  // /var/www/[vagrant_machine_name]/docroot
+  // /var/www/${project.machine_name}/docroot
   'root' => '/var/www/${project.machine_name}/docroot',
   // vagrant_hostname
   'uri' => '${project.local.uri}',
-  // vagrant_hostname
-  'remote-host' => '${project.local.hostname}',
-  'remote-user' => 'vagrant',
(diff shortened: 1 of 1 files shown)


geerlingguy:
Note that this seems to work for me—`get_current_user()` works when I'm running on my Mac/host, but inside Drupal VM it returns empty... but I'm not sure if there's an easier/faster way to determine if we're inside a VM or not.

lcatlett:
@geerlingguy i've used the following for the same: 

```
// [vagrant_machine_name].local
$aliases['${project.machine_name}.local'] = array(
  // /var/www/${project.machine_name}/docroot
  'root' => '/var/www/${project.machine_name}/docroot',
  // vagrant_hostname
  'uri' => '${project.local.uri}',
  );
if ('vagrant' != $_SERVER['USER']) {
  $aliases['${project.machine_name}.local'] += array(
    // vagrant_hostname
    'remote-host' => '${project.local.hostname}',
    'remote-user' => 'vagrant',
    'ssh-options' => '-o PasswordAuthentication=no -i ' . drush_server_home() . '/.vagrant.d/insecure_private_key'
  );
}
```

geerlingguy:
@lcatlett - Ooh, that could actually be better. I'll change my PR tomorrow, unless you want to build off this one or send up a new one before I get to it.

geerlingguy:
@lcatlett / @grasmash - rebased and updated. Looks better to me!

PR #1361, opened 2017-04-11
Porting acsf:* commands to Robo.

malikkotob:


src/Robo/Commands/Acsf/AcsfCommand.php
```diff
@@ -0,0 +1,95 @@
...
+class AcsfCommand extends BltTasks {
+
+  /**
+   * Initializes ACSF support for project.
+   *
+   * @command acsf:init
+   *
+   * @options acsf-package
+   */
+  public function acsfInitialize($options = ['acsf-package' => 'drupal/acsf:^1.33.0']) {
```
grasmash: I'd just make the version constraint the option. The `drupal/acsf` part should never change.

src/Robo/Commands/Acsf/AcsfCommand.php
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Refreshes the ACSF settings and hook files.
+   *
+   * @command acsf:init:drush
+   */
+  public function acsfDrushInitialize() {
+    $drushBin = $this->getConfigValue('drush.bin');
+    $drushAlias = $this->getConfigValue('drush.default_alias');
```
grasmash: I think we want to use `drush.alias` and not `default_alias`. We can make `drush.alias = drush.default_alias` in DefaultConfig. This would allow people to override it at run time via `blt acsf:init -D drush.alias=whatever`.

src/Robo/Commands/Acsf/AcsfCommand.php
```diff
@@ -0,0 +1,95 @@
...
+    $result = $this->taskExec("composer require {$acsfPackage}")
+      ->interactive()
+      ->printOutput(TRUE)
+      ->dir($this->getConfigValue('repo.root'))
+      ->run();
+
+    if (!$result->wasSuccessful()) {
+      $this->logger->error("An error occurred while requiring drupal/acsf.");
+      $this->say("This is likely due to an incompatibility with your existing packages.");
+      $confirm = $this->confirm("Should BLT attempt to update all of your Composer packages in order to find a compatible version?");
```
grasmash: Nice!

src/Robo/Commands/Acsf/AcsfCommand.php
malikkotob: good point - adjusting now.

src/Robo/Commands/Acsf/AcsfCommand.php
malikkotob: Hmm, is there actually a `drush.alias` config option or are you saying we'd be defining a new key value pair in config essentially? I didn't find one in our config files ([`build.yml`](https://github.com/acquia/blt/blob/8.x/phing/build.yml#L84), [`project.yml`](https://github.com/acquia/blt/blob/8.x/template/blt/project.yml#L23), `project.local.yml`). 

Also, I don't know if we can access drush alias info in `DefaultConfig` without some refactoring, since [it's instantiated before](https://github.com/acquia/blt/blob/8.x/bin/blt-robo-run.php#L43) `project.yml` and `project.local.yml` are imported. We could add a setter method to `DefaultConfig` and then invoke it after [this line](https://github.com/acquia/blt/blob/8.x/bin/blt-robo-run.php#L49), but that feels hacky.

src/Robo/Commands/Acsf/AcsfCommand.php
malikkotob: @grasmash the above logic was borrowed from https://github.com/acquia/blt/blob/8.x/src/Robo/Commands/Vm/VmCommand.php#L200.

Seems like something we may want to abstract out so that we can leverage anywhere we're running composer commands.

src/Robo/Commands/Acsf/AcsfCommand.php
grasmash: `drush.alias` is not intended to be a configurable option. Instead, it's a variable that represents the "active" alias. So, we default it to be the same as `drush.default_alias` but various other processes are free to change its value.

PR #1377, opened 2017-04-13
Issue #1375: Return a Robo\Result to set the proper error code.

geerlingguy:
[WIP] - See issue #1375.

PR #1381, opened 2017-04-14
Fixes #1377: Failing behat tests still allow travisci to "pass".

grasmash:


geerlingguy:
This seems to be the proper fix.

```
3 scenarios (2 passed, 1 failed)
12 steps (11 passed, 1 failed)
0m8.54s (82.05Mb)
[Testing\Behat]  Exit code 1 
[error]  Behat tests failed 

vagrant@local:/var/www/canary$ echo $?
1
```

Before, that was a `0`. Applied the patch and now it's a `1`.

PR #1398, opened 2017-04-17
Run config-import twice

danepowell:
It's sometimes necessary to run config-import twice to completely import all configuration changes. This shouldn't be necessary in theory, but I've identified a number of core bugs that can cause this (such as [this](https://github.com/drush-ops/drush/issues/2709), [this](https://www.drupal.org/node/2622160#comment-12038215), and some other related to dependency detection that I haven't been able to pin down yet).

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -229,6 +229,10 @@
         <drush command="config-import" assume="yes" alias="${drush.alias}" passthru="false">
           <param>${cm.core.key}</param>
         </drush>
+        <!-- Sometimes config-import has to be run twice to catch everything. Can be removed once the core configuration system is more stable. -->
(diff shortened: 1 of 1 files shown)


phing/tasks/setup.xml
```diff
@@ -229,6 +229,10 @@
         <drush command="config-import" assume="yes" alias="${drush.alias}" passthru="false">
           <param>${cm.core.key}</param>
         </drush>
+        <!-- Sometimes config-import has to be run twice to catch everything. Can be removed once the core configuration system is more stable. -->
```
bhasselbeck: @danepowell maybe add a URL referencing the bug (in core) that's in question here...

danepowell: I would but I haven't been able to isolate all of the different ways this can happen... I think this workaround will be necessary for a while until the whole config subsystem is generally more mature.

PR #1406, opened 2017-04-18
Remove sudo mysql starts

unn:


thomashowell:
Should we follow the instructions here: https://docs.acquia.com/pipelines/yaml#services

And add the services element to replace the removed sudo command?

Sample: https://github.com/acquia/pipelines-examples/blob/e6c5f74c0a69e7eba3daff272f17307bd690e7db/tutorial-501/acquia-pipelines.yml#L4

PR #1413, opened 2017-04-20
Create new custom Robo Drush task.

malikkotob:


src/Robo/Commands/Acsf/AcsfCommand.php
```diff
@@ -34,13 +34,11 @@ public function acsfInitialize($options = ['acsf-version' => '^1.33.0']) {
...
   public function acsfDrushInitialize() {
-    $drushBin = $this->getConfigValue('drush.bin');
-
     $this->say('Executing initialization command for acsf module.');
 
-    $this->taskExec("{$drushBin} acsf-init --include={$this->getConfigValue('docroot')}/modules/contrib/acsf/acsf_init --yes")
-      ->printOutput(TRUE)
-      ->dir($this->getConfigValue('docroot'))
+    $this->taskDrush()
+      ->option("--include={$this->getConfigValue('docroot')}/modules/contrib/acsf/acsf_init")
```
grasmash: `--include` seems like a good global argument.

grasmash: It feels weird to be to define the option before the command.

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+  private $assume;
+
+  /**
+   * Attach tty to process for interactive input.
+   *
+   * @var bool
+   *
+   * @see ExecTrait::$interactive
+   */
+  private $passthru;
```
grasmash: This is an anachronism leftover from Phing. Why not just let interactive be set directly?

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+
+  /**
+   * Indicates if command output should be printed.
+   *
+   * @var bool
+   *
+   * @see ExecTrait::isPrinted
+   */
+  private $logOutput;
```
grasmash: Similarly, why not set printOutput() directly?

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+    if (!$this->defaultsInitialized) {
+      $this->init();
+    }
+
+    if ($this->alias) {
+      $command = "@{$this->alias} {$command}";
+    }
+
+    if (!empty($this->uri)) {
+      $this->option('-l', $this->uri);
```
grasmash: I'd suggest using full option rather than shortcuts. I.e., `--uri=` rather than `-l`.

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+      $assumption = $this->assume ? 'yes' : 'no';
+      $this->option("--{$assumption}");
+    }
+
+    if ($this->verbosityThreshold() >= VerbosityThresholdInterface::VERBOSITY_VERBOSE) {
+      $this->verbose(TRUE);
+    }
+
+    if ($this->verbose) {
+      $this->option('-v');
```
grasmash: Similarly `--verbose`.

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+
+    // Add in arguments set via option method and clear for next invocation.
+    $command = $command . $this->arguments;
+    $this->arguments = '';
+
+    return $this->exec($command)
+      ->dir($this->dir)
+      ->interactive($this->passthru)
+      ->printOutput($this->logOutput)
+      ->printMetadata(false);
```
grasmash: Can you run this file through PHPCS ?

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+class Drush extends CommandStack
+{
+  use CommandArguments;
+
+  /**
+   * Site alias to prepend to each command.
+   *
+   * @var string
+   */
+  private $alias;
```
grasmash: I prefer to default to protected unless there's a particular reason for private.

src/Robo/Tasks/Drush.php
```diff
@@ -0,0 +1,267 @@
...
+   *
+   * @return $this
+   *
+   * @see ExecTrait::$isPrinted
+   */
+  public function logOutput($logOutput) {
+    if (is_string($logOutput)) {
+      $this->logOutput = ($logOutput === 'yes' || $logOutput === 'true');
+    } else {
+      $this->logOutput = !!$logOutput;
```
grasmash: This is weird.

src/Robo/Tasks/Drush.php
grasmash: I think I'd would not set a default for this.

src/Robo/Commands/Acsf/AcsfCommand.php
malikkotob: Since `include` is reserved for PHP's `include` function, I added `includePath` as a method name which sets an `include` property.

malikkotob: I agree, it's odd defining the option before the command. I'll refactor in a subsequent PR (unless you feel like it's a blocker - which I would understand).

src/Robo/Tasks/Drush.php
malikkotob: Happy to do that - only left it in as I was unsure if you were hoping to continue to use that terminology.

src/Robo/Tasks/Drush.php
malikkotob: Happy to make this change as well.

src/Robo/Tasks/Drush.php
malikkotob: No problem, I did it because it appeared cleaner programatically. I didn't see support through the `option` function to pass in an `=` to separate the option name and value as opposed to the space that they default to.

src/Robo/Tasks/Drush.php
malikkotob: Done

src/Robo/Tasks/Drush.php
malikkotob: Will run it through code sniffer. Happy to remove it - I set a default value because I think I saw that you had done so in your temporary `drush` function that you had put together.

src/Robo/Tasks/Drush.php
malikkotob: True - good point. Changed.

src/Robo/Tasks/Drush.php
malikkotob: Agreed, lol - I borrowed this notation [from the existing `DrushTask`](https://github.com/acquia/blt/blob/8.x/phing/phingcludes/DrushTask.php#L223) from the `phing` days. Happy to change.

src/Robo/Tasks/Drush.php
malikkotob: @grasmash are you running files through `phpcs` based on `drupal` standards or default `phpcs` standards?

PR #1414, opened 2017-04-20
Ensure integrity of configuration exports

danepowell:
This helps to ensure integrity of configuration exports by adding a test for configuration overrides subsequent to each config import. If a configuration remains overridden after import, it's a strong indication that you have a corrupt configuration export (typically uncaptured configuration changes.)

Note that these types of errors have been hidden for so long that I suspect almost all projects using core config or config split have at least one. Thus, I expect this test will fail when a lot of projects update BLT. We should make it clear in the release notes that this test is being added, so no one is surprised.

I think it's important that this test runs by default, but I have provided the option of disabling it via the `cm.allow-overrides` flag, if projects need to do that in the interim while fixing already-broken config.

### DIFF
--- a/phing/tasks/setup.xml
+++ b/phing/tasks/setup.xml
@@ -234,6 +234,14 @@
           <param>${cm.core.key}</param>
         </drush>
       </then>
+      <if>
+        <not><equals arg1="${cm.allow-overrides}" arg2="true"/></not>
+        <then>
+          <phingcall target="setup:update:config-override-check">
+            <property name="cm.dir" value="${cm.core.key}"/>
+          </phingcall>
+        </then>
+      </if>
     </if>
 
   </target>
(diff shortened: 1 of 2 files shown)


danepowell:
@grasmash I have no idea what this means, do you?

> /home/travis/build/acquia/blt-project/./vendor/acquia/blt/phing/tasks/setup.xml:226:8: IntrospectionHelper (unknown) doesn't support the 'not' creator/adder.

grasmash:
This looks like some sort of a Phing syntax error related to the not tag. Looks fine to me though.

danepowell:
I see what I did wrong. That error message could have been a little more helpful...

However it looks like you converted setup tasks to Robo so I guess I'll need to convert this PR as well.

grasmash:
@danepowell The command now lives here: https://github.com/acquia/blt/blob/8.x/src/Robo/Commands/Setup/ConfigCommand.php

danepowell:
I recreated this as #1421

PR #1418, opened 2017-04-21
Fixes #1403: Randomize user 1 name on install.

### LINKED ISSUE #1403: Randomize user 1 name on install
It's best practice to not use the default "admin" username for user 1, since that's the first thing brute-force attackers will attempt to use. All projects should change this eventually, so let's just automate it and get projects off on the right foot.

We can do this easily using the `--account-name` option when running `drush si`.

I would suggest using a username of the form `super-user-1234567890` (obviously randomize the number string), since this will pass Acquia Cloud Insight tests (don't want 'admin' anywhere in the name.)

I'm happy to implement this if you approve, just let me know.

grasmash:


src/Robo/Commands/Drupal/DrupalCommand.php
```diff
@@ -20,14 +21,16 @@ class DrupalCommand extends BltTasks {
    */
   public function install() {
 
+    $account_name = 'super-user-' . RandomString::string(10);
```
greylabel: does having `'super-user-'`in the name also defeat the purpose?

grasmash: I suppose that if someone can review and index of all the usernames, then it poses an issue.

greylabel: Core, at least D7, makes this pretty easy -- https://www.drupal.org/user/autocomplete/super-

PR #1421, opened 2017-04-21
Ensure configuration export integrity

danepowell:
See #1414 for explanation

This is completely untested at the moment, since I don't have any projects running robo setup commands, and I'm not currently able to install and patch the 8.8.3 release (@grasmash see my email)


src/Robo/Commands/Setup/ConfigCommand.php
```diff
@@ -66,6 +66,13 @@ public function import() {
             $task->exec("drush @$drush_alias pm-enable config_split --yes");
             $task->exec("drush @$drush_alias config-import sync --yes");
           }
+          if (!$this->getConfigValue('cm.allow-overrides')) {
+            $this->say("Checking for config overrides...");
+            $config_overrides = $task->exec("drush @$drush_alias cex sync -n | grep 'Differences of the active config'");
```
danepowell: I don't think task output is captured quite the same as Phing, this will need some work

danepowell:
The 8.x branch is too unstable to test right now (I can't even get a local install: #1447, #1446, #1448).

I'll circle back to this in a few days if things seem more stable.

danepowell:
Should be good to go now

PR #1422, opened 2017-04-21
Update drupal/simplesamlphp_auth module.

dpagini:
I'm taking a little shot in the dark here. I noticed the SimpleSAMLPHP_Auth module that BLT will add to your project is pointed to a commit hash. I tried looking into this quickly, and couldn't find an apparent reason. I'm guessing that this was pinned to a commit hash because at the time it was added, there was ONLY a dev branch available, and a commit hash is more stable than a flat out dev branch. As of today, there is an -rc2 release of this module, so I would think it would be more appropriate to remove the commit hash. Thoughts?

Changes proposed:
- Update the BLT version of `drupal/simplesamlphp_auth` module.

Pinging @dooleymatt on some of these questions...?

_@grasmash question: would you prefer a corresponding issue be added to go along with a PR like this?_


dpagini:
Actually, as I'm working on this in my project now... I have a question about another potential improvement to SAML library too. It looks like BLT copies a file called "acquia_config.php" to your root. Is there some room for improvement here by allowing some of these config files to live in BLT vendor directory and import from there, and then override in a single `simplesamlphp/config.php` file?

grasmash:
@dpagini as long as all values in it can be overridden, keeping that fine in vendor seems better.

Can you confirm that 3.0 contains 283994f92637181924cc2f606f9d0c78929e9c8c ?

dpagini:
Looks like that commit was an addition by @balsama last July to set the `simplesamlphp/simplesamlphp` library to ~1.14.2 and it has since been updated to ~1.14.4 in the RC2, so I think it's safe to say we have that commit. This is even making me think more it was just a commit to prevent from using a DEV branch in composer.

I'll look into the config values and what can be moved into vendor as we're getting into using this feature more. Thanks for the feedback!

PR #1426, opened 2017-04-24
Fixes #1328: Spaces in project paths break BLT.

### LINKED ISSUE #1328: Spaces in project paths break BLT

grasmash:


grasmash:
This is a partial solution. Need to search other command that use paths directly in taskExec() or execute() calls and wrap in quotes.

PR #1433, opened 2017-04-25
BLT-1048: adding configuration parameter to unit tests that will boot…

### LINKED ISSUE #1048: Allow PHPUnit Tests to Bootstrap Drupal
Currently PHPUnit doesn't include any references to Drupal (specifically, no bootstrap, and no reference to the Drupal config file at ${docroot}/core/phpunit.xml.dist).

It would be ideal if BLT had some method of allowing PHPUnit tests to include this parameter, as it is possible to have tests that are perfectly valid WITH this parameter that will FAIL without.

Note that I have attempted to modify the phing call https://github.com/acquia/blt/blob/8.x/phing/tasks/tests.xml#L243 per the parameters specific at https://www.phing.info/docs/guide/trunk/PHPUnitTask.html (see configuration) but it doesn't seem to actually load that configuration file. 

`<phpunit configuration="${docroot}/core/phpunit.xml.dist" printsummary="true" failureproperty="phpunit.failures" errorproperty="phpunit.errors">`

Unclear if this is a bug in phing itself or something wrong with BLT. I have attempted this with the file renamed to just be .xml (no luck). I've also attempted to pass a bootstrap parameter (instead of the configuration parameter) and that also appears to be non-functional.

mikemadison13:
…strap drupal if config: true.

Fixes #1048  .

Changes proposed:
- expands phpunit logic to check for a config boolean and if set to true will add a config option to the phpunit call that will allow drupal to boostrap.

### DIFF
--- a/src/Robo/Commands/Tests/PhpUnitCommand.php
+++ b/src/Robo/Commands/Tests/PhpUnitCommand.php
@@ -30,6 +30,18 @@ class PhpUnitCommand extends BltTasks {
   protected $testsDir;
 
   /**
+   * Wether or not to include the Drupal config file for PHPUnit.
+   *
+   * @var bool*/
+  protected $configStatus;
+
+  /**
+   * The path to Drupal's configuration file for PHPUnit.
+   *
+   * @var string*/
+  protected $configFile;
+
+  /**
    * This hook will fire for all commands in this command file.
    *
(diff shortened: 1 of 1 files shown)


mikemadison13:
see https://github.com/acquia/blt/pull/1435

PR #1444, opened 2017-04-26
BLT-000: Adds safe way to add a project dependency if using config_split.

bhasselbeck:
Changes proposed:
- Expands on the composer update methodology documented [here](http://blt.readthedocs.io/en/latest/readme/configuration-management/#updating-core-and-contributed-modules)


src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
       ->run();
   }
 
+  /**
+   * Checks the repo if there is modified files.
+   */
+  public function checkCleanRepo() {
+    $clean = $this->taskExec("git status --porcelain")
+      ->dir($this->getConfigValue('repo.root'))
+      ->interactive()
```
grasmash: We can probably remove this. The latest Robo detects interactive automatically.

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
       ->run();
   }
 
+  /**
+   * Checks the repo if there is modified files.
+   */
+  public function checkCleanRepo() {
```
grasmash: Because this is public, it will become a command automatically. I don't think we want that, so make it protected.

grasmash: We need a @return annotation.

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
...
+      ->interactive()
+      ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
+      ->run();
+    return $clean;
+  }
+
+  /**
+   * Adds a project dependency and ensures configuration management is in sync.
+   *
+   * @command add-dependency
```
grasmash: Please document the argument like this:
`* @param integer $one The first parameter.`

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
...
+   */
+  public function addDependency($dependency) {
+    $modified = $this->checkCleanRepo();
+    if (!empty($modified)) {
+      $this->output()->writeln("<comment>Please note the following files need to be commited or reverted:</comment>");
```
grasmash: Typo, `commited` to `committed`

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
...
+      $this->output()->writeln("<comment>$modified</comment>");
+    } else {
```
grasmash: Scan this with PHPCS. This does not follow Drupal coding standards.

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
...
+      $this->taskExec("composer update $dependency --with-dependencies")
```
grasmash: So, this method isn't actually updating a dependency. It's updating one, so maybe it should be renamed.

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -134,4 +134,45 @@ public function installAlias() {
...
+        ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
+        ->run();
+      $alias = $this->getConfigValue('drush.alias');
+      $this->taskExec("drush @$alias updb")
+        ->run();
+      $this->taskExec("drush @$alias cex")
+        ->run();
+      $modified = $this->checkCleanRepo();
+      if (!empty($modified)) {
+        $this->output()->writeln("<comment>The following files have configuration changes after add the $dependency</comment>");
```
grasmash: More than configuration may have changed. Might want to re-word this.

grasmash: This would say something like "after add the acquia/lightning".

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -142,4 +142,50 @@ public function installAlias() {
...
+   * @command add-dependency
+   */
+  public function addDependency($dependency) {
+    $modified = $this->checkCleanRepo();
+    if (!empty($modified)) {
+      $this->output()->writeln("<comment>Please note the following files need to be committed or reverted:</comment>");
+      $this->output()->writeln("<comment>$modified</comment>");
+    }
+    else {
+      $this->taskExec("composer update $dependency --with-dependencies")
```
grasmash: This will update a dependency, not add it. You need `composer require` to add it.

src/Robo/Commands/Setup/BuildCommand.php
```diff
@@ -142,4 +142,50 @@ public function installAlias() {
...
+        ->dir($this->getConfigValue('repo.root'))
+        ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
+        ->run();
+      $alias = $this->getConfigValue('drush.alias');
+      $this->taskExec("drush @$alias updb")
```
grasmash: If this is a module, it wasn't enabled. How would config have changed?

bhasselbeck: I'm lost with this - config is imported after the update instantiates... thus the config should be fine on top of it? No?

grasmash:
Closing this now that a `composer:require` command has been added to BLT.

PR #1460, opened 2017-05-02
Refactor drush task

malikkotob:

### DIFF
--- a/src/Robo/Tasks/DrushTask.php
+++ b/src/Robo/Tasks/DrushTask.php
@@ -72,7 +72,21 @@ class DrushTask extends CommandStack {
   protected $include;
 
   /**
-   * Runs the given drush command.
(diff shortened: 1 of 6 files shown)


src/Robo/Tasks/DrushTask.php
```diff
@@ -245,4 +249,84 @@ protected function mixedToBool($mixedVar) {
...
+  /**
+   * Set an option to be used for each drush command in the stack.
+   *
+   * @param string $option
+   * @param string $value
+   *
+   * @return $this
+   */
+  public function globalOption($option, $value = null) {
+    if ($option !== null and strpos($option, '-') !== 0) {
```
grasmash: I don't totally understand why it's necessary to override parent::option and also have a separate globalOption method. Can you explain and perhaps add a comment to the code with the explanation?

PR #1464, opened 2017-05-03
Return $this in override of option method in drush task.

malikkotob:


malikkotob:
Replaced by https://github.com/acquia/blt/pull/1465

PR #1482, opened 2017-05-05
BLT-1481: allowing for combo of features and default config.

### LINKED ISSUE #1481: Allow Both Features and Default CMI
There are some things that features does not do well, and for some sites that are already using both features and config/default we need to support this (as the configuration will not work properly if it isn't imported during the drupal:install process).

**Proposed Solution**
Add a new condition into the drupal:install that checks for keys in the project.yml file so that even if features is involved, the config file can be defined.

mikemadison13:
Fixes #1481  .

Changes proposed:
- allow both features and default configuration on the same project
- imports default configuration during installation (and then imports features after)


grasmash:
This was addressed in #1488

PR #1492, opened 2017-05-09
Exclude composer files from artifact.

dpagini:
Throwing this up as a reaction to some of the composer BoF discussion at Drupalcon. I don't know if there some consideration here for why we want `composer.json`, `composer.lock` and `/patches` in your artifact. If we don't want people running `composer install` on production, this takes away that option.

This is probably more for discussion, but figured getting this up would be the best way to start the convo.  =)

### DIFF
--- a/phing/files/deploy-exclude.txt
+++ b/phing/files/deploy-exclude.txt
@@ -17,6 +17,7 @@
 /docroot/sites/*/private
 /docroot/themes/contrib
 /drush/contrib
+/patches
 /README.md
 /readme
 /sites/development.services.yml
@@ -29,5 +30,6 @@ example.*
 local.settings.php
 local.drushrc.php
 node_modules
+/composer.*
 /vendor
 project.local.yml


dpagini:
Alternate thought: maybe leave `composer.lock` in the artifact? It _sort of_ is an artifact of `composer.json`, provides some information about the installed modules in the repo, but still limits the end user's ability to run `$ composer install` on production.

grasmash:
Hm. I'm trying to think of reasons to keep these files. The only one that I can think of is that it would be nice to use composer to inspect the package versions, and that it might be useful as a manifest representing how the artifact was constructed.

dpagini:
Let it breathe for a few days and leave open for comments?  =)  I'm also not offended if you'd rather leave as is for now... just feel free to close this.

I was trying to think of similar reasons to keep it... this really all started b/c I was wondering why we deploy `patches/` in the artifact? And then we had the BoF and I feel like I heard similar things.

So given your response though, it would seem to make more sense for my second suggestion. `composer.json` isn't _really_ helpful, b/c it's really just going to say what the minimum requirements are, right? Or do you mean running something like `$ composer info drupal/core` on your prod server?

grasmash:
After some thought, I'd think I'd prefer to keep these files in the artifact unless it presents a clera problem for someone.

dpagini:
Fair enough... anything in particular make you lean that way?  j/w

PR #1515, opened 2017-05-17
Removing unneeded composer update call.

grasmash:


grasmash:
This will be split into multiple, smaller PRs.

PR #1529, opened 2017-05-22
Remove double quote from drush command.

malikkotob:


src/Robo/Commands/Setup/SyncCommand.php
```diff
@@ -116,7 +116,7 @@ public function syncDbDefault() {
       $task->option('sanitize');
     }
 
-    $task->drush('cache-clear drush');
+    $task->drush('cc drush');
```
grasmash: Please use full command, not aliases.

PR #1561, opened 2017-05-31
Update slack.sh

sherakama:
It looks like the path to this has changed and is no longer in $HOME.

Fixes # .
I did not create an issue. I hope this suffices. 

Changes proposed:
- Path variable to the stored slack_settings file.

### DIFF
--- a/scripts/cloud-hooks/hooks/common/slack.sh
+++ b/scripts/cloud-hooks/hooks/common/slack.sh
@@ -1,10 +1,10 @@
 #!/bin/sh
 
-FILE=$HOME/slack_settings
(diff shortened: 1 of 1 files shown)


grasmash:
> It looks like the path to this has changed

Not sure what you mean by this, BLT doesn't ship slack_settings. We expect you to create a file and place it in `$HOME/slack_settings`. Is that directory not writable anymore?

sherakama:
ah gotcha. This was just me then mis-reading the documentation. closing.

grasmash:
Feel free to update the docs if they're unclear.

PR #1563, opened 2017-06-01
Drupal libraries path in composer install should match .gitignore

danielnitschelc:
- [.gitignore](https://github.com/acquia/blt/blob/8.x/template/.gitignore#L23) references `docroot/libraries/contrib`
- but [composer.json](https://github.com/acquia/blt/blob/8.x/template/composer.json#L22) uses `docroot/libraries/{$name}` 

It seems to me these should be the same.

This pull request updates the libraries path in `composer.json` to match `.gitignore`.

### DIFF
--- a/template/composer.json
+++ b/template/composer.json
@@ -19,7 +19,7 @@
             "docroot/profiles/custom/{$name}": ["type:drupal-custom-profile"],
             "docroot/themes/contrib/{$name}": ["type:drupal-theme"],
             "docroot/themes/custom/{$name}": ["type:drupal-custom-theme"],
-            "docroot/libraries/{$name}": ["type:drupal-library"],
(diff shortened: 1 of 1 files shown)


greylabel:
The composer path should not change, IMO. You can always specify /contrib as the installer path in your individual library definition in the need that.

danielnitschelc:
@greylabel do you think the .gitignore should be changed then?

greylabel:
@danielnitschelc maybe adding `docroot/libraries` to `.gitignore` makes the most sense. I can't think of any contrib modules that actually want for libraries in `docroot/libraries/contrib`. Webform, for example, expects libraries to be placed directly in `docroot/libraries`

grasmash:
> maybe adding docroot/libraries to .gitignore makes the most sense.

Agreed.

Issue #1564, opened 2017-06-01
Execute the front-end tests target while the Drush Webserver is active


aweingarten:
As a front-end developer I want to be able to run tests against the drush webserver as part of a pipelines build.    Right now the webserver is being initiated as part of the behat task and then terminated immediately after the behat tests are run.  It can't be leveraged for front-end tasks like visual regression tests, or WCAGAA testing.  

Ideally the web-server should be left open for the front-end tests target.

grasmash:
Eh, I'm not sure I want to do this for everyone.

What we could do is introduce a command like `blt tests:run-server` and `blt tests:kill-server` so that you can launch and kill it yourself during frontend testing. 

How's that sound?

aweingarten:
yeah, I agree this gets messy.  How do you see tests:run-server and tests:kill-server getting invoked? Would one override the front-end tests command to add those calls?  Or did you imagine something else

grasmash:
The front end tests command just calls a target hook. You can use that target hook to invoke a script. E.g.,

```
blt tests:run-server
# something else
blt tests:kill-server
```

PR #1570, opened 2017-06-01
Added frontend-web-test hook to BLT.

aweingarten:
Fixes https://github.com/acquia/blt/issues/1564

Changes proposed:
- Add a new frontend-web-test hook that fires inside the behat tests.
-

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -170,6 +170,11 @@ target-hooks:
     dir: ${docroot}
     # E.g., `npm test`.
     command: null
+  frontend-web-test:
+    # E.g., ${docroot}/sites/all/themes/custom/mytheme.
+    dir: ${docroot}
+    # E.g., `npm test`.
+    command: null
   pre-commit:
     # E.g., ${docroot}
     dir: ${docroot}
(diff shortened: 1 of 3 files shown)


grasmash:
Can you explain the need for this?

aweingarten:
Need an ability to run front-end tests with a web-server, and installed drupal site present.   Doing this here seemed to be the safest way to accomplish this goal. 

Any other approach would likely result in a second duplicate robo target which could needlessly increase the build times.

grasmash:
Why can't we just use frontend-test for this?

aweingarten:
Nope because the webserver doesn't exist during frontend-test.  

We could totally have the front-end test recreate all the stuff that is done in the behat target but that seems redundant.

grasmash:
Hm, it still doesn't make sense to me that we'd create a new target for this. Rather, it seems like we need to execute tests in a different order. E.g.,

```
blt run-server
blt tests:all
blt kill-server
```

We'd basically remove the server start/kill logic from the tests command and require that these be run manually.

grasmash:
Closing in favor if #1615

PR #1616, opened 2017-06-06
Cleaned up ACSF init script.

danepowell:
Various parts of this were just broken.

### DIFF
--- a/bin/blt-robo-run.php
+++ b/bin/blt-robo-run.php
@@ -6,7 +6,7 @@
  */
 
 use Acquia\Blt\Robo\Config\YamlConfigProcessor;
-use Robo\Config\YamlConfigLoader;
+use Consolidation\Config\Loader\YamlConfigLoader;
 use Robo\Robo;
 use Symfony\Component\Console\Input\ArgvInput;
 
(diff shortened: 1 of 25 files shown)


grasmash:
It looks like this just enables the acsf module before executing the init script. Is that necessary?

danepowell:
Yes, the ACSF module needs to be enabled and Drupal needs to be bootstrapped (hence the addition of the drush aliases) in order for acsf_init to run.

danepowell:
Having said that, I don't understand why tests are failing because Drupal can't be bootstrapped... I would have expected tests to fail _before_ this change...

lcatlett:
I dont think drupal needs to be boostrapped to run acsf-init which is why it uses the includes.

danepowell:
When I run this:
`drush --include=docroot/modules/contrib/acsf/acsf_init acsf-init`

I get this:
`Command acsf-init needs a higher bootstrap level to run`

So it does seem like Drupal needs to be bootstrapped. That's why I added the alias.

Once I added the alias, I ran the command without the acsf module enabled, and got this:
`Command acsf-init needs the following extension(s) enabled to run: acsf_init`

That seems like pretty clear evidence that Drupal has to be bootstrap and the module needs to be enabled, hence these changes.

danepowell:
This documentation should also be added to the mkdocs

danepowell:
I think Travis fails because ACSF init now needs to happen after Drupal can be bootstrapped

danepowell:
This is hopelessly out of date

PR #1618, opened 2017-06-07
Inspector 'VM is running' bug

simesy:
Inspector was telling me the VM wasn't running. This seems to fix it but might not be the right way.


simesy:
Fixed in acquia/blt#1619

PR #1622, opened 2017-06-08
BLT requires PHP7 due to ocramius/package-versions

simesy:


grasmash:
Did you run into some issue with using PHP 5.6? I've just tested using PHP 5.6 against BLT 8.x and I did not encounter any warnings or errors. I also executed `composer update` and had no issues building the dependencies.

grasmash:
Additional `ocramius/package-versions` is not a dependency of BLT or BLT based projects as far as I can tell.

```
composer info ocramius/package-versions

  [InvalidArgumentException]
  Package ocramius/package-versions not found

```

simesy:
That's odd:
https://github.com/Ocramius/PackageVersions/blob/1.1.1/composer.json#L13

```
$ composer why ocramius/package-versions
ocramius/proxy-manager  2.1.1  requires  ocramius/package-versions (^1.1.1)

$ composer why ocramius/proxy-manager
sensiolabs/behat-page-object-extension  v2.1.0  requires  ocramius/proxy-manager (^1.0||^2.0)

$ composer why sensiolabs/behat-page-object-extension
acquia/blt-project foo-branch  requires  sensiolabs/behat-page-object-extension (^2.0)
```

simesy:
I'd be interested in what's different about your setup @grasmash?

simesy:
OK, so we picked that up from using blt-project. *sigh*

PR #1638, opened 2017-06-12
Fixes #1631: Front-end Tests not executed during pipelines build.

### LINKED ISSUE #1631: Front-end Tests not executed during pipelines build
When I run a pipelines build on the head of 8.9.x the following hook does not fire
```
target-hooks:
  frontend-test:
    dir: '${docroot}/themes/custom/freerange'
    command: 'yarn install && npm run pa11y'
```

I think a regression was introduced when the tests were refactored to provide a webserver in CI.

grasmash:


grasmash:
We may want to change the name of this command to `tests:frontend`. Also, manually launch and kill server before all commands, rather than doing it per command.

PR #1640, opened 2017-06-12
Remove nonexistent ansi option.

webbj74:
In 8.8.4 `scripts/pipelines/run_tests` fails because `--ansi` is not a phing option

Changes proposed:
- Removes the `--ansi` option from the call to `blt validate:all`
- Removes the `--ansi` option from the clall to `blt ci:setup`

NOTE: the presence of the `--ansi` option did not seem to affect the call to `blt tests:all`


grasmash:
Thank you. The `--ansi` option was mistakenly backported from 8.x to 8.8.x.

PR #1647, opened 2017-06-13
Fixes #1646: Improve BLT's handling of development.services.yml.

### LINKED ISSUE #1646: Improve BLT's handling of development.services.yml
BLT's `composer.required.json` includes code similar to the below:

```
  "extra": {   
    "drupal-scaffold": {
      "excludes": [
        "sites/development.services.yml"
      ]
    },
  },
```

This has led to `sites/development.services.yml` drifting from the version of this file provided by Drupal core.

I feel we should do one of the below, in order of my preference:
1. Remove this exclusion to bring `sites/development.services.yml` completely inline with the version provided by Drupal core.
1. Remove this exclusion and use `post-drupal-scaffold-cmd` to patch-in the current customizations.
1. Manually update `sites/development.services.yml` to bring it more inline with the version provided by Drupal core.

greylabel:
Fixes #1646: Improve BLT's handling of development.services.yml.

Changes proposed:
- Remove exclusion of `development.services.yml` from `drupal-scaffold` to bring this file inline with the version provided by Drupal core.


grasmash:
We'll also need to remove template/docroot/sites/development.services.yml from BLT.

grasmash:
Actually, [core's development.services.yml](http://cgit.drupalcode.org/drupal/tree/sites/development.services.yml) and [BLT's](https://github.com/acquia/blt/blob/8.x/template/docroot/sites/development.services.yml) are functionally different. Namely, BLT disables the twig cache by default.

Are we willing to lose that?

greylabel:
I think BLT should either patch it or leave it up to users to patch, but I feel using `exclude` is not the right solution.

grasmash:
I personally would like to keep BLT's patches to a minimum, given the their tendency to cause hotfixes.

grasmash:
@danepowell I think you may have initially added this file. Do you have an opinion?

danepowell:
@grasmash let's split up development.services.yml... we'll use the stock version provided by Drupal core, and then maintain a separate file with just the BLT additions (i.e. the Twig cache setting). They can both be included via [local.settings.php](https://github.com/acquia/blt/blob/8.x/settings/default.local.settings.php#L31)

greylabel:
@danepowell Any reason to not just patch core's `development.services.yml` with `post-drupal-scaffold-cmd`?

grasmash:
I've learned that providing patches by default in BLT almost always requires a fast hot fix down the line. On principal, I just prefer not to use them.

greylabel:
Makes sense. The dual file approach is fine, but segments things a bit. FWIW, I think the customizations in BLT's `development.services.yml` are pretty much a given for any local dev setup, but I just really don't want to miss out on upstream changes to Drupal's version.

danepowell:
You probably can't patch this file anyway, due to limitations with Drupal Scaffold (same as htaccess).

Dual-file approach seems like a great solution (of course I'm biased, since I proposed it)

greylabel:
Not to move too far off topic, but @danepowell Is there more specific inflation on the the limitations?

Quick test:

I added a patch file at `patches/htaccess-ssl.patch`.

Added to `composer.json`:

```
    "scripts": {
        "post-drupal-scaffold-cmd": [
            "cd docroot && patch -p1 <../patches/htaccess-ssl.patch"
        ]
    },
```

Removed `/docroot/core` and other drupal-scaffold provided files.

Ran `composer update` -- last part of the output:

```
> cd docroot && patch -p1 <../patches/htaccess-ssl.patch
patching file .htaccess
```

Noticed that the expected changes from `patches/htaccess-ssl.patch` are present in `.htaccess`

danepowell:
@greylabel sure that method of patching probably works fine (never tried it myself). I just meant that cweagans/composer-patches and drupal-composer/drupal-scaffold don't play nicely together, at least last time I checked. If you simply add a patch for htaccess (or development.services.yml) to the extra.patches array in composer.json, it won't be applied.

grasmash:
@greylabel If you refactor this PR to use the "dual file approach", I'll merge it.

greylabel:
SG. I'll update this PR shortly.

greylabel:
Made a quick pass at this by creating a `blt.development.services.yml` file. We could also rename this to something like `twig.development.services.yml` since it really just concerns twig related params.

grasmash:
Well done, thank you!

PR #1650, opened 2017-06-13
Removing .htaccess files from non-docroots.

dpagini:
Changes proposed:
- Removing the `.htaccess` files from the config folders, since these are above the docroot folders. There may be a reason we're keeping them here (just in case?), but I guess this is my way of asking that question...?  =)

If we want them there, maybe we use 1 `.htaccess` file just in `config/.htaccess` that protects the whole set of folders maybe?


grasmash:
I think these may have somehow been autogenerated by Drupal.

PR #1661, opened 2017-06-13
Importing config splits after site install.

grasmash:


danepowell:
I've tested this and can confirm it works. Although I'm on the fence about whether we should be calling config-import just for config-split here, or call `setup:update` unconditionally. The current approach is definitely faster (no unnecessary cache clears, etc), but calling `setup:update` would be a little more robust. :man_shrugging:

grasmash:
For a core only config approach, the additional call would do nothing, right? Not sure of feautures.

danepowell:
Correct. For features I think the only benefit to a dedicated import would be the automatic check for overrides. I'm not sure how many people even use that though.

PR #1664, opened 2017-06-14
Added pinned version of node to v6.11.0 latest LTS.  Also added yarn.

aweingarten:
Adds node basics that all projects depend on.

### DIFF
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -16,6 +16,9 @@ events:
               - composer validate --no-check-all --ansi
               - composer install --ansi
               - source ${BLT_DIR}/scripts/pipelines/setup_env
+              - nvm install v6.11.0
+              - npm install
+              - npm install --global yarn
         - validate:
             type: script
             script:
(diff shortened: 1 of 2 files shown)


danepowell:
@aweingarten not sure if by "all projects" you are referring to Cog (that's the only standard tool I'm aware of that requires Yarn). FYI, Cog is actually moving away from Yarn: https://github.com/acquia-pso/cog/pull/100

danepowell:
Also I'm not sure it's safe to bump the node version like that without coordinating with the Cog team. I know it has a lot of deprecated dependencies that could start failing with higher node versions.

scripts/pipelines/acquia-pipelines.yml
```diff
@@ -16,6 +16,9 @@ events:
               - composer validate --no-check-all --ansi
               - composer install --ansi
               - source ${BLT_DIR}/scripts/pipelines/setup_env
+              - nvm install v6.11.0
```
briantully: i would suggest using the 'lts/boron' name in case there are any incremental patch or minor updates, e.g., 

`nvm install lts/boron`

grasmash:
Rather than installing any version of node or any node package, I would prefer to call a script belonging to the front end tool (cog) and let it take care of the packages.

This is what the `blt frontend:setup` command exists for. 

I think a better approach would be to add a default target hook for `frontend-setup` that calls a Cog  script. The Cog script can then install its required version of node, yarn, or whatever else.

PR #1665, opened 2017-06-14
Run the appropriate behat tags when testing on pipelines.

arknoll:
The test script currently hard codes the behat.tags to just ~experimental. This should not be the case.

### DIFF
--- a/scripts/pipelines/tests
+++ b/scripts/pipelines/tests
@@ -4,6 +4,6 @@ set -ev
 
 export PATH=${COMPOSER_BIN}:${PATH}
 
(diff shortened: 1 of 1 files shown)


grasmash:
This override was used to workaround a previous issue in Lightning. If it is no longer required, we can just remove the behat.tags argument entirely.

arknoll:
@grasmash I updated the PR. I am going to test to make sure this patch works for me. I'll post here when it does.

arknoll:
@grasmash this PR worked for me.

PR #1668, opened 2017-06-15
BLT-1667: Provide memcache integration and sample settings template.

### LINKED ISSUE #1667: Update memcache to recommended version and provide SASL settings template
The suggested version of memcache should be 2.0-alpha3. Per-site memcached and SASL authentication is now available on Acquia Cloud - provide sample settings.php template for memache integration.

lcatlett:
Fixes #1667 

Changes proposed:
- Update memcache suggested version to 
- Patch core to allow for graceful cache fallback
- Provide sample settings.php for configuring memcache for default cache, per-site memcached and SASL authentication

### DIFF
--- a/settings/sample-memcache.settings.php
+++ b/settings/sample-memcache.settings.php
@@ -1,0 +1,20 @@
+<?php
+
+/**
+ * @file
+ * Template for memcache configuration and SASL authentication for Acquia Cloud
+ */
+
+if ($is_ah_env) {
+  $settings['cache']['default'] = 'cache.backend.memcache';
+  $settings['memcache']['sasl'] = [
+  'username' => 'user',
+  'password' => 'password',
+];
+  // When using SASL, Memcached extension needs to be used
+  // because Memcache extension doesn't support it.
+  $settings['memcache']['extension'] = 'Memcached';
+  $settings['memcache']['options'] = [
+  \Memcached::OPT_BINARY_PROTOCOL => TRUE,
+];
+}
--- a/composer.suggested.json
+++ b/composer.suggested.json
@@ -4,13 +4,16 @@
     "drupal/acquia_connector": "^1.5.0",
     "drupal/cog": "^1.0.0",
     "drupal/qa_accounts": "^1.0.0-alpha1",
-    "drupal/memcache": "2.x-dev",
+    "drupal/memcache": "^2.0-alpha3",
     "drupal/seckit": "^1.0.0-alpha2",
     "drupal/security_review": "*",
     "drupal/shield": "^1.0.0"
   },
   "extra": {
     "patches": {
+        "drupal/core": {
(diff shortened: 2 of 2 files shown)


settings/sample-memcache.settings.php
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @file
+ * Template for memcache configuration and SASL authentication for Acquia Cloud
+ */
+
+if ($is_ah_env) {
+  $settings['cache']['default'] = 'cache.backend.memcache';
+  $settings['memcache']['sasl'] = [
+  'username' => 'user',
```
greylabel: Is there documentation for finding/setting these values? We should probably provide a link.

typhonius: Is SASL supported within Acquia? I'm wondering if SASL should be included only with a configuration parameter. Additionally, would we want to also include memcache on local environments (or even CI)?

grasmash:
Based on experience, I would prefer not to include any patches in BLT itself. We can track the progress of the Drupal issue here: https://www.drupal.org/node/2766509

greylabel:
If we want to provide a boilerplate memcache config, there is more we should add. We should also look at removing the APCu hack the provided in settings currently. I can give this a shot.

settings/sample-memcache.settings.php
```diff
@@ -0,0 +1,20 @@
...
+  'password' => 'password',
+];
+  // When using SASL, Memcached extension needs to be used
+  // because Memcache extension doesn't support it.
+  $settings['memcache']['extension'] = 'Memcached';
+  $settings['memcache']['options'] = [
+  \Memcached::OPT_BINARY_PROTOCOL => TRUE,
```
greylabel: @lcatlett Do you have any experience running this config prod? I would love to see or hear about any perf data from sites running this in the wild.

http://cgit.drupalcode.org/memcache/tree/README.txt?h=8.x-2.x#n268

```
Other options you could experiment with:
 + Memcached::OPT_BINARY_PROTOCOL => TRUE,
    * This enables the Memcache binary protocol (only available in Memcached
      1.4 and later). Note that some users have reported SLOWER performance
      with this feature enabled. It should only be enabled on extremely high
      traffic networks where memcache network traffic is a bottleneck.
      Additional reading about the binary protocol:
        https://raw.githubusercontent.com/memcached/old-wiki/master/MemcacheBinaryProtocol.wiki
        Note: The information on the link above will eventually be ported to
        the new wiki under https://github.com/memcached/memcached/wiki.
```

grasmash:
8.x already requires `^2.0-alpha3`.  Boilerplate code will be addressed by #1668.

settings/sample-memcache.settings.php
greylabel: @lcatlett still curious about this, though, despite the issues being closed if you get a moment.

PR #1671, opened 2017-06-15
Setting tests.run-server to false by default.

grasmash:


arunavak111:
May I know why this setting set us false by default?

Issue #1681, opened 2017-06-16
Behat Tests Fail


mikemadison13:
My system information:
* Operating system type: Mac OS
* Operating system version: 10.11.6
* BLT version: 8.9.x dev

When I run this command:
```
blt tests:all
blt tests:behat
```

I get the following output:
```
Executing command blt tests:behat  --define drush.alias=self inside of Drupal VM...
[warning] The xDebug extension is loaded. This will significantly decrease performance.
[Testing\Behat] Running behat  --format pretty /vagrant/tests/behat --colors --no-interaction --stop-on-failure --strict --config /vagrant/tests/behat/local.yml --profile local --tags '~ajax&&~experimental&&~lightningextension' -v

PHP Fatal error:  Cannot redeclare composerRequire50aaad88ede7074a839163f1fb337cad() (previously declared in /vagrant/vendor/composer/autoload_real.php:63) in /var/www/<project>/vendor/composer/autoload_real.php on line 70

Fatal error: Cannot redeclare composerRequire50aaad88ede7074a839163f1fb337cad() (previously declared in /vagrant/vendor/composer/autoload_real.php:63) in /var/www/<project>/vendor/composer/autoload_real.php on line 70
[Testing\Behat]  Exit code 255 
[error]  Behat tests failed! 
```

And I expected this to happen:

mikemadison13:
@geerlingguy FYI this seems to be an issue with an interaction between BLT and Vagrant... what @grasmash and I are seeing is that the tests work fine "inside" the VM but when attempting to run outside for some reason things are looking in /vagrant (and when I do a vagrant exec pwd it also points at /vagrant instead of my repo root inside the vm).

I just re-provisioned with a ssh_home variable in box/config.yml... any thoughts on what else might be going on here?

typhonius:
The documentation here: http://docs.drupalvm.com/en/latest/extras/drupal-console/#remote-command-execution-using-vagrant-exec works when I manually create a Vagrantfile.local. With this in mind, should BLT create a Vagrantfile.local with this config during one of the settings phases?

grasmash:
@geerlingguy Would you consider adding such setting to Drupal VM?

geerlingguy:
I would rather go the route @typhonius recommends — see my notes in upstream PR https://github.com/geerlingguy/drupal-vm/pull/1435

geerlingguy:
[...]

PR #1688, opened 2017-06-19
Issue #1681: Set vagrant exec default directory in BLT Vagrantfile.

geerlingguy:
We would also need to actually update the path inside the Vagrantfile inside the VmCommand.php class... but I'm thinking this would work fine for BLT's needs.

### DIFF
--- a/scripts/drupal-vm/Vagrantfile
+++ b/scripts/drupal-vm/Vagrantfile
@@ -11,3 +11,8 @@ ENV['DRUPALVM_DIR'] = "vendor/geerlingguy/drupal-vm"
 
 # Load the real Vagrantfile
 load "#{__dir__}/#{ENV['DRUPALVM_DIR']}/Vagrantfile"
+
+# Set up vagrant exec for BLT.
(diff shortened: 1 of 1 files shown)


scripts/drupal-vm/Vagrantfile
```diff
@@ -11,3 +11,8 @@ ENV['DRUPALVM_DIR'] = "vendor/geerlingguy/drupal-vm"
 
 # Load the real Vagrantfile
 load "#{__dir__}/#{ENV['DRUPALVM_DIR']}/Vagrantfile"
+
+# Set up vagrant exec for BLT.
+if Vagrant.has_plugin?('vagrant-exec')
+  config.exec.commands '*', directory: /var/www/${project.machine_name}
```
grasmash: I don't think that a Robo variable in this file will get resolved.

grasmash: Resolved in #1690.

geerlingguy: +1

PR #1712, opened 2017-06-21
Fix docs for creating a deploy-exclude.txt file.

greylabel:
See: https://github.com/acquia/blt/tree/8.x/scripts/blt/deploy

Changes proposed:
- Update to docs for _Committing dependencies_

### DIFF
--- a/readme/committing-dependencies.md
+++ b/readme/committing-dependencies.md
@@ -16,7 +16,7 @@ However, there are sometimes extenuating circumstances that require you to commi
 
 * Create a custom deploy.exclude_file and reference its location in your project.yml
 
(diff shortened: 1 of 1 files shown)


grasmash:
@greylabel FYI I don't think these docs are listed in the menu structure because this approach is somewhat untested. I'd not be surprised to find issues with the approach, though there is an open ticket for adding this feature.

greylabel:
@grasmash Are you referring to menu structure on http://blt.readthedocs.io/? I assume this is the tested/supported approach - http://blt.readthedocs.io/en/stable/readme/extending-blt/?

Issue #1711, opened 2017-06-21
Potential problem with "--define" syntax?


dpagini:
Scenario...

I'm looking at the `blt/scripts/travis/run_tests` file to try and debug some of my Travis issues, and have honed in just a little bit on the `$ blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --yes -v` line.

Should I be able to run this locally? It seems to me like the "${drush.aliases.ci}" piece is getting ignored completely. My setting for `drush.aliases.ci` is "self" currently. 
Here is my attempt to run two commands locally:
```sh
$ blt setup:toggle-modules --define drush.alias='${drush.aliases.ci}'
[Acquia\Blt\Robo\Tasks\DrushTask] Running /Users/dpagini/Projects/Websites/platform/vendor/bin/drush @platform.vm pm-enable --skip dblog devel field_ui features_ui page_manager_ui seckit stage_file_proxy views_ui webform_ui --uri=default --yes
```

```sh
$ blt setup:toggle-modules --define drush.alias=self
[Acquia\Blt\Robo\Tasks\DrushTask] Running /Users/dpagini/Projects/Websites/platform/vendor/bin/drush @self pm-enable --skip dblog devel field_ui features_ui page_manager_ui seckit stage_file_proxy views_ui webform_ui --uri=default --ye
```

Should that first command pickup the value of `drush.aliases.ci`? Is there any reason Travis would be able to get that value, but I wouldn't get it locally?

_Note: I only chose "setup:toggle-modules" b/c it was a command that ran fast that I knew used the value of `drush.alias`._

grasmash:
Hm. I cannot reproduce your issue. When I set drush.aliases.ci to `dpagini` and run `blt setup:toggle-modules --define drush.alias='${drush.aliases.ci}'`
I see:

```
Running /Users/matthew.grasmick/Sites/acquia/blted8/vendor/bin/drush @dpagini pm-enable --skip dblog devel seckit views_ui --uri=default --yes in /Users/matthew.grasmick/Sites/acquia/blted8/docroot
```

grasmash:
What is your PHP version and what shell are you running?

dpagini:
Hmm... odd. Ok, I will keep debugging myself then. 

```
$ php -v
PHP 5.6.21 (cli) (built: May 24 2016 23:20:56) 
Copyright (c) 1997-2016 The PHP Group
[...]

PR #1714, opened 2017-06-21
Connects to #1711: Adding config:get and config:dump commmands.

grasmash:


grasmash:
@aweingarten Pretty sure you asked for this one too.

PR #1733, opened 2017-06-27
Provide initial template for Gitlab Pipelines

snufkin:
- Adding an initial example gitlab CI configuration file.
- Adding Robo task that copies the template to the project root.

### DIFF
--- a/scripts/gitlab/gitlab-ci.yml
+++ b/scripts/gitlab/gitlab-ci.yml
@@ -1,4 +1,5 @@
 stages:
+  - validate
   - build
   - test
   - deploy
@@ -9,16 +10,22 @@ variables:
 before_script:
   - echo "hello world"
 
+validate:
+  stage: validate
+  script:
+    - blt validate:all
+
 build:
   stage: build
   script:
-    - composer install
+    - composer validate --no-check-all --ansi
+    - composer install --ansi
     - blt setup:build
 
-validate:
+test:
   stage: test
   script:
-    - blt validate
+    - blt tests --define tests.run-server=true --yes -v
 
 production:
   stage: deploy
--- a/scripts/gitlab/gitlab-ci.yml
+++ b/scripts/gitlab/gitlab-ci.yml
@@ -1,0 +1,31 @@
+stages:
+  - build
+  - test
+  - deploy
+
+variables:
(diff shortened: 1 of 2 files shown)


grasmash:
Nice!

scripts/gitlab/gitlab-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+  APP_NAME: Project
+
+before_script:
+  - echo "hello world"
+
+build:
+  stage: build
+  script:
+    - composer install
+    - blt setup:build
```
grasmash: Any reason not to simply run `blt setup` ?

scripts/gitlab/gitlab-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+
+validate:
+  stage: test
+  script:
+    - blt validate
+
```
grasmash: Seems like we should call `blt tests` here with some arguments like `--define tests.run-server=true`

scripts/gitlab/gitlab-ci.yml
snufkin: Well in my build step I'm actually composing a docker image, which contains just the static site build, without actually installing it. `blt setup` would also require mysql, and strictly speaking it would not belong to the build stage. (having a site install in there is a valid point though)

scripts/gitlab/gitlab-ci.yml
snufkin: In pipelines you define various build stages and then tasks for each stage. This particular example was to demonstrate that within the `test` stage you can have different task executions, e.g. in this one only linting happens.

grasmash:
I think this is incomplete. I'm still willing to merge it, but I'd like to add a warning after the command is executed:

`$this->logger->warning("Gitlab support is experimental and may not support all BLT features.");`

Or something like that.

snufkin:
Let me push some updates in the coming few workdays before you merge.

snufkin:
- Updated the template file to reflect correct invocations of BLT commands, based on the other CI examples.
- Added a validation step separately from testing so that CI would fail on CS violations.
- Added an extra warning about the template being experimental.

PR #1735, opened 2017-06-28
Revert "Defaulting web driver to Selenium."

grasmash:
Reverts acquia/blt#1729


danielnitschelc:
Sorry if I missed the context on this, why was this reverted?

PR #1740, opened 2017-06-29
Adding frontend task to setup:build.

grasmash:
@dpagini ^^


dpagini:
Awesome thanks! I'm up and running on 8.9.* now. I have a handful of changes that I need to make in a _better way_, but everything is working as expected now I think. Thanks again for all your work!

PR #1743, opened 2017-06-29
Switched pipelines default web-driver to chrome.

aweingarten:
Fixes # .

Changes proposed:
-
-


grasmash:
We  also need to modify the pipelines initialization command so that it sets the Behat config appropriately.

PR #1754, opened 2017-07-02
Making Drupal-VM config customizable. 🇺🇸

dpagini:
This PR attempts to make the installation directory (currently `box/`) configurable when the vm command runs. That will remain the default, but this command uses `askDefault()` method to request a change to the installation.

The location of the drupal-vm config file is moved into `build.yml` config and will be overridable in the local `project.yml`. In fact, if you specify an override at runtime, that new location is written to your `project.yml`

The input is stripped of the directory separators, so you can enter `/box2` or `box2/` and they will all equal `box2`. The proper files are copied into this new location, and all references to the file are updated (Vagrantfile).

I also cut down on the size of the Drupal-VM post-provision script. I'm not sure if the lines deleted in that file were doing something else I'm unaware of, or just getting the docroot to install the alias. Instead of "cd"ing into the `/var/www` path, I just specified the install-alias script from the `/vagrant/vendor` directory, and this seemed to work for me. Thoughts on this approach? That felt too simple, so maybe I'm missing something here...?

#### What could be better?
* It seems to me the block to write to the `project.yml` file is a little larger than it feels like it should be. I copied this example from the SAML command, but there might be room to make a class for updating that file to make it easier.
* Validation of user entry. I wonder what happens if you enter "../box"?
* Tests to ensure default and custom entries work, and that `vm:nuke` considers custom filepath. Anything else?

_I'm not sure how this is going to play with tests, so I'll just watch Travis. I may need help trying to figure out how to "hit enter" or enter an alternate value._

### DIFF
--- a/scripts/drupal-vm/Vagrantfile
+++ b/scripts/drupal-vm/Vagrantfile
@@ -4,7 +4,7 @@ ENV['DRUPALVM_PROJECT_ROOT'] = "#{__dir__}"
 
 # The relative path from the project root to the config directory where you
 # placed your config.yml file.
-ENV['DRUPALVM_CONFIG_DIR'] = "box"
+ENV['DRUPALVM_CONFIG_DIR'] = "${drupalvm.config.dir}"
 
 # The relative path from the project root to the directory where Drupal VM is located.
 ENV['DRUPALVM_DIR'] = "vendor/geerlingguy/drupal-vm"
@@ -18,4 +18,3 @@ Vagrant.configure('2') do |config|
     config.exec.commands '*', directory: "/var/www/${project.machine_name}"
   end
 end
-
--- a/src/Drush/Command/BltDoctorCommand.php
+++ b/src/Drush/Command/BltDoctorCommand.php
@@ -8,6 +8,7 @@ use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Output\ConsoleOutput;
 use Symfony\Component\Yaml\Yaml;
 use Drupal\Core\Installer\Exception\AlreadyInstalledException;
+use Acquia\Blt\Robo\Commands\Vm\VmCommand;
 
 /**
  *
@@ -709,8 +710,9 @@ class BltDoctor {
   protected function checkDrupalVm() {
     if ($this->drupalVmEnabled) {
       $passed = TRUE;
-      if (!file_exists($this->repoRoot . '/box/config.yml')) {
-        $this->logOutcome(__FUNCTION__ . ':init', "You have DrupalVM initialized, but box/config.yml is missing.", 'error');
+      $drupal_vm_config = $this->getDrupalVmConfigFile();
+      if (!file_exists($this->repoRoot . $drupal_vm_config)) {
+        $this->logOutcome(__FUNCTION__ . ':init', "You have DrupalVM initialized, but $drupal_vm_config is missing.", 'error');
 
         $passed = FALSE;
       }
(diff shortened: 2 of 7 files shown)


dpagini:
Cool, so looks like this passes. Let me know if you would like me to investigate putting tests in anyways. If so, do you have any idea how I interact with the `askDefault()` method in a test?

Any other suggestions or requested changes, let me know. Thanks!

dpagini:
Hmm, before this goes in I think I need to adjust a few more files...
* `src/Drush/Command/BltDoctorCommand.php`
* `src/Robo/Inspector/Inspector.php`
* `src/Robo/Commands/Blt/DoctorCommand.php`

Let me know any thoughts on this one if this isn't an appropriate direction....

src/Drush/Command/BltDoctorCommand.php
```diff
@@ -199,7 +200,7 @@ protected function setStacksEnabled() {
     }
 
     if (file_exists($this->repoRoot . '/Vagrantfile')
-      && file_exists($this->repoRoot . '/project.local.yml')
+      && file_exists($this->repoRoot . '/blt/project.local.yml')
```
dpagini: I think this was a bug? Fixing the place it's looking for the `project.local.yml` file.

dpagini:
One note, I was having trouble with [this line](https://github.com/acquia/blt/blob/8.x/src/Drush/Command/BltDoctorCommand.php#L734) which detects the remote host from the drush alias file. I don't believe this is related to my change here though.

dpagini:
Saw the changes in #1767 and I think I'll need to refactor this, or that might need a few tweaks. This variable I'm adding to `build.yml` might make more sense as a default setting under the `vm` key... thoughts?

src/Drush/Command/BltDoctorCommand.php
grasmash: Looks like it!

grasmash:
Quick note... I like this PR, but I'm going to wait to merge it until after 8.9.0. Trying not to introduce new features at the moment.

dpagini:
Cool, thanks! I'll keep an eye on it. I can make some updates after the 1767 PR goes in, and I can use the github patch file of this PR to upgrade our site to 8.9.0 when it's released. 

J/w, what's your schedule looking like for 8.9.0?

src/Drush/Command/BltDoctorCommand.php
dpagini: I'm going to submit another quick/small PR to update this line if you want to include it.

dpagini:
@grasmash do you have a preference where this config key lives? `vm.config` or `blt.config-files.drupal-vm`? or another place?

Files updated and rebased on 8.9.0-rc2

src/Drush/Command/BltDoctorCommand.php
```diff
@@ -770,10 +772,20 @@ protected function checkDrupalVm() {
   }
 
   /**
+   * @return string
+   */
+  protected function getDrupalVmConfigFile() {
```
grasmash: I don't think we need this method. You can just call `$this->getConfigValue('blt.config-files.drupal-vm');`

scripts/drupal-vm/Vagrantfile
```diff
@@ -4,7 +4,7 @@ ENV['DRUPALVM_PROJECT_ROOT'] = "#{__dir__}"
 
 # The relative path from the project root to the config directory where you
 # placed your config.yml file.
-ENV['DRUPALVM_CONFIG_DIR'] = "box"
+ENV['DRUPALVM_CONFIG_DIR'] = "${drupalvm.config.dir}"
```
grasmash: Shouldn't this be `blt.config-files.drupal-vm` ?

dpagini: So this needs to be a relative path, and that config value (`blt.config-files.dir`) is the absolute path when calculated, I believe. I create the relative path in the `Vagrantfile` copy command, and set this variable just for the purposes of this command.

dpagini:
I'm sort of leaning towards a `vm.config` config key (now that we have the vm section) instead of `blt.config-files.drupal-vm`. I think it would be a very small update here. Does that make more sense?

src/Robo/Commands/Vm/VmCommand.php
```diff
@@ -118,6 +124,31 @@ protected function install() {
    * @command vm:config
    */
   public function config() {
+    $drupalvm_install_dir = $this->askDefault("Where would you like to install Drupal-VM files?", $this->vmConfigDir);
```
grasmash: I'm fine with allowing people to overwrite the location of the VM, but I'd rather not prompt them to do it. There should be a bit more ownership of the customization IMO.

dpagini: Hmm... That's fair, I'll put up another PR with an alternate treatment. Let me know what you like better...?
The only reason I would disagree with the prompt point is that the VM command already prompts the user to boot the VM. It's also a specific command that is run by a single user after the BLT project has already been set up. It's not like this would ever be run non-interactively, right?

PR #1762, opened 2017-07-05
Removing patch from core that no longer applies.

grasmash:


danepowell:
It seems like core is moving away from the strategy used by this patch (prefixing Twig caches with the deployment identifier), and instead adding a method that hosting companies will have to override to clear twig caches across web heads. While that might be a better solution long-term, it unfortunately leaves existing sites in the lurch until that process plays out.

I'd propose that we update and maintain the old-style patch ourselves until this gets completely sorted out. Or else put a really big disclaimer in the release notes that Twig caches will start breaking again.

bkosborne:
Damn this just bit me as I was trying to upgrade D8 versions. I'm more comfortable using the original patch until the more recent one has been tested more and committed. I think my only recourse is to fork BLT to remove the patch, since I'm having trouble getting composer patches to ignore the patch defined in BLT.

bkosborne:
Ok I was able to do this. In my project's composer.json, I had to had to do this:

```
        "merge-plugin": {
            "require": [
                "vendor/acquia/blt/composer.required.json",
                "blt/composer.overrides.json"
            ],
            "merge-extra": true,
            "merge-extra-deep": true,
            "merge-scripts": true,
            "replace": true
        }
```

And in the `composer.overrides.json` file add this:

```
{
  "extra": {
    "enable-patching": true,
    "patches": {
      "drupal/core": {
        "Clear Twig caches on deploys": "patches/no_reliable_method-2752961-29.patch"
      }
    }
  }
}
```

Which is a reference to that original patch, but now stored locally in my project and re-rolled so it applies cleanly.

Note that that template for composer.json that BLT provides originally had things setup like this:

```
        "merge-plugin": {
            "require": [
                "vendor/acquia/blt/composer.required.json"
            ],
            "include": [
                "blt/composer.overrides.json"
            ],
            "merge-extra": true,
            "merge-extra-deep": true,
            "merge-scripts": true,
            "replace": true
        }
```

Which wasn't working. I guess because it includes the the "include" stuff before the "require" stuff?

PR #1782, opened 2017-07-08
Updating settings command.

dpagini:
Expands changes made in #1765. 
If `docroot/sites/default/default.settings.php` does not exist, don't try to copy it to the current multisite `default.settings.php`, and also don't try to copy it to current multisite `settings.php`.

### DIFF
--- a/src/Robo/Commands/Setup/SettingsCommand.php
+++ b/src/Robo/Commands/Setup/SettingsCommand.php
@@ -66,7 +66,6 @@ class SettingsCommand extends BltTasks {
       $project_local_drush_file = "$multisite_dir/local.drushrc.php";
 
       $copy_map = [
-        $default_project_default_settings_file => $project_default_settings_file,
         $blt_local_settings_file => $default_local_settings_file,
         $default_local_settings_file => $project_local_settings_file,
         $blt_local_drush_file => $default_local_drush_file,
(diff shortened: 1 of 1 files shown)


src/Robo/Commands/Setup/SettingsCommand.php
```diff
@@ -66,17 +66,20 @@ public function generateSiteConfigFiles() {
...
       ];
 
       // Only add the settings file if the default exists.
-      if (file_exists($project_default_settings_file)) {
+      if (file_exists($default_project_default_settings_file)) {
+        $copy_map[$default_project_default_settings_file] = $project_default_settings_file;
         $copy_map[$project_default_settings_file] = $project_settings_file;
       }
+      else {
+        $this->yell("No $default_project_default_settings_file file found.");
```
dpagini: I added this "yell" just to make sure the user is aware that an expected file is missing.

grasmash: I'd prefer to use `$this->logger>warning()` rather than `$this->yell()`.

dpagini: Sure thing, let me make that update.

dpagini: Updated ✅

PR #1783, opened 2017-07-10
Moving ExampleTest.php to examples:init command.

dpagini:
SSIA.
Moving the example PHPUnit test to the `examples:init` command. Not sure if ExampleTest.php needs to be in the PHPUnit folder to function, but I thought this made some sense.


grasmash:
Mine as well do Behat tests too. But, will people know to do this? I'm a little concerned that people will miss it.

dpagini:
Updated to move `Examples.feature` as well into `$ blt examples.init` command.

> But, will people know to do this? I'm a little concerned that people will miss it.

I saw that the `examples:init` command gets run on new project creation, so at that point it's up to the user if they want to keep the files or not, and they have a command to get them back. Is there anyway to detect a new install of BLT (vs. an upgrade)? If so, it could be run then as well...?

Anyways, the only reason I posted the PR is I just saw the command already existed, and these files seemed like they belonged in the same grouping. I don't have my heart set on this one.  =)  Up to you...

PR #1801, opened 2017-07-14
Normalize and covert svg line endings to native on checkout

zweishar:
Changes proposed:
- Normalize svg line endings on checkout

There is no `.gitattributes` entry for `.svg` files, which means we're relying on what ever the value of `core.autocrlf` is on each developers machine. I propose that we add a new entry to `template/.gitattributes` to set this on a per-repository basis.

In my opinion, `.svg` files should be treated as text so that their line endings will be converted to native on checkout. This makes sense to me because `.svg` files can be edited manually, and they are not binary files. If others disagree I'm happy to hear counter arguments.

### DIFF
--- a/template/.gitattributes
+++ b/template/.gitattributes
@@ -40,6 +40,7 @@
 *.txt     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 *.xml     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 *.yml     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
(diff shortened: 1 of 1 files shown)


greylabel:
Makes sense to me. Drupal core includes an entry for `.svg` as well -- http://cgit.drupalcode.org/drupal/tree/.gitattributes?h=8.3.x#n38

template/.gitattributes
```diff
@@ -40,6 +40,7 @@
 *.txt     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 *.xml     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
 *.yml     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
+*.svg     text eol=lf whitespace=blank-at-eol,-blank-at-eof,-space-before-tab,tab-in-indent,tabwidth=2
```
greylabel: This list should be kept in alphabetical order, though.

PR #1817, opened 2017-07-18
Allow Drupal-VM config flexibility.

dpagini:
Alternate approach to 1754. If this goes in, this closes #1754.

Two main changes here...
* I used `vm.config` as my config key as opposed to `blt.config-files.drupal-vm`. This alternate key feels like a more appropriate place to me after the `vm` key was created in build.yml.
* The `vm:config` command no longer prompts the user for an installation path for Drupal-VM, per request of @grasmash in 1754. The decision to use an alternate installation path for Drupal-VM becomes slightly more manual, but the flexibility still exists to move it, which is the key for me.

_Just FYI, I created this new PR as I'm referencing 1754.patch directly in my project, and this change in approach will require me to update my project._

Similar to comments in the other PR, the one thing I'm not 100% confident in the changes to `post-provision.sh` script. I believe referencing the alias script directly at `/vagrant/vendor/.../alias` works as intended, I'm just not sure if there was reason this didn't work originally (maybe other OS or something?).

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -210,3 +210,4 @@ validate:
 
 vm:
   enable: false
+  config: ${repo.root}/box/config.yml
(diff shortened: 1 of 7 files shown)


dpagini:
Thanks @grasmash!

PR #1821, opened 2017-07-20
Autoload commands from Robo plugins.

grasmash:


grasmash:
Need to add update hooks to change namespaces in existing custom hooks.

danepowell:
Does this need to be reopened against 9.1.x, since it's an enhancement?

danepowell:
Looks like this is blocked on upstream issue: https://github.com/consolidation/Robo/pull/604

PR #1829, opened 2017-07-24
Prevent doctrine/inflector from advancing to 1.2.x.

timcosgrove:
Changes proposed:
- Fix doctrine/inflector on 1.1.x, as 1.2.x requires PHP7

As of 1.2.0 (https://github.com/doctrine/inflector/commits/v1.2.0) doctrine/inflector requires 7.0, rather than 5.6|7.0. 

As Acquia has not yet rolled PHP7 out to production environments yet, we probably want to hold ourselves back.

See similar issue in for Laravel: https://github.com/laravel/framework/issues/20214

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
         "composer/installers":              "^1.2.0",
         "composer-plugin-api":              "^1.0.0",
         "doctrine/common":                  "^2.5",
+        "doctrine/inflector":               "~1.1.0",
         "oomphinc/composer-installers-extender": "^1.1",
         "dflydev/dot-access-data":          "^1.1.0",
         "drupal/coder":                     "^8.2.11",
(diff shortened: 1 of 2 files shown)


timcosgrove:
Apologies. Getting myself set up w/ proper BLT development locally.

timcosgrove:
Added updated composer.lock, updated w/ `composer update doctrine/inflector`

mikemadison13:
FYI I posted about this over on the acquia/lightning space as well: https://github.com/acquia/lightning/issues/411

timcosgrove:
Yeah. On one hand, this can be resolved by using PHP 5.6 where you run `composer update`. On the other hand, I would imagine that many people who run BLT at all run Vagrant or Docker locally, which means they can run 5.6 in their vm/image and run whatever locally. You can take the trouble to run `composer update` in your VM at that point. 

We'll probably just ask all our engineers to run 5.6 locally, but I'll leave this PR open for @grasmash to decline or to not decline.

PR #1846, opened 2017-07-28
Fixes #1835: provide --exclude-paths configuration key.

### LINKED ISSUE #1835: Allow --exclude-paths to be configured for sync:files
My system information:
* BLT version: 8.9.0

When I run `blt sync:files` I have no option to override the `--exclude-paths` option. For a particular project, `sites/$site_dir/files/styles` is used for image styles instead of css styles and the only way to sync them using BLT is to replace the command entirely.

If an array could be added to `project.yml`, `implode()` could be used to pass this configuration to the drush task. I'm happy to submit a PR to this effect, but would like to take some input on what this key should be called and if it should be made a sub-key of some existing key (possibly `sync`).

bobbygryzynger:
Fixes #1835.

Changes proposed:
- allow configuration of --exclude-paths

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -144,6 +144,11 @@ sync:
   # Set this value to 'true' or pass -Dsync.files=true
   # to override this behavior.
   files: false
+  # Paths to exclude during file syncing operations.
+  exclude-paths:
+    - styles
+    - css
+    - js
 
 # Custom tasks that are triggered at pre-defined times in the build process.
 # The tasks hook into BLT's existing commands.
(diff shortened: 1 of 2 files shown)


alanoakden:
@bobbygryzynger Thanks for this, can you explain exactly where this should get configured?

alanoakden:
> @bobbygryzynger Thanks for this, can you explain exactly where this should get configured?

Ideally devs could override this by adding entries under sync: in their `blt.yml`

For now, I had to directly edit the `build.yml` file in the BLT vendor directory.

Furthermore, if you provide no exclude-paths config and comment all of that out, then the command errors as it expects some kind of exclude paths.

Therefore, I actually removed `->option('exclude-paths', implode(':', $this->getConfigValue('sync.exclude-paths'))); ` from `vendor/acquia/blt/src/Robo/Commands/Drupal/SyncCommand.php ` in order to sync all CSS and JS files that I needed.

PR #1869, opened 2017-08-01
Acsf init dev

msherron:
Fixes # 1868.

Changes proposed:
- Adds composer:require command
- Adds acquia/acsf-tools to composer in acsf:init command


grasmash:
Would you mind applying this new command to:

- https://github.com/acquia/blt/blob/8.x/src/Robo/Commands/Vm/VmCommand.php#L234
-
 https://github.com/acquia/blt/blob/8.x/src/Robo/Commands/Saml/SimpleSamlPhpCommand.php#L56

msherron:
Can do, will update 👍

PR #1878, opened 2017-08-03
Issue #1874: Fixed VM commands requiring interactive shell

### LINKED ISSUE #1874: sync:files fails due to non-interactive shell
My system information:
* Operating system version: 17.04
* BLT version: 8.9

Output of `blt doctor`:
all green

When I run this command:
```
blt sync:files
```

I get the following output:
```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/foo/vendor/bin/drush rsync '@foo.test:%files' /var/www/foo/docroot/sites/default/files --exclude-paths='styles:css:js' in /var/www/foo/docroot
You will delete files in /var/www/foo/docroot/sites/default/files and replace with data from foo.test@foo:/var/www/html/foo.test/docroot/sites/default/files
Do you really want to continue? (y/n): Cancelled.
[Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 75  Time 2.775s
```

Files should be synced

Like the litany of recent issues related to sync:refresh more generally, this seems to be due to Robo running the commands in a non-interactive shell.

danepowell:
Fixes #1874

### DIFF
--- a/src/Robo/Commands/Sync/DbCommand.php
+++ b/src/Robo/Commands/Sync/DbCommand.php
@@ -51,7 +51,9 @@ class DbCommand extends BltTasks {
    * Copies remote db to local db for default site.
    *
    * @command sync:db
-   * @executeInDrupalVm
+   *
+   * This command does not use @executeInDrupalVm because it
+   * requires an interactive shell.
    */
   public function syncDbDefault() {
     $this->invokeCommand('setup:settings');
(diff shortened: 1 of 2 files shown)


danepowell:
There's slightly more to this than this actually... and it might be better to figure out why vagrant-exec is non-interactive

PR #1890, opened 2017-08-07
BLT-1888: defining platform as PHP 5.6 by default.

### LINKED ISSUE #1888: Provide Default Platform Configuration in Composer
Currently the composer configuration in BLT is based on the host machine. This can be problematic for newer systems (running PHP 7) or systems that rely on their VMs to match production / cloud configuration.

I would propose that we add default platform settings to the composer.required.json file so that BLT defaults all composer dependencies to PHP 5.6. See: https://getcomposer.org/doc/06-config.md#platform

```
"config": {
  "platform": {
    "php": "5.6"
  }
},
```

mikemadison13:
Fixes #1888 .

Changes proposed:
- defines platform as php 5.6 by default
-

### DIFF
--- a/composer.required.json
+++ b/composer.required.json
@@ -50,6 +50,11 @@
       }
     }
   },
+  "config": {
+    "platform": {
+      "php": "5.6"
+    }
+  },
   "scripts": {
     "blt-alias": "blt install-alias -y --ansi",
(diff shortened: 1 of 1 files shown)


grasmash:
Replaced by #1901.

PR #1901, opened 2017-08-08
Fixes #1888: Default to PHP 7.1.

### LINKED ISSUE #1888: Provide Default Platform Configuration in Composer

grasmash:


grasmash:
@danepowell @geerlingguy @mikemadison13 Will this have any notable consequences that I should be aware of?

composer.required.json
```diff
@@ -35,6 +35,11 @@
       "Acquia\\Blt\\Custom\\": "../../../blt/src/"
     }
   },
+  "config": {
+    "platform": {
+      "php": "7"
```
danepowell: Won't this prevent BLT from being installed or run on host machines using PHP <7?

danepowell:
We should also set up Pipelines to use PHP7:
````
version: 1.1.0
services:
  - mysql
  - php:
      version: 7.1
````

We should also make sure to include release notes indicating that the default versions have changed, and folks should re-initialize DrupalVM, Travis, and Pipelines to get the change locally.

mikemadison13:
People will def. need to change back to 5.6 if not running 7. That will cause problems for sure.

composer.required.json
dpagini: Curious about this answer...? Also, if this is the general approach of BLT, can't we undo #1829 as well?

Issue #1816, opened 2017-07-18
Provide better multisite support for ACSF locally


japerry:
As a site developer, I want to be able to use multiple ACSF sites on my local development environment.

Earlier this year, multisite had a key within project.yml that allows us to identify different domains. However, when 'autodetect multisite' came around, there was an assumption that every multisite had its own folder. In 8.9, the multisites array was changed a bit, and the site variable no specifies the site you're on. 

Since ACSF hosts all sites in the 'g' folder, this approach doesn't quite work. You can use the 'default' folder with some tweaking to make it work with multiple databases for local development. Using the 'g' folder locally probably doesn't work because it requires some functions that are only available on ACSF cloud servers. 

Looking through what I know (and there is probably a few things I'm missing here), I'm attempting to bring ACSF support locally:

### Proposal
- [ ] Add an acsf flag in project.yml that indicates to the multisite functions not to look at different directories other than default for sites
- [ ] Add each site in project.yml that exist in ACSF.
- [ ] Change local.settings.php to dynamically load a database based on the host you're trying to access
- [x] Change drupal-vm config to allow for multiple mysql databases (#1916)
- [ ] Update drupal-vm and acsf drush aliases to not only connect to multiple sites locally, but sync with ACSF sites on cloud.
- [x] Decide folder strategy to contain config per site, using config split.
- [x] Update all of the robo commands that utilize multisite so drush can properly execute based on the current site uri.

msherron:
I think this makes sense, but wanted to add the following to the conversation:

1) I think we have two common use cases in ACSF, one where there are only a handful of sites (1-10), similar to traditional Drupal multisite. For these builds, I think it makes sense to to maintain folders for each site, and to be able to host multiple databases locally. The other case is for larger 'platform' builds that contain potentially hundreds of sites - In this case, maintaining folders for every site isn't possible/desirable. I think whatever changes we make to BLT need to keep both use cases in mind, and not limit the functionality of either.
2) I'm already handling dynamic updating of drush aliases in acsf-tools, which we'll shortly be adding to BLT in the acsf:init command - so that work is largely already handled here: https://github.com/acquia/acsf-tools/blob/8.x-dev/acsf_tools_generate_aliases.drush.inc - Feedback/PRs welcome!
3) Can we point to examples of config split working well on larger platform builds yet? I think it makes sense for smaller multisite builds, but I think we should continue to avoid proscribing a specific ACSF config management strategy, and maintain flexibility.

cc: @japerry

[...]

PR #1916, opened 2017-08-10
Improve support for multisites in DrupalVM.

danepowell:
This is the first of many PRs to support #1816

This just changes the default DrupalVM config file to be in a format that's more easily extensible for multisite projects, and adds related documentation.

This should have no effect on existing single-site installs, and doesn't require projects to nuke and rebuild their vm (unless they want to take advantage of local multisites, in which case they can just reprovision).


danepowell:
I did a more thorough review of the documentation and updated it where appropriate.

PR #1920, opened 2017-08-11
Sync db and files to ODEs automatically.

danepowell:
Currently, BLT doesn't do anything to enable CD on-demand environment (commonly "ODE") integration.

You can add these lines to your acquia-pipelines.yml:

            - deploy:
            script:
              - pipelines-deploy

... but you'll just get an empty environment (code but no db or files).

I think most projects would want to populate new ODEs with a database and files (based on `drush.aliases.remote`) and run updates.

@grasmash should I also add those deploy lines to BLT's default acquia-pipelines.yml, so that users of BLT + CD/Pipelines get ODEs for new PRs out of the box?

### DIFF
--- a/scripts/cloud-hooks/functions.sh
+++ b/scripts/cloud-hooks/functions.sh
@@ -13,7 +13,7 @@ deploy_updates() {
       ;;
     01devup|01testup|01update)
       ;;
-    ode[1-10])
+    ode[1-9])
       deploy_sync
       ;;
     *)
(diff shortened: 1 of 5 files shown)


grasmash:
Yes.

danepowell:
Well, this is unfortunate. I think the code here is all good, but it appears that Pipelines containers don't have the SSH keys necessary to connect to other ACE environments and download the DB. That seems like a showstopper unless you have any other ideas.

danepowell:
For now I've changed this to do fresh installs instead of sync, which is the only thing that seems possible at the moment. I've tested it and it works great.

Up to you whether you want to merge this as-is and finish adding sync support later, or hold our breath waiting for sync support.

PR #1921, opened 2017-08-12
Fixed failing npm installs in Pipelines.

danepowell:
If you run `npm install` on a vanilla BLT + Cog install in Pipelines, it will fail about 80% of the time. The problem seems to be that older versions of npm are less memory/cpu efficient and exhaust some sort of container resource, causing the process to be killed.

Upgrading to a later version of npm seems to be an adequate workaround.

### DIFF
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -16,6 +16,8 @@ events:
               - composer validate --no-check-all --ansi
               - composer install --ansi
               - source ${BLT_DIR}/scripts/pipelines/setup_env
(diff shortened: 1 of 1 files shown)


grasmash:
Should we have this performed in a frontend-setup target hook instead?

danepowell:
@grasmash my instinct is that BLT tasks (such as frontend:setup) should not do global installations... that could lead to nasty surprises. A recent version of NPM is simply a prerequisite for for most projects, just like PHP or Node in general, therefore it should be installed on a per-environment basis prior to BLT even getting involved.

PR #1930, opened 2017-08-17
Allow config override for sync:refresh.

thom8:
Changes proposed:

Allow you to override the specific steps of `sync:refresh`.

Would be nice to be able to do this for any command which calls subcommands.

### DIFF
--- a/src/Robo/Commands/Sync/RefreshCommand.php
+++ b/src/Robo/Commands/Sync/RefreshCommand.php
@@ -56,12 +56,7 @@ class RefreshCommand extends BltTasks {
    * @command sync:refresh
    */
   public function refreshDefault() {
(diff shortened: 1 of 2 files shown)


src/Robo/Commands/Sync/RefreshCommand.php
```diff
@@ -56,12 +56,7 @@ protected function refreshMultisite($multisite_name) {
    * @command sync:refresh
```
grasmash: Funny, I had this exact same idea previously and even began to implement it. I was thinking that each top level command, e.g. setup, tests, sync,  could be structured this way.

We should probably make an alias for this command that is simply `sync`,  so that it corresponds with having configuration that lives under the key `sync.commands`.

PR #1946, opened 2017-08-23
Adding generate:multisite command.

grasmash:


grasmash:
The purpose of this command is to allow easy creation of new multisite and a corresponding site.yml with the correct values.

It's a WIP that needs some help getting over the finishline.

danepowell:
This needs to be reopened against 9.1.x right?

grasmash:
Yes.

PR #1967, opened 2017-09-06
Use the latest chromedriver.

ericchenshine:


SaiPriyanka:
This issue is not fixed in latest chromeDriver version 2.32

ericchenshine:
@SaiPriyanka could you provide more details or error message, please?

SaiPriyanka:
I am refering to this issue https://github.com/acquia/blt/issues/1968
The exception we get is 
unknown error: Element is not clickable at point"

After chromedriver update to 2.32 code is not autoscrolling through pages

ericchenshine:
@SaiPriyanka could you provide more details including BLT version, chrome and chromedriver version in the error log, please? Are you also get the error in the CI? Note that this fix is only available in BLT 8.9.3 or above.

PR #1982, opened 2017-09-12
Fixes #1816: Provide better multisite support for ACSF locally.

### LINKED ISSUE #1816: Provide better multisite support for ACSF locally

grasmash:


settings/acsf.local.settings.php
```diff
@@ -0,0 +1,32 @@
...
+            'password' => 'drupal',
+            'host' => 'localhost',
+            'port' => '3306',
+            'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
+            'driver' => 'mysql',
+            'prefix' => '',
+          ),
+      ),
+  );
+}
```
japerry: It'd be good to also add dynamic files and config folder paths here:

   $config_directories['sync'] = $dir . '/config/' . $name . '/local';
+
+    /**
+     * Private file path.
+     */
+    $settings['file_private_path'] = $dir . "/files-private/${name}";
+
+    /**
+     * Public file path.
+     */
+    // Set public filesystem per local development site.
+    $settings['file_public_path'] = "sites/default/files/${name}";
+  }

grasmash: I've updated the pull request to set the file paths dynamically. However, I'm not convinced that the configuration directory structure that you suggested makes sense as a default.

My thinking is that most applications will want to use the following configuration directory structure:
- config
    - default:  default configuration, shared by all sites
    - envs:
        - local:  local environment configuration, shared by all sites
    - site1:  configuration for docroot/sites/site1
    - site2: configuration for docroot/sites/site2

It would still be possible for each site to have its own specialized local configuration, but that seems overly granular as a default. Nothing prohibits you from taking that approach by setting custom configuration directories in your applications settings.php file. 

At present, BLT will automatically set the status of the site1 configuration split to `true` when the normal multisite approach is being used (I.E., non ACSF). do we want these site-specific splits to be enabled automatically in non-local environments?  I would think that having a bit of custom logic in your settings.php file to activate splits for site factory sites would be the safest approach here.

lcatlett: I would err on the side of enabling the splits on ACSF or providing a task that generates the needed factory hook to encourage / enforce config management best practices which is typically a pain point on ACSF and difficult to debug locally.  With improved multisite and config split support in BLT it will be much easier to maintain environment, site, and profile-specific layers of configuration on ACSF so I would encourage enabling the splits from the onset and making the decision to disable them a conscious one by the developer.

grasmash: @lcatlett  How would you suggest determining which split should be enabled on the site factory environment? In the local environment, we are using the host name. Do you think that that is sufficient?

grasmash:
dump of environmental variables insight factory environment:

```
rogue1@web-1577:/mnt/gfs/home/rogue1$ printenv
AH_CURRENT_REGION=us-east-1
XDG_SESSION_ID=165220
SHELL=/bin/bash
TERM=xterm-256color
HISTSIZE=10000000
TMPDIR=/mnt/tmp/rogue101live
SSH_TTY=/dev/pts/1
USER=rogue1
TEMP=/mnt/tmp/rogue101live
ACQUIA_HOSTING_DRUPAL_LOG=/var/log/sites/rogue1.01live/logs/web-1577/drupal-requests.log
AH_SITE_NAME=rogue101live
AH_REALM=enterprise-g1
PATH=/usr/local/php5.6/bin:/usr/local/drush8:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
MAIL=/var/mail/rogue1.01live
AH_NON_PRODUCTION=1
PWD=/mnt/gfs/home/rogue1
LANG=en_US.UTF-8
AH_SITE_ENVIRONMENT=01live
HOME=/home/rogue1
SHLVL=2
LOGNAME=rogue1.01live
AH_SITE_GROUP=rogue1
XDG_RUNTIME_DIR=/run/user/10725
HISTTIMEFORMAT=%F %T
_=/usr/bin/printenv
```

settings/blt.settings.php
```diff
@@ -48,18 +55,54 @@
...
+  if ($is_local_env) {
+    // When developing locally, we use the host name to determine which site
+    // factory site is active. The hostname must have a corresponding entry
+    // under the acsf.sites key.
+    $input = new ArgvInput($_SERVER['argv']);
+    $config_initializer = new ConfigInitializer($repo_root, $input);
+    $config = $config_initializer->initialize();
+
+    // The hostname must match the pattern [sitename].local, where [sitename]
+    // is a value in the acsf.sites array.
```
grasmash: This feels pretty hacky and I would love to hear a better idea.

grasmash:
@japerry @lcatlett If you have any availability to test this code in a site factory environment, that would be appreciated. :-)

danepowell:
I resolved the merge conflicts and code formatting errors, but now there is another error. It seems maybe BLT has changed how it handles configuration objects since the last time this PR was updated.

> Error: Cannot use object of type Acquia\Blt\Robo\Config\DefaultConfig as array in /home/travis/build/acquia/blt-project/docroot/core/includes/errors.inc on line 319

settings/blt.settings.php
```diff
@@ -5,6 +5,8 @@
  * Setup BLT utility variables, include required files.
  */
 
+use Acquia\Blt\Robo\Config\ConfigInitializer;
```
alex-moreno: seems something is missing:

```Fatal error: Class 'Acquia\Blt\Robo\Config\ConfigInitializer' not found in /var/www/bayerphweb/vendor/acquia/blt/settings/blt.settings.php on line 107```

grasmash: This needs to be resubmitted to 9.1.x.

danepowell:
I took a crack at rebasing this in the linked PR

PR #1983, opened 2017-09-12
Remove xdebug warnings

danepowell:
On 8.9.2 using DrupalVM, any BLT command executed in the VM produces a warning about the xdebug extension being enabled.

I'm not even sure why BLT issues this warning. I looked through the Git logs and there's no clear reasoning for it. I'm assuming it's based on the old Composer warnings about xdebug impacting performance, but this is [no longer an issue with Composer](https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer).

So unless there's clear evidence of a performance impact, I'm thinking we should just remove the warning.

### DIFF
--- a/src/Robo/Inspector/Inspector.php
+++ b/src/Robo/Inspector/Inspector.php
@@ -685,7 +685,6 @@ class Inspector implements BuilderAwareInterface, ConfigAwareInterface, Containe
     if (!$this->warningsIssued) {
       $this->warnIfPhpOutdated();
       $this->warnIfDrupalVmNotRunning();
-      $this->warnIfXdebugLoaded();
       $this->warningsIssued = TRUE;
     }
   }
@@ -700,16 +699,6 @@ class Inspector implements BuilderAwareInterface, ConfigAwareInterface, Containe
     $current_php_version = phpversion();
     if ($current_php_version < $minimum_php_version) {
       throw new BltException("BLT requires PHP $minimum_php_version or greater. You are using $current_php_version.");
-    }
-  }
-
-  /**
-   * Warns the user if the xDebug extension is loaded.
-   */
-  protected function warnIfXdebugLoaded() {
-    $xdebug_loaded = extension_loaded('xdebug');
-    if ($xdebug_loaded) {
-      $this->logger->warning("The xDebug extension is loaded. This will significantly decrease performance.");
     }
   }
 


grasmash:
When asked the bug is enabled, everything that uses PHP goes much slower. That includes composer (in older versions) that it also includes Drupal itself.

I have found that some people running Drupal in a virtual machine experience excruciating performance issues. Disabling xdebug on both the host and the guest machine does have a significant performance impact.

danepowell:
So maybe we need to take a more nuanced approach to this. Just because xdebug is enabled in the CLI doesn't mean that it's enabled for Apache or FPM, and vice versa. The test that BLT runs is _only_ going to indicate whether xdebug is loaded for the CLI. And the only known performance impact for the CLI is Composer, which is a non-issue on recent versions. Maybe we should actually be testing for a recent composer version?

It just seems like bad UX to me that an out-of-the-box BLT + DrupalVM install throws all of these warnings any time you run a command. It seems like we should either determine that there really is a performance problem, and disable xdebug by default, or determine that there is no significant impact, and get rid of the warning.

grasmash:
> the only known performance impact for the CLI is Composer

That is not correct. There is a known performance impact on PHP generally. All PHP code,  regardless of Composer, Apache, FPM, etc., will have degraded performance. In some cases it is prohibitively impactful.

To quote [a StackOverflow article](https://stackoverflow.com/questions/3522182/will-enabling-xdebug-on-a-production-server-make-php-slower)

> when you add debugging behavior to your server, the debug engine "attaches" to the PHP process and receive messages of the engine to stop at breakpoints, and this is BAD, because introduces a high performance blow to have another process stopping or "retaining" the PHP parser.

The quick benchmark of `blt setup`:

With xdebug disabled:
```
real	4m22.620s
user	2m23.527s
sys	0m31.844s
```

With xdebug enabled:
```
real	6m59.522s
user	5m15.530s
sys	0m28.533s
```

In this example, execution of the command is 40% faster with xdebug disabled and, that is without using a virtual machine  which can potentially add a second instance of xdebug and further compound the performance issue.

danepowell:
Just to be clear, are those performance differences with xdebug _loaded_ or xdebug _enabled_? (It can be loaded without being enabled)

If it's okay to load xdebug without enabling it, then we just need to change the BLT inspection to check whether xdebug is actually enabled (it currently just checks if it's loaded).

If there's still a performance impact by loading xdebug (even without enabling it), then we probably need to open a PR for DrupalVM to allow xdebug to be completely unloaded for the CLI.

grasmash:
Loading the extension causes the performance impact, even if Xdebug is disabled.

Yes, I have often thought that we need a way to unload it in the virtual machine.

thom8:
here's some background -- https://github.com/geerlingguy/ansible-role-php/issues/140

danepowell:
I tested this myself, and got the same results as Matt (average of several runs of `blt setup`):
- Xdebug loaded, enabled: 2'9"
- Xdebug loaded, not enabled: 2'5"
- Xdebug unloaded: 1'26"

I opened an issue for DrupalVM: https://github.com/geerlingguy/ansible-role-php-xdebug/issues/53

Once that's resolved, BLT should adopt the new config option to completely unload xdebug for CLI.

@thom8 I see how you could use that to disable xdebug, but can you elaborate on that could be used to completely unload it (i.e. remove or comment out `/etc/php/5.6/cli/conf.d/20-xdebug.ini`?)

thom8:
@danepowell can be done with a custom task like https://github.com/geerlingguy/ansible-role-php-xdebug/issues/34#issuecomment-237420740

Just add a variable to a `when` case and set a default in BLT.

However, it would be nice if this task was provided upstream by the role

thom8:
Created upstream PR -- https://github.com/geerlingguy/ansible-role-php-xdebug/pull/54

danepowell:
Well killing the xdebug test is probably not the best path forward, so I'll close this and open an issue to track the changes to DrupalVM: https://github.com/acquia/blt/issues/2034

PR #2006, opened 2017-09-19
BLT-2005: updating pipelines schema version and including additional …

### LINKED ISSUE #2005: Pipelines Files Out of Date
The various pipeline files (templates, docs, etc.) are not as robust as they could be, and are using an old schema.

1. propose: embedding PHP 7 by default
2. propose: commenting out but including nvm and yarn as examples

mikemadison13:
…parameters in template.

Fixes #2005 2005.

Changes proposed:
- update pipelines schema version
- add additional parameters to the generated file

### DIFF
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -1,6 +1,8 @@
-version: 1.0.0
+version: 1.1.0
 services:
   - mysql
+  - php:
+      version: 7.1
 
 variables:
   global:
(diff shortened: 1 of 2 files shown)


scripts/pipelines/acquia-pipelines.yml
```diff
@@ -16,6 +18,8 @@ events:
               - composer validate --no-check-all --ansi
               - composer install --ansi
               - source ${BLT_DIR}/scripts/pipelines/setup_env
+              # - nvm install v6.9.1
```
grasmash: This seems like something better executed in your front end script.

mikemadison13: @grasmash my concern with that is that the local already has these things installed, where as the pipelines still does not. The examples I've seen from several other projects are doing a similar method to ^^. Not sure that everyone "will" but I couldn't find any good documentation on how to do it, so was trying to provide some basic examples.

PR #2032, opened 2017-09-25
Fixes #2026: Support profile installation from config.

### LINKED ISSUE #2026: Support profile installation from config
I'm really hoping that Lightning is about to [add a Drupal core patch](https://www.drupal.org/node/2908097) that would automatically import site configuration (from /config/default) at install time when installing a custom subprofile of Lightning. The awesome thing about this patch is that it seamlessly handles the differences in UUIDs between site config and profile config, so you no longer have to deal with all of the [related](https://github.com/acquia/lightning/pull/400) [muck](https://github.com/acquia/lightning/issues/387).

In order to support this, BLT just needs to stop passing the `--config-dir` argument to `drush si`, since Drupal with this patch will automatically import config. We no longer need Drush to do it (which nicely dodges this bug as well: https://github.com/drush-ops/drush/issues/2953)

We can either provide a configuration switch for users of Lightning (and other supported profiles) that turns `--config-dir` on or off, or we can just make this change across the board and document/mandate that this is the new best practice. Even if folks aren't using Lightning, they should be able to quickly patch their profile to support this. The only problem I foresee would be for users that have already patched core in a way that conflicts with the necessary upstream patches... these users would have to keep using `--config-dir`.

Note that neither the core patch or Lightning's incorporation of it are a sure thing yet, but I just want to be prepared if/when they do land.

grasmash:


grasmash:
This pull request will allow you to set `setup.drupal.install.import-config` to `false` and thereby prevent the `--config-dir` argument from being set during `drush si`. 

This will not be set automatically for anybody, but it does allow people to manually address this issue without any hackery.

We can implement a more sophisticated solution after the patch is added to lightning merged into core.

PR #2039, opened 2017-09-26
Fixing VM config property expansion.

grasmash:


timcosgrove:
As near as I can tell, this line was necessary: https://github.com/acquia/blt/pull/2039/files#diff-230747b7bce089cf23649b106a2ec653L285

Without it, in <repo>/Vagrantfile, you are left with:
`ENV['DRUPALVM_CONFIG_DIR'] = "${drupalvm.config.dir}"`

PR #2043, opened 2017-09-28
Add back vagrantfile property expansion removed in #2039

timcosgrove:
Fixes removal of https://github.com/acquia/blt/pull/2039/files#diff-230747b7bce089cf23649b106a2ec653L285

If this is in error, apologies for the noise and close the PR, but I'm pretty sure this wants to be there.


timcosgrove:
Error found organically, after running `composer create-project acquia/blt-project:8.x-dev --no-interaction MY_PROJECT` and going through setup as normal and wondering why BLT/drupalvm was not finding my machinename correctly.

PR #2055, opened 2017-09-29
Add additional settings file include options #2054

### LINKED ISSUE #2054: Add additional settings file include options
While BLT provides a lot of boilerplate config through its `*.settings.php` files, it would be nice to have some additional options to extend these with project customizations.

- [ ] Add support for including multiple global settings files.
- [ ] Add support for including a local global settings file.
- [ ] Add support for site-specific secrets files.

greylabel:
Fixes #2054.

Changes proposed:
- Add additional settings file include options.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -114,9 +114,14 @@ if ($is_ah_env) {
 
   // Store API Keys and things outside of version control.
   // @see settings/sample-secrets.settings.php for sample code.
-  $secrets_file = sprintf('/mnt/gfs/%s.%s/secrets.settings.php', $_ENV['AH_SITE_GROUP'], $_ENV['AH_SITE_ENVIRONMENT']);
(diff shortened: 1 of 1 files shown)


settings/blt.settings.php
```diff
@@ -229,7 +249,11 @@
  * Keep this code block at the end of this file to take full effect.
  */
 if ($is_local_env) {
-  // Load local machine settings.
+  // Load local settings for all sites.
+  if (file_exists(DRUPAL_ROOT . "/sites/settings/local.settings.php")) {
+    require DRUPAL_ROOT . "/sites/settings/local.settings.php";
```
greylabel: Should probably make sure this will not be added to deploy artifacts. Looks like these should handle it - https://github.com/acquia/blt/tree/8.x/scripts/blt/deploy

PR #2062, opened 2017-10-01
Issue 1543 twig lint replacement

### LINKED ISSUE #1543: Twig Lint is not maintained and fails valid twig code
My system information:
* Operating system type: OSX/drupal-vm
* Operating system version: Sierra/drupal-vm 4.3
* BLT version: 8.7

When I run this command:
```
blt validate:twig
```

I get the following output:
(Removed some project specific paths and file names)
```
blt > validate:twig:
     [echo] Validating twig syntax for all custom modules and themes...
KO in [path/to/]templates/content/node--[somecontenttype].html.twig (line 22)
   20           #}
   21           {% for image in node.field_product_offering_images %}
>> 22             {% set isRowEnd = loop.index is divisible by(4) %}
>> Unexpected token "name" of value "by" ("end of statement block" expected).
   23             {% set imageUrl = image.url %}
   24             {% set imageTitle = image.title %}
```

`divisible by(4)` is valid twig syntax and should not fail https://twig.sensiolabs.org/doc/1.x/tests/divisibleby.html

There's an issue on the twig lint project and a couple of PRs that haven't been updated since 2015 for this problem

https://github.com/asm89/twig-lint/issues/25
https://github.com/asm89/twig-lint/issues/33

PR that would address:
https://github.com/asm89/twig-lint/pull/23

Filing this as it looks like we should probably patch this somehow, or find another linter that is being more actively maintained.

wu-edward:
Fixes #1543.

Changes proposed:
- Remove asm89/twig-lint and use Symfony twig linter
- Run twig lint command procedurally instead of in parallel, otherwise error messages are suppressed.
- Account for Twig functions and filters added by Drupal, so that linter does not flag as invalid.
- Provide config for users to add custom Twig filters and functions that can be specified in project.yml

### DIFF
--- a/src/Robo/Blt.php
+++ b/src/Robo/Blt.php
@@ -21,9 +21,12 @@ use Robo\Common\ConfigAwareTrait;
 use Robo\Config\Config;
 use Robo\Robo;
 use Robo\Runner as RoboRunner;
+use Symfony\Bridge\Twig\Command\LintCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Twig\Environment;
+use Twig\Loader\FilesystemLoader;
 
 /**
  * The BLT Robo application.
@@ -78,6 +81,7 @@ class Blt implements ContainerAwareInterface, LoggerAwareInterface {
     $this->configureContainer($container);
     $this->addBuiltInCommandsAndHooks();
     $this->addPluginsCommandsAndHooks();
+    $this->addSymfonyCommands($application);
     $this->runner = new RoboRunner();
     $this->runner->setContainer($container);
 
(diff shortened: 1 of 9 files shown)


src/Robo/Blt.php
```diff
@@ -116,6 +120,60 @@ private function addPluginsCommandsAndHooks() {
...
   /**
+   * Adds Symfony (non-Robo) command classes to the application.
+   *
+   * @param \Acquia\Blt\Robo\Application $application
+   */
+  protected function addSymfonyCommands(Application $application) {
+    $twig = new Environment(new FilesystemLoader());
+
+    // Get any custom defined Twig filters to be ignored by linter.
+    $twig_filters = (array) $this->getConfig()->get('validate.twig.filters');
```
grasmash: Just curious, do you see an instance in which we will need to use this array? I see that it is currently empty.

src/Robo/Blt.php
```diff
@@ -116,6 +120,60 @@ private function addPluginsCommandsAndHooks() {
...
+    // Add Twig filters from Drupal TwigExtension to be ignored.
+    // @todo Find a way so that this list doesn't need to be maintained.
```
grasmash: Yeah, I think this is where I started to lose the thread. I wanted to somehow load `Drupal\Core\Template\TwigExtension` into the `$twig` environment, but that led down a rabbit hole of determining how to get the right services to inject into `TwigExtension`.

src/Robo/Blt.php
```diff
@@ -116,6 +120,60 @@ private function addPluginsCommandsAndHooks() {
...
+    $drupal_filters = [
```
grasmash: Nobody loves regular expressions, but you could load the text content of `docroot/core/lib/Drupal/Core/Template/TwigExtension.php` and then use a regex to identify the functions and filters:

`$found = preg_match_all("#new \\Twig_SimpleFunction\('([^']+)',#", file_get_contents('docroot/core/lib/Drupal/Core/Template/TwigExtension.php'), $matches);`

src/Robo/Blt.php
wu-edward: If developers add Twig functions or filters to their custom modules and use them in their twig templates, they'll need to add them to their project.yml, otherwise they'll fail validation. I suppose the empty array can be left out of build.yml, and the ability to add functions and filters documented somewhere instead?

src/Robo/Blt.php
wu-edward: Oh, that's not a bad idea. I'll try to update the PR soon.

wu-edward: Made the updates, and tested against canary.
I'm not sure why Travis is failing, though.

danepowell:
I'm restarting the test to be sure, but it looks like something about this is causing the pre-commit validators to not run correctly.

wu-edward:
@danepowell I think you're right. I haven't had much luck reproducing locally, but I'll look into it more.

src/Robo/Commands/Validate/TwigCommand.php
```diff
@@ -26,35 +22,11 @@ public function lintFileSets() {
...
-    $fileset_ids = $this->getConfigValue('validate.twig.filesets');
-    $filesets = $fileset_manager->getFilesets($fileset_ids);
-
-    $bin = $this->getConfigValue('composer.bin');
-    $command = "'$bin/twig-lint' lint --only-print-errors '%s'";
-    foreach ($filesets as $fileset_id => $fileset) {
-      $filesets[$fileset_id] = $fileset_manager->filterFilesByFileset($files, $fileset);
-    }
-
+    $command = "'$bin/blt' 'validate:twig:files' '%s'";
```
danepowell: Aren't you removing the `validate:twig:files` command above? Doesn't seem like it makes sense to call it here.

wu-edward: `validate:twig:files` gets added in `src/Robo/Blt.php` line 157.

wu-edward:
Created new PR against 9.1 #2151

PR #2063, opened 2017-10-01
Provide better Memcached boilerplate config.

### LINKED ISSUE #2027: [stub] Provide better Memcached boilerplate config and add DrupalVM support.
It would be nice to have better support, documentation, and configurations examples for using `memcached` and the [Memcache API and Integration](https://www.drupal.org/project/memcache) module for Drupal 8.

See:
* [Memcache API and Integration](https://www.drupal.org/project/memcache) module
* [Ansible Role - Memcached](https://github.com/geerlingguy/ansible-role-memcached#readme)
* [Using Memcached](https://docs.acquia.com/acquia-cloud/performance/memcached) - Acquia

Implementation (incomplete):
- [ ] Add `memcached` support in DrupalVM default config.

greylabel:
Fixes #2027 .

Changes proposed:
Provide better Memcached boilerplate config - see #2027 for details.


settings/memcache.settings.php
```diff
@@ -0,0 +1,42 @@
...
+  if ($modules && isset($modules['module']['memcache'])) {
+    // Use Memcached extension.
+    $memcached_exists = class_exists('Memcached', FALSE);
+
+    if ($memcached_exists) {
+      $settings['memcache']['extension'] = 'Memcached';
+    }
+
+    // Use memcache as the default bin.
+    $settings['cache']['default'] = 'cache.backend.memcache';
```
grasmash: Should this be contained within the $memcached_exists condition?

settings/memcache.settings.php
```diff
@@ -0,0 +1,42 @@
...
+ * @see: https://docs.acquia.com/cloud/performance/memcached
+ */
+
+use Drupal\Core\Config\BootstrapConfigStorageFactory;
+use Drupal\Core\Database\Database;
+
+if ($is_ah_env) {
+  // Note that this connects the database, so has to go after any
+  // $databases definitions. See: https://www.drupal.org/node/2766509
+  Database::setMultipleConnectionInfo($databases);
```
grasmash: Can this be removed after https://www.drupal.org/node/2766509 is resolved?

grasmash: Something about this line makes me nervous.  it feels like we should let core handle this, lest we accidentally change the database connection behavior. Do you feel that this is safe?

settings/memcache.settings.php
greylabel: Not really, this tells Drupal the use the cache backend provided by the Memacahe module. The `memcached_exists` check serves a different purpose -- it selects the PECL library to talk to the `memcached` daemon. PHP 5.6 on ACE is build with both the Memcache and Memcached libraries, while PHP 7.1 only includes Memcached, which is better maintained.

settings/memcache.settings.php
greylabel: In theory yes, although have not looked at the patches extensively and I am not sure [#2766509](https://www.drupal.org/node/2766509 ) takes care the case where we move locks to the memcache backend, specially  `$settings['container_yamls'][] = __DIR__ . '/memcache.yml';`

I share your concern and I would get a strong code smell reviewing this if I came across it. I mostly wanted to get your thoughts on the approach. I am not sure if there is a better/another way to check in `settings.php` if a module is enabled, which is what we need to do until the above mentioned issue of landed. Open to other suggestions. I have tested this code and it seems to work fine, but I have not tested extensively on ACE.

I think the other option is make memcache integration opt-in and just document how to configure it for environments that are known to have Drupal installed and the Memcache module enabled, and won't be reinstalled, e.g. stage and prod.

greylabel: I was just looking https://www.drupal.org/files/issues/gracefully_handle_a-2766509-49.patch and it looks like `cache.backend.null` is used in the case that the configured backend is provided by a module that is not installed yet. I think this means that performance could be potentially terrible if BLT assumes a site will use the Memcache module on ACE, but it is never actually enabled as cacheing would essentially be disabled. I am leaning more to the opt-in solution.

greylabel:
@grasmash let me know if you prefer this approach.

grasmash:
@greylabel I have merged this because memcache support is already opt-in. But,  why not automatically enable it on cloud environments and permit opting out? It seems like most of our customers will want to use this.

greylabel:
I think that will cause more problems than it will solves because of the
whole module-needs-to-be-enabled situation. My previous comments sum up my
concerns. Let me know if you are seeing something I missed.

PR #2064, opened 2017-10-01
Leverage "extra" to determine if BLT this is a "new" project

amcgowanca:
## Problem

Only project's whose `composer.json` specify the `"name"` property as `"acquia/blt-project"` can indicate a new project is being created for the first time.

## Proposed changes

Use the `"extra"` property in the following schema structure:

```
"extra": {
  "blt": {
    "new": true
  }
}
```

to identify whether or not BLT being installed and therefore the `internal:create-project` command should be invoked verse `internal:add-to-project`. When a project is being created for the first time and the `internal:add-to-project` is invoked, the following errors are received:

```
$ Creating BLT templated files...
 > /[project-dir]/vendor/acquia/blt/bin/blt internal:add-to-project --ansi -y
Installing new Composer dependencies provided by BLT. This make take a while...
PHP Fatal error:  Class 'Robo\Collection\CompletionWrapper' not found in /[project-dir]/vendor/consolidation/robo/src/Collection/Collection.php on line 89
```

This, I suspect (but have not confirmed in anyway), is the result of the autoloader not being ready for use.

## Additional changes

Additional changes to the [BLT Project](https://github.com/acquia/blt-project)'s `composer.json` would also be needed so that these changes would be effective for additional projects which may start from the `acquia/blt-project` repository.

### DIFF
--- a/src/Composer/Plugin.php
+++ b/src/Composer/Plugin.php
@@ -255,7 +255,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface {
    */
   protected function isNewProject() {
     $composer_json = json_decode(file_get_contents($this->getRepoRoot() . '/composer.json'), TRUE);
-    if (!empty($composer_json['name'] && $composer_json['name'] == 'acquia/blt-project')) {
+    if (isset($composer_json['extra']['blt']['new']) && $composer_json['extra']['blt']['new']) {
       return TRUE;
     }
     return FALSE;
--- a/src/Robo/Common/ComposerMunge.php
+++ b/src/Robo/Common/ComposerMunge.php
@@ -33,6 +33,12 @@ class ComposerMunge {
       $output['require-dev'] = (object) $output['require-dev'];
     }
 
+    // Remove the existence of the new project identifier from invoking the
+    // create-project again in subsequent update calls.
+    if (isset($output['extra']['blt']['new'])) {
+      unset($output['extra']['blt']['new']);
+    }
+
     $output_json = json_encode($output, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES);
 
     return $output_json;


grasmash:
Wouldn’t we need to remove this value during the installation process in order to avoid the application being identified as new during subsequent composer commands?

amcgowanca:
@grasmash You are right, totally overlooked that. Thoughts on location on where to handle that? My initial thoughts would be within the `Acquia\Blt\Robo\Common\ComposerMunge::munge`, however, I'm not sure if that is the best place for it. Only other alternatives that are quickly apparent would result in change in how the `ComposerMunge::munge` functions.

src/Composer/Plugin.php
```diff
@@ -255,7 +255,7 @@ protected function isInitialInstall() {
    */
   protected function isNewProject() {
     $composer_json = json_decode(file_get_contents($this->getRepoRoot() . '/composer.json'), TRUE);
-    if (!empty($composer_json['name'] && $composer_json['name'] == 'acquia/blt-project')) {
+    if (isset($composer_json['extra']['blt']['new']) && $composer_json['extra']['blt']['new']) {
```
grasmash: I think we actually need to check the existence of this value with array_key_exists() rather than isset().

grasmash: Merging this would require a corresponding change in `acquia/blt-project`. It may be better to use an OR operator here rather than replacing the condition.

src/Robo/Common/ComposerMunge.php
```diff
@@ -33,6 +33,12 @@ public static function munge($file1, $file2) {
       $output['require-dev'] = (object) $output['require-dev'];
     }
 
+    // Remove the existence of the new project identifier from invoking the
+    // create-project again in subsequent update calls.
+    if (isset($output['extra']['blt']['new'])) {
+      unset($output['extra']['blt']['new']);
```
grasmash: This doesn't really seem to belong inside this method. Perhaps we should move the file loading and writing functionality out of this method and simply accept and return an array.

We can then set this value and write to file here: https://github.com/acquia/blt/blob/d7c0e4616769c0a3c4a10cf9200491796d4faf30/src/Robo/Commands/Blt/UpdateCommand.php#L338

danepowell:
I changed the base branch to 9.1.x, since 8.x is no longer in use, but this now has merge conflicts that need to be resolved in addition to Matt's comments.

PR #2066, opened 2017-10-02
Removing sudo from Travis.

grasmash:


grasmash:
this was preventing deployments in BLT itself.

PR #1505, opened 2017-05-16
Refactoring git:pre-commit to use collection builder.

grasmash:

PR #2073, opened 2017-10-03
Refactor git:pre-commit to use collection builder.

danepowell:
This is a re-roll of #1505 against the latest 8.x.

Hopefully this is still desired? It looks like the pre-commit validation changed somewhat in the last few months, so that it's based on exceptions instead of exit codes. I'm not sure if that affects what you're trying to do here.


danepowell:
Reopening against 9.1.x since this is an enhancement

PR #2085, opened 2017-10-04
Swapping drush --no with --no-interaction.

grasmash:


src/Robo/Tasks/DrushTask.php
```diff
@@ -269,7 +269,7 @@ protected function setGlobalOptions() {
     }
 
     if (isset($this->assume) && is_bool($this->assume)) {
-      $assumption = $this->assume ? 'yes' : 'no';
+      $assumption = $this->assume ? 'yes' : 'no-interaction';
```
danepowell: @grasmash if my understanding is correct, there is no difference between `--yes` and `--no-interaction` in Drush 9. Both simply use whatever the "default response" is for any prompts.

There's no way to force either yes or no.

So I think we should just remove the "assume" option completely, and pass `--no-interaction` for all commands. Unless you want to have interactive commands as an option.

PR #2098, opened 2017-10-06
Adds a parsing function to ConfigInitializer to correctly parse robo config options.

timcosgrove:
I ran into a situation today where BLT 8.x was seemingly not loading site-specific config correctly. It seemed like later in the application cycle, `site` would be in the global config object as specified on the command line, but config defined at `sites/$site/site.yml` would not load.

I noticed I was passing `blt command -Dsite=mysite` originally, and that adding a space there i.e. `blt command -D site=mysite` would load config. I also noticed that `blt command --define=site=mysite` would not load config correctly.

After looking around, I traced my issue to ConfigInitializer. I saw that ConfigInitializer is using InputInterface->hasParameterOption() to find these commandline params, presumably because the global Robo config object isn't loaded yet. InputInterface was only parsing the parameters correctly accidentally, as it turns out, and only some of the time.

It seems Robo can understand commandline config parameters in 4 forms:
```
-D key=val # Parsed 'correctly' by ConfigInitializer
-Dkey=val # Not parsed
--define key=val # Parsed correctly
--define=key=val # Not parsed
```

InputInterface only parses the two above correctly because they're separated from their option flag by a space, and so it sees them as entirely separate items. In the second case above, the parameter as far as InputInterface is concerned is `-Dkey` and its value is `val`; in the fourth case, the parameter is `--define` and the value is `key=val`. 

I wrote a parsing function that attempts to handle a broader variety of cases. Very open to different approaches. However, it's important to my use case that site-specific config loads reliably.

Thx!


timcosgrove:
While I'm completely open to the test failure being my doing, the error seems to indicate the `--no` option is no longer valid in Drush:

```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /home/travis/build/acquia/blt-project/vendor/bin/drush cex sync --uri=default --no --config=/home/travis/build/acquia/blt-project/drush/drushrc.php in /home/travis/build/acquia/blt-project/docroot
                                                          
  [Symfony\Component\Console\Exception\RuntimeException]  
  The "--no" option does not exist.                                                                           
```

This test run used drush 9.0.0-beta6, which was tagged ~1 hour ago at this writing.

grasmash:
As for the site parameter loading incorrectly, I would take a look at the 10.x branch. That logic is repackaged significantly.

As for drush 9-beta6 issues, please take a look at BLTs 9.x branch. That is where address compatibility work is ongoing.

Lastly, as for the parsing issue, I was under the impression that only two of those forms should be correctly parsed by Robo. The old form (without a space) was specific to Phing.

danepowell:
The [Symfony Console](https://symfony.com/doc/current/components/console/console_arguments.html) is supposed to follow [docopt standards](http://docopt.org/), which say that the four forms listed in the OP should be accepted, including -Dfoo.

I'm not sure I understand what BLT or Robo is doing wrong to break that, but I do agree it should be fixed.

danepowell:
Also the 8.x branch is no longer used, we'd want to get this into 9.1.x first, which also fixes the Drush errors you observed. Although it looks like there's something wrong with this PR, it picked up a bunch of extra commits in the rebase.

PR #2099, opened 2017-10-06
Pass in the site variable rather than forcing default drush alias.

timcosgrove:
Commands passed to DrupalVM should not force the `@self` alias on the command, and they should pass the site variable so that the VM-internal command loads the correct config. 

This is dependent on https://github.com/acquia/blt/pull/2098 or some similar solution being merged, since currently site-specific config does not load correctly when passing `--define=site=example.com`.

### DIFF
--- a/src/Robo/Hooks/ExecuteInDrupalVmHook.php
+++ b/src/Robo/Hooks/ExecuteInDrupalVmHook.php
@@ -31,7 +31,7 @@ class ExecuteInDrupalVmHook extends BltTasks {
         $event->disableCommand();
         $command = $event->getCommand();
         $new_input = $this->createCommandInputFromCurrentParams($command, $event->getInput());
-        $new_input->setOption('define', 'drush.alias=self');
+        $new_input->setOption('define', 'site=' . $this->getConfigValue('site'));
 
         // We cannot return an exit code directly, because disabled commands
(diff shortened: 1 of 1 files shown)


src/Robo/Hooks/ExecuteInDrupalVmHook.php
```diff
@@ -31,7 +31,7 @@ public function executeInDrupalVm(ConsoleCommandEvent $event) {
         $event->disableCommand();
         $command = $event->getCommand();
         $new_input = $this->createCommandInputFromCurrentParams($command, $event->getInput());
-        $new_input->setOption('define', 'drush.alias=self');
```
grasmash: This is going to require a bit of testing. This is done because often the drush alias is not resolved directly inside of the virtual machine.

danepowell: I was concerned about this too. However, after some (very limited) testing, it appeared that local aliases worked just as well as `self` inside the VM. We should try to figure out why we decided to use self in the first place, and what the edge cases might be.

danepowell:
@timcosgrove we've moved all enhancements to the 9.1.x branch. Can you please open a new PR against 9.1.x? Sorry for the inconvenience.

PR #2126, opened 2017-10-12
Only sniff modified files.

dariusrestivan:
Fixes # .

Changes proposed:
- only sniff modified PHP files because working on bigger projects implies cleaning after other people

### DIFF
--- a/src/Robo/Commands/Git/GitCommand.php
+++ b/src/Robo/Commands/Git/GitCommand.php
@@ -47,16 +47,42 @@ class GitCommand extends BltTasks {
    * @return int
    */
   public function preCommitHook($changed_files) {
+
+
+    $changed_files_list = explode("\n", $changed_files);
+    $modified_files = [];
+
+    $extension_list = [
+      'module',
+      'install',
+      'inc',
+      'theme',
+      'profile',
+      'engine',
+      'php',
+    ];
+
+    foreach ($changed_files_list as $changed_file) {
+      $extension = explode('.', $changed_file);
+
+      if (in_array(array_pop($extension), $extension_list)) {
+        $php_files[] = $changed_file;
+      }
+    }
+
+    $php_files_list = implode("\n", $php_files);
+
+
+
     $this->invokeCommands([
       // Passing a file list to be PHPCS will cause all specified files to
       // be sniffed, regardless of the extensions or patterns defined in
       // phpcs.xml. So, we do not use validate:phpcs:files.
-      'validate:phpcs' => [],
+      'validate:phpcs:files' => ['file_list' => $php_files_list],
       'validate:twig:files' => ['file_list' => $changed_files],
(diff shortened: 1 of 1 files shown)


src/Robo/Commands/Git/GitCommand.php
```diff
@@ -47,16 +47,42 @@ public function commitMsgHook($message) {
    * @return int
    */
   public function preCommitHook($changed_files) {
+
+
+    $changed_files_list = explode("\n", $changed_files);
+    $modified_files = [];
+
+    $extension_list = [
+      'module',
```
grasmash: I understand the value of this, but it has a negative consequence of ignoring any customizations to PHPCS  configuration that a user may have made to their phpcs.xml file.  To avoid this, we would need to load that file and parse it to determine the correct extensions, patterns, and items to ignore.

dariusrestivan: I see, do you think you can go back to the way it's used to be before, when the PHPCS only sniffed the modified files? Is it worth investing more time in this?

grasmash: I would rather not go back to the way it was before because it required us to build and maintain our own system for managing code sniffer customizations. With the exception of this disadvantage, the current approach is much better.

I'd be open to a pull request to resolve this issue, but I did look into it myself for a bit and determined that it was not worth the time to develop.

The issue is that once you pass a list of files to phpcs, it then ignores all other settings (like include and exclude patterns).  this behavior is intentional and there's been some disagreement around it.

@see https://github.com/acquia/blt/pull/2126

> The ignoring of file extensions for specific [specified] files is a feature of PHP_CodeSniffer and is the only way to check files without an extension.

grasmash: I haven't looked into it, but we might be able to solve this by creating a custom bootstrap file: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#using-a-bootstrap-file

> These custom files may be used to perform such taks as manipulating the internal settings of PHP_CodeSniffer that are not exposed through command line arguments.

grasmash:
Maybe something like:
```
$cli = $phpcs->cli;
$specified_files = $cli->values['files'];
$cli->values['files'] = [];

$phpcs = new PHP_CodeSniffer($values['verbosity'], null, null, null);
$phpcs->setCli($cli);
// filter.
$phpcs->initStandard($standard, $values['sniffs'], $values['exclude']);
$values = $this->values;

$intersection = [];
foreach ($values['files'] as $file) {
  foreach ($specified_files as $specified_file) {
    if (strpos($specified_file, $file) !== FALSE) {
      $intersection[] = $specified_file;
    }
  }
}

$values['files'] = $intersection;
```

grasmash:
Closed in favor of #2134

PR #2130, opened 2017-10-12
Most people should use vagrant up, not blt vm

danepowell:
`blt vm` should only be run once per project, when the config files for the VM are first generated. Folks should subsequently just interact with Vagrant directly.

### DIFF
--- a/readme/onboarding.md
+++ b/readme/onboarding.md
@@ -28,7 +28,7 @@ You have probably been linked to this documentation by a project that is using B
 If your project uses a virtual development environment such as DrupalVM:
 
 1. Make sure you have installed any prerequisites. For DrupalVM, see the [quick start guide](https://github.com/geerlingguy/drupal-vm#quick-start-guide).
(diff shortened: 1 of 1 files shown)


malikkotob:
I agree with the actual contents of the change here, but in terms of `blt vm` should only be run once per project, isn't it run on occasion in special scenarios? e.g. upgrading to new version of BLT?

danepowell:
Well I guess you would run it if you previously nuked the vm `blt vm:nuke`, which you might do as part of a BLT upgrade.

I guess what I meant to say is that only the TA or lead developer for a project will ever have to run `blt vm`. It's better for other developers to be in the habit of interacting with vagrant directly.

danepowell:
Oh I just realized that `blt vm` does one critical thing: it sets your local default Drush alias to the DrupalVM alias. That's actually pretty important. So maybe it is better for developers to keep using it for the initial install.

PR #2150, opened 2017-10-17
Issue #2093: Replace Drush assume with no-interaction.

### LINKED ISSUE #2093: Fix usage of drush assume (--yes / --no)
If my understanding is correct, there is no difference between `--yes` and `--no-interaction` in Drush 9. Both simply use whatever the "default response" is for any prompts. Source: https://github.com/drush-ops/drush/issues/3014#issuecomment-334316826

There's no way to force either yes or no.

So I think we should just remove the "assume" option completely, and pass `--no-interaction` for all commands. Unless you want to have interactive commands as an option.

danepowell:
Fixes #2093

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "doctrine/common": "^2.5",
         "doctrine/inflector": "~1.1.0",
         "drupal/coder": "^8.2.11",
-        "drush/drush": "^9.0.0-beta7",
+        "drush/drush": "^9.0.0-beta8",
         "grasmash/drupal-security-warning": "^1.0.0",
         "grasmash/yaml-cli": "^1.0.0",
         "grasmash/yaml-expander": "^1.2.0",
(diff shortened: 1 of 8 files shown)


danepowell:
Hmm. Apparently the default non-interactive response for `drush si` is `no`, which is problematic. We need a way to force site install non-interactively.

This is also confusing because I thought `--yes` and `--no-interaction` were equivalent in Drush 9. So I'm not sure why this PR would break tests. Seems like an undocumented "feature".

grasmash:
Yeah I noticed some quirks when I was working on this issue, they don't appear to be equivalent.

danepowell:
This is blocked on https://github.com/drush-ops/drush/issues/3082

danepowell:
New upstream issue: https://github.com/drush-ops/drush/pull/3084

It sounds like Drush is actually going to preserve `--yes` and `--no`, along with `--no-interaction`, I'm just trying to confirm. So we'll have to think more carefully about which of the three options to use.

danepowell:
We should be able to use --no-interaction for site installs in Drush 9 beta8

PR #2151, opened 2017-10-18
Issue 1543

### LINKED ISSUE #1543: Twig Lint is not maintained and fails valid twig code

wu-edward:
Fixes #1543.

Changes proposed:
- Remove asm89/twig-lint and use Symfony twig linter
- Run twig lint command procedurally instead of in parallel, otherwise error messages are suppressed.
- Account for Twig functions and filters added by Drupal, so that linter does not flag as invalid.
- Provide config for users to add custom Twig filters and functions that can be specified in project.yml
- Reorder precommit lint commands so that PHPUnit tests do not fail on Travis.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
     "license": "GPL-2.0",
     "require": {
         "php": ">=5.6",
-        "asm89/twig-lint": "^1.0.2",
         "composer-plugin-api": "^1.0.0",
         "composer/installers": "^1.2.0",
         "composer/semver": "^1.4",
@@ -35,6 +34,7 @@
         "phpunit/phpunit": "^4.8",
         "squizlabs/php_codesniffer": "^2.7",
         "symfony/console": "^3.2",
+        "symfony/twig-bridge": "^3.3",
         "symfony/yaml": "^3.2",
         "tivie/php-os-detector": "^1.0",
         "wikimedia/composer-merge-plugin": "^1.4.1"
--- a/composer.lock
+++ b/composer.lock
@@ -4,59 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "08938fe475c403ef8fcff82ae53b1b38",
+    "content-hash": "15b12d1dae0fce79c8e75ac143388dbd",
     "packages": [
-        {
-            "name": "asm89/twig-lint",
-            "version": "1.0.2",
-            "source": {
-                "type": "git",
-                "url": "https://github.com/asm89/twig-lint.git",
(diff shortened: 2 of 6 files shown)


src/Robo/Commands/Git/GitCommand.php
```diff
@@ -47,13 +47,16 @@ public function commitMsgHook($message) {
    * @return int
    */
   public function preCommitHook($changed_files) {
+    $changed_files_list = explode("\n", $changed_files);
     $this->invokeCommands([
+      // Passing a file list to be PHPCS will cause all specified files to
```
grasmash: We don't need to add this comment back :)

src/Robo/Blt.php
```diff
@@ -116,6 +120,46 @@ private function addPluginsCommandsAndHooks() {
   }
 
   /**
+   * Adds Symfony (non-Robo) command classes to the application.
+   *
+   * @param \Acquia\Blt\Robo\Application $application
+   */
+  protected function addSymfonyCommands(Application $application) {
```
grasmash: I do wish we could move this out of the Blt.php file. It feels like it doesn't belong here.

src/Robo/Commands/Git/GitCommand.php
wu-edward: I refactored and removed all the changes to this class.

src/Robo/Blt.php
wu-edward: I agree, so I refactored and put everything in the TwigCommand class. I think it's cleaner now.

PR #2154, opened 2017-10-18
Issue #1816: Provide better multisite support for ACSF locally.

### LINKED ISSUE #1816: Provide better multisite support for ACSF locally

danepowell:
Fixes #1816 

Re-roll of #1982 against 9.1.x


grasmash:
This just needs testing on ACSF.

config/build.yml
```diff
@@ -1,3 +1,8 @@
+acsf:
```
danepowell: Instead of creating a new top-level key for `acsf` sites, I think this should use the existing `multisites` array below. This will allow you to run commands like `blt sync:db:all` for ACSF sites as well (at some point in the future when multiple local ACSF sites are supported, anyway)

PR #2167, opened 2017-10-20
It looks like @executeInDrupalVm crept back in to the sync:db command…

gabe-connolly:
… in 8.9.9, pulling it back out to re-enable successful database syncing from remote sites.


grasmash:
Duplicate of https://github.com/acquia/blt/pull/2153 ?

gabe-connolly:
@grasmash , you are correct that this is very close to #2153 .  It looks like I pulled out one more instance of syncDbAll from @executeInDrupalVm which wasn't in #2153 , but otherwise they are on the same track.

grasmash:
I have resolved the conflict in the pull request.

gabe-connolly:
@grasmash thanks!

PR #2168, opened 2017-10-20
[WIP] Fixes #2157: Adding aliases command.

### LINKED ISSUE #2157: Port club `aliases` command to BLT
Port [club aliases command](https://github.com/acquia/club/blob/master/src/Command/ACAliasesCommand.php) to BLT https://github.com/acquia/club.

ACSF alias generation should not be ported, as it is part of https://github.com/acquia/acsf-tools. 

@msherron perhaps acsf-tools could be updated to use Symfony commands?

grasmash:
Fixes #2157 

This is a completely untested work in progress, and has only been opened to serve as a starting point for others to complete the work.


grasmash:
This is a completely untested work in progress, and has only been opened to serve as a starting point for others to complete the work.

grasmash:
This should be modified in a few ways:

- Site factory aliases should not be generated
- Aliases should be in YAML format as per Drush 9 compatibility requirements, rather than using the current twig templates.
- We should probably replace the curl commands with a library like guzzle. See https://github.com/acquia/blt/blob/9.1.x/src/Robo/Commands/Tests/SecurityUpdatesCommand.php#L24.

danepowell:
This produces quite a few warnings about the Acquia SDK's use of abandoned guzzle libraries when it's installed. I'm not sure we want to have that in BLT, it's going to create a negative perception and unnecessary support requests. Not to mention the possible stability / security implications of relying on libraries that old.

danepowell:
I confirmed that the Acquia PHP SDK is officially abandoned (and not even really an Acquia tool). If we want to use it, we should take over ownership and upgrade it to use N3.

typhonius:
I've created a new project that uses Cloud API v2 (N3) here: https://github.com/typhonius/acquia-php-sdk-v2. If it makes sense to use that (or extend it) for this PR instead then hopefully it'll iron out bugs and catch edge cases.

danepowell:
Awesome thanks, I'll check it out

danepowell:
I'm trying to get authentication to work. I also need to figure out why tests are pulling stale versions of Drupal.

PR #2187, opened 2017-10-24
Pr blt issue 2173 drush multisite

### LINKED ISSUE #2173: sql-sync not availabe in drush
so, I think there is a bug in blt because of an “incompatibility” in Drush. Drush does not support --uri in sql-sync, but blt tries to execute this:

``` /vendor/bin/drush sql-sync @bayerphweb.01live @bayerphweb.local --structure-tables-key=lightweight --create-db --sanitize --uri='http://phdebetaferon.bayerphweb.acsitefactory.com' --yes ```

Instead, using acsf tools properly configured, blt should try to execute this drush command:

```drush sql-sync @phdebetaferon.01live @bayerphweb.local```

When I run this command:
```
blt sync -D site=SITENAME
```

I get the following output:
```
Allways pulls the default database
```

```
It should pull the database for SITENAME
```

### PR DESCRIPTION
**Pull request by alex-moreno**
Fixes #2173 .

Changes proposed:
- initialize() sends a new parameter from project.yml, project.machine_name, 
- setSiteConfig() is now multisite aware, adapting not just site but also drush remote alias.

### DIFF
--- a/src/Robo/Config/DefaultConfig.php
+++ b/src/Robo/Config/DefaultConfig.php
@@ -66,9 +66,19 @@ class DefaultConfig extends BltConfig {
    * @param string $site_name
    *   The name of a multisite. E.g., if docroot/sites/example.com is the site,
    *   $site_name would be example.com.
+   * @param string $machine_name
+   *   The machine name in a multisite environment.
+   *   E.g., @machine_name.sitename.dev
    */
-  public function setSiteConfig($site_name) {
+  public function setSiteConfig($site_name, $machine_name = NULL) {
     $this->config->set('site', $site_name);
+
+    // Adapt remote alias for the multisite.
+    if ($environment = $this->config->get('drush.aliases.remote_env')) {
+      echo ' environment:: ' . $environment;
+      $this->config->set('drush.aliases.remote', $machine_name . '.' . $site_name . '.' . $environment);
+    }
+
     if (!$this->config->get('drush.uri')) {
       $this->config->set('drush.uri', $site_name);
     }
--- a/src/Robo/Config/DefaultConfig.php
+++ b/src/Robo/Config/DefaultConfig.php
@@ -75,7 +75,6 @@ class DefaultConfig extends BltConfig {
 
     // Adapt remote alias for the multisite.
     if ($environment = $this->config->get('drush.aliases.remote_env')) {
-      echo ' environment:: ' . $environment;
       $this->config->set('drush.aliases.remote', $machine_name . '.' . $site_name . '.' . $environment);
     }
 
--- a/src/Robo/Config/ConfigAwareTrait.php
+++ b/src/Robo/Config/ConfigAwareTrait.php
@@ -23,6 +23,7 @@ trait ConfigAwareTrait {
    *   The config value, or else the default value if they key does not exist.
    */
   protected function getConfigValue($key, $default = NULL) {
+
     if (!$this->getConfig()) {
       return $default;
     }
--- a/src/Robo/Hooks/InitHook.php
+++ b/src/Robo/Hooks/InitHook.php
@@ -31,8 +31,9 @@ class InitHook extends Tasks implements IOAwareInterface, ConfigAwareInterface, 
     // input has been processed.
     $multisites = $this->getConfigValue('multisites');
     $first_multisite = reset($multisites);
+    $machine_name = $this->getConfigValue('project.machine_name');
     $site = $this->getConfigValue('site', $first_multisite);
-    $this->getConfig()->setSiteConfig($site);
+    $this->getConfig()->setSiteConfig($site, $machine_name);
   }
 
 }

### CODE REVIEW

File: src/Robo/Config/DefaultConfig.php [hunk 1]
```diff
@@ -66,9 +66,19 @@ public function populateHelperConfig() {
...
+   *   The machine name in a multisite environment.
+   *   E.g., @machine_name.sitename.dev
    */
-  public function setSiteConfig($site_name) {
+  public function setSiteConfig($site_name, $machine_name = NULL) {
     $this->config->set('site', $site_name);
+
+    // Adapt remote alias for the multisite.
+    if ($environment = $this->config->get('drush.aliases.remote_env')) {
+      echo ' environment:: ' . $environment;
```
alex-moreno: debug code to be removed

src/Robo/Config/DefaultConfig.php
```diff
@@ -66,9 +66,18 @@ public function populateHelperConfig() {
...
+   *   The machine name in a multisite environment.
+   *   E.g., @machine_name.sitename.dev
    */
-  public function setSiteConfig($site_name) {
+  public function setSiteConfig($site_name, $machine_name = NULL) {
     $this->config->set('site', $site_name);
+
+    // Adapt remote alias for the multisite.
+    if ($environment = $this->config->get('drush.aliases.remote_env')) {
+      $this->config->set('drush.aliases.remote', $machine_name . '.' . $site_name . '.' . $environment);
```
alex-moreno: this needs a new yml value:

```
drush:
  aliases:
    remote_env: 'prod'
```

alex-moreno:
this executes:

```vendor/bin/drush sql-sync @machine_name.sitename.prod @bayerphweb.local```

same for rsync:

```
drush rsync '@machine_name.sitename.prod:%files/' /Users/alejandromoreno/projects/acquia/bayer/it_bayer_engagementchannel_platform/docroot/sites/phdeengagev1/files
```

requires machine_name.drushrc.aliases.php to be configured correctly (acsf_tools should help). This will return:

```
alejandromoreno@Alejandros-MBP:~/projects/acquia/bayer/it_bayer_engagementchannel_platform$ drush sa
@machine_name
@machine_name.local
@machine_name.SITE1.dev
@machine_name.SITE1.prod
@machine_name.SITE1.test
@machine_name.SITE2.dev
@machine_name.SITE2.prod
@machine_name.SITE2.test
@none
```

grasmash:
I am a little hesitant to merge this because a significant amount of multisite work has been done in the 9.1.x branch. 8.9.x is currently being reserved for bug fixes only.

alex-moreno:
I am happy to port this to 9.x if that helps? Just need a day or so to test
it

grasmash:
It would be great if you could port this to 9.1.x. The 9.x  Branch is currently reserved for drush nine compatibility work only. The 9.0.x releases will essentially be 8.9.x functionality with drush nine compatibility added.

However, you may find that this issue is already resolved in the 9.1.x branch, since a lot of multisite compatibility work is already been done there.  But, the only way to know for sure is to test it.

danepowell:
Tell me if I understand this correctly.

Let's say your project machine name / ACSF account name is `foo`. With this patch, let's say you set `drush.aliases.remote_env` to `prod`.

Then if you execute `blt sync:refresh -D site=bar`, the database will be synced from `@foo.bar.prod` to `@foo.local` (or whatever your local alias is).

That seems useful enough, but I need to check how it interacts with other upcoming ACSF / multisite changes (#2154)

This also needs to go into 9.1.x. 8.9.x is maintenance only.

danepowell:
@alex-moreno please open this against 9.1.x and I'll merge it. If you can also add documentation to `multisite.md`, that would be great, otherwise I can do it.

alex-moreno:
yes @danepowell , that's the workflow.

Only issue is if you go with your vm full multisite/multiple local domains. I am finding issues syncing when using a full multisite environment locally. drush does not identify correctly which one of the databases you want to sync, so it only works when your sites use the default database, 

Will have a look at that too, but for now it fixes the issue that blt didn't detect remote databases in ACSF.

alex-moreno:
will do both, branch and doc. I would like to keep this PR merged in 8.9, as it's a pain point for those still on this version

danepowell:
Sorry, 8.9.x is bug fixes only at this point, we are trying to get 9.1.x stable as quickly as possible in order to provide better multisite / ACSF support. If we keep trying to improve 8.9.x, we'll never get there :)

alex-moreno:
given it's an improvement backported, and that the effort has already been done, IMO it should be merged. I don't have all the context so I am probably wrong. Just my two cents :-)

danepowell:
You're not wrong, we just have to consider how adding enhancements to an already stable 8.9.x branch could increase the maintenance burden, which would detract from our ability to get 9.x out as quickly. This seems fairly low risk but it's not always easy to anticipate how a new feature can cause problems down the line.

alex-moreno:
gotcha, and you are absolutely right :-). I guess we can live with a couple
of patches while we migrate to 9.x, and act as beta testers at the same
time :-)

Thank you.

PR #2210, opened 2017-10-27
Fixes #2087: Port blt-doctor to Drush 9.

### LINKED ISSUE #2087: Port `doctor` drush command to drush 9
We may need to split the doctor command into a new acquia/blt-doctor  package of type drupal-drush in order to gain access to the container. See reference below.

> grasmash [8:42 PM] 
can anyone tell me how I can get drush to discover a new annotated drush command if that command does not belong to a module, but to a package in vendor?
I’m not sure what the discovery process is

> chi [12:25 AM] 
@U6ZKNKUUF You need to specify `"type:drupal-drush"` in composer.json. (edited)
And site that consuming this package should have correct "installer-paths" section in its composer.json file.
https://github.com/drush-ops/drush/issues/2885

danepowell:


danepowell:
I've gotten Drush to detect the new command file but get this error:
> Uncaught Error: Call to undefined function Drupal\Core\drupal_valid_test_ua() in docroot/core/lib/Drupal/Core/DrupalKernel.php:374

drupal_valid_test_ua() exists in bootstrap.inc. It's like Drupal isn't bootstrapping or something. Not sure where to go from here.

danepowell:
And now it seems like we hit the problem of not being able to use dependency injection...
> Error: Call to a member function listAllFilePaths() on null in Drush\Commands\core\StatusCommands->getPropertyList()

Ordinarily, I think the site alias manager would be set via DI:
https://github.com/drush-ops/drush/blob/c9e6ac15b88d9329f560e90cc6c801b9b7859ad7/src/Runtime/DependencyInjection.php#L108

danepowell:
Sweet, this works now on my local. I just need to figure out why it doesn't work on Travis.

PR #2223, opened 2017-10-31
Site inspector should use Drush 9 commands.

danepowell:
Also port to 9.x.

Think if this can be combined with the alias checks in the doctor command, which seem to duplicate this.


danepowell:
Actually this doesn't exist in 9.x

PR #2227, opened 2017-10-31
Acsf hooks deploy

### LINKED ISSUE #2003: Replace post-code-deploy ACSF hook with db update Factory Hook
Currently ACSF BLT code deploy updates are run via a post-code-deploy hook which runs after the ACSF code update has completed. This was a workaround until a db-update factory hook was created in ACSF which will allow for customizing the current ACSF code update process on the update environments rather than the live environments. Replacing the post-code-deploy blt invocation with the db-update Factory Hook will allow for running BLT commands on the update environment and switching the DNS after completion rather than running them directly on the live environments which may present challenges with scalability. Internal ticket: DG-17926

lcatlett:
Fixes #2003 

Changes proposed:
- running BLT updates via post-code-deploy hooks are no longer recommended for ACSF in favor of db-update which will ensure zero-downtime deployments. 
- post-code-update hooks should be disabled by default for the 01live environment until a permanent solutions is evaluated for production deployments running a branch on 01live


scripts/cloud-hooks/functions.sh
```diff
@@ -8,12 +8,13 @@ drush_alias=${site}'.'${target_env}
...
   case $target_env in
-    01dev|01test|01live)
+    01dev|01test)
       acsf_deploy
       ;;
-    01devup|01testup|01update)
+    # Do not run deploy updates on 01live in case a branch is deployed in prod. 
+    01devup|01testup|01update|01live)
       ;;
-    ode[[:digit:]]*)
```
danepowell: Did you intend to remove the ode check here? Isn't this going to cause all deployments to ACE to install the site from scratch?

scripts/cloud-hooks/hooks/common/post-code-deploy/post-code-deploy.sh
```diff
@@ -20,9 +20,13 @@ repo_url="$5"
 repo_type="$6"
 
 
+acsf_file="/mnt/files/$AH_SITE_GROUP.$AH_SITE_ENVIRONMENT/files-private/sites.json"
```
danepowell: It would be better to use the provided script variables here (i.e. `$site.$target_env`), rather than environment variables, so that this script can be tested more easily.

settings/acsf/db-update/db-update.sh
```diff
@@ -0,0 +1,25 @@
...
+db_role="$3"
+# The public domain name of the website.
+domain="$4"
+
+# BLT executable:
+blt="/var/www/html/$site.$env/vendor/acquia/blt/bin/blt"
+
+deployupdate="$blt deploy:update --define environment=$env --define drush.uri=$domain -v -y"
+
+$deployudpate
```
danepowell: typo?

danepowell:
For anyone like me wondering why updates were disabled in production ACSF environments: you should be able to replicate this behavior by making sure to only deploy tags to production (don't use branches in production), and when you deploy make sure to tell ACSF to run DB updates. Then the BLT db-update factory hook will fire and run your usual updates like configuration imports.

Issue #2159, opened 2017-10-19
Add Factory Hooks required for ACSF site installs and updates


lcatlett:
New Factory hooks are available which will allow hooking into BLT tasks for install, update, and deploy. Additionally, there are number of hooks needed on most ACSF projects that should be provided as suggestions. (suggestions may be commented as needed) The following should be provided:

- [x] db-pupdate #2003
- [x] post-install #2440
- [ ] post-settings-php
- [ ] post-sites-php
- [ ] post-theme-deploy
- [ ] pre-settings-pjhp
- [ ] pre-sites-php

Legacy cloud hooks should be removed where possible in favor of the above factory hooks

danepowell:
One specific task I'll work on is replacing our current post-code-deploy/update hooks with factory-hooks/db-update: https://kbv2.acquia.com/site-factory/tiers/paas/workflow/hooks/dbupdate

damontgomery:
@lcatlett, based on your work in #2440, should we mark off the post-theme-deploy hook?

Do the other hooks just need templates so that there is broad understanding that they exist? If so, I can add those. If we need other recommended hooks, do you have examples I can look through?

Thanks.

lcatlett:
this can be closed with #2440 , #2227

PR #2229, opened 2017-10-31
Acsf factory hooks suggestions

### LINKED ISSUE #2209: ACSF support for setup and sync
In 9.1.x, you can now create multiple databases locally for ACSF sites, BLT will automatically switch between them based on hostname and the multisites config array.

But we still need a way to run `blt setup` and `blt sync` against specific sites/databases. Seems like the easiest way is to automatically transform the site parameter into a local drush alias (e.g. `@project_name.site_name.local`). But this requires that local drush aliases exist for each site. We could probably automatically construct these in the same way that acsf-tools does via sfa for remote aliases.

Once this is set up, we should test and document the entire ACSF workflow.

lcatlett:
Fixes #2209 #2159 

Changes proposed:
- Adds post-theme-deploy hook to clear twig cache on all web heads on theme deploy
-Adds support for running blt commands on site install. This will allow for using config split and on site install for new sites.


wimleers:
There should be a `@todo` to revert this at some point in the future, when all clients have a Drupal 8 version that is at least the one in which https://www.drupal.org/node/2752961 ships.

settings/acsf/post-install/post-install.php
```diff
@@ -0,0 +1,20 @@
...
+
+$site = $_ENV['AH_SITE_GROUP'];
+$env = $_ENV['AH_SITE_ENVIRONMENT'];
+
+// The public domain name of the website.
+// Run updates against requested domain rather than acsf primary domain.
+$domain = $_SERVER['HTTP_HOST'];
+
+
+exec('/mnt/www/html/$target_env/vendor/acquia/blt/bin/blt deploy:update --define environment=$env --define drush.uri=$domain -v -y');
```
grasmash: I don't think that variable substitution will work here since you're using single quotes.

grasmash: Please use the long form of all options. E.g., `--verbose --yes`.

settings/acsf/post-install/post-install.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/**
+ * @file
+ * Factory Hook: post-install.
+ *
+ * This hook enables you to execute PHP after a new website is created
+ * in your subscription. Unlike most API-based hooks, this hook does not
+ * take arguments, but instead executes the PHP code it is provided.
```
grasmash: Add `@todo Remove this script after https://www.drupal.org/node/2752961 is resolved and the corresponding Drupal core version is set as a minimum requirement for BLT.`

settings/acsf/post-theme-deploy/clear-twig-cache.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Clears TWIG cache for a site after a theme deploy. Executes via drush alias.
```
grasmash: Add @todo Remove this script after https://www.drupal.org/node/2752961 is resolved and the corresponding Drupal core version is set as a minimum requirement for BLT.

damontgomery:
I opened #2440 to clear up some of the confusion around this PR.

PR #2230, opened 2017-10-31
Clear twig caches on all web servers on deploy / update.

### LINKED ISSUE #2228: Clear twig caches on deploy
Currently, twig caches are cleared only on the first web server on deploy which is a problem for deploys on environments with multiple web server (such as most Acquia cloud production environments). The cloud hooks templates currently provided should be modified to clear the caches on all webs.

lcatlett:
Fixes #2228 

Changes proposed:
-  Adds support for clearing twig caches on deploy on all web servers rather than just the primary.

### DIFF
--- a/scripts/cloud-hooks/functions.sh
+++ b/scripts/cloud-hooks/functions.sh
@@ -45,6 +45,7 @@ acsf_deploy() {
   for uri in "${sites[@]}"; do
   #Override BLT default deploy uri.
   blt deploy:update --define environment=$target_env --define drush.uri="$uri" -v -y
+  drush @"${drush_alias}" --uri=$uri ev '\Drupal\Core\PhpStorage\PhpStorageFactory::get("twig")->deleteAll();'
   if [ $? -ne 0 ]; then
       echo "Update errored for site $uri."
       exit 1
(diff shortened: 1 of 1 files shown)


danepowell:
@lcatlett wouldn't this only clear the Twig cache on the primary web head?

Previously we had to use a script to SSH into each web server in sequence and run `\Drupal\Core\PhpStorage\PhpStorageFactory::get("twig")->deleteAll();`, has something changed that would clear it across all servers?

wimleers:
There should be a `@todo` to revert this at some point in the future, when all clients have a Drupal 8 version that is at least the one in which https://www.drupal.org/node/2752961 ships.

danepowell:
@lcatlett or @wimleers , can either of you respond to my comment above? I'm pretty sure calling `PhpStorageFactory::get("twig")->deleteAll();` will only delete the twig cache on that particular server. What mechanism is there to clear is across all servers?

lcatlett:
I mistakenly thought we needed to make the same change for ACE as in the ACSF factory hook even when using the deployment_identifier. It seems the deployment identifier absolves the need for this additional flush on deploy so this can be closed

danepowell:
Thanks for following up.

Issue #2242, opened 2017-11-02
TASK/POC: Optionally isolate configs per site


alex-moreno:
Here is something it can be useful for someone else, but happy to received feedback if this is an awful, terrible, bad idea. 

In some (multisite) situations, due to complexity or some other factors, you want to keep your sites configurations isolated. With this PR, you can have:

```
config/site1/
config/site2/
...
```

and then your splits:

```
config/dev
config/test
...
```

When doing a 

```
blt setup:config-import -D site=SITE1
```

What happens is this:

```
...[]/drush @machinename.local pm-enable config_split --uri=SITE1 --yes ...[]
...[]drush @machinename.local config-import sync --uri=SITE1 --yes ...[]
```

so, effectively, it will be importing the configs using uri as the folder, and applying the splits.

danepowell:
I think this functionality already exists in 9.1.x, albeit via a slightly different mechanism:
- https://github.com/acquia/blt/blob/9.1.x/settings/config.settings.php#L88
- https://github.com/acquia/blt/blob/9.1.x/readme/config-split.md

Can you confirm that that would satisfy your use case? (if you were on 9.1.x of course)

danepowell:
Closing due to inactivity

danepowell:
Sorry, didn't see the comment in the associated PR

danepowell:
@alex-moreno see my comment here, I actually don't think any code change is necessary to support this functionality. Although maybe we could make it a little more user friendly.

https://github.com/acquia/blt/pull/2243#discussion_r150624916

danepowell:
Closing due to inactivity, and the fact that I think this is already supported without any code change necessary. Reopen if necessary.

PR #2243, opened 2017-11-02
POC: Optionally isolate configs per site

alex-moreno:
Fixes # 2242.

Changes proposed:
- Isolate configs per site


src/Robo/Commands/Setup/ConfigCommand.php
```diff
@@ -39,8 +39,24 @@ public function import() {
       // If using core-only or config-split strategies, first check to see if
       // required config is exported.
       if (in_array($strategy, ['core-only', 'config-split'])) {
-        $core_config_file = $this->getConfigValue('docroot') . '/' . $this->getConfigValue("cm.core.dirs.$cm_core_key.path") . '/core.extension.yml';
 
+        // We don't want configs to be shared between multisites.
+        if ($this->getConfigValue('cm.share-configs') == FALSE) {
+          $drush_uri = $this->getConfigValue('drush.uri');
+
```
wouteradem: @alex-moreno Are we sure 'drush.uri' is always set? If it's not shouldn't we fall back to 'default'? Also, is drush.uri the best variable to use here?

alex-moreno: niiiiice, good point. If we fall back to default/ you could have some sites using their isolated config, but some using the default. Love it.

Will commit the change in a second

danepowell:
See my comment in #2242, this functionality should already exist in 9.1.x, because BLT automatically sets the current site split to be active for you. Is this functionally different?

alex-moreno:
Yes, it is different .this gives you the option the option to isolate
configs per site, so you don't have 100s of them with different configs
colliding between them (nightmare scenario).

Not sure if this would collide with the splits, I'd suggest to put the site
configs on a /sites/ subfolder, so the splits are somewhere else

danepowell:
Okay, I can see the benefit here. We currently assume that folks will use config splits (along with a default base configuration) to manage multisites, because ideally multisites will share a significant portion of their config. But if you have several completely independent multisites sharing little or no configuration, it would be better to have them completely isolated. Although I would also argue that they should be completely separate codebases at that point (rather than a multisite) :smile:

src/Robo/Commands/Setup/ConfigCommand.php
```diff
@@ -39,8 +39,23 @@ public function import() {
...
-        $core_config_file = $this->getConfigValue('docroot') . '/' . $this->getConfigValue("cm.core.dirs.$cm_core_key.path") . '/core.extension.yml';
+        // Read the config from the default folder (/config/default). This will
+        // Also be the fallback option, so some domains can still use default.
+        $core_config_file = $this->getConfigValue('docroot') . '/' .
+          $this->getConfigValue("cm.core.dirs.$cm_core_key.path") .
+          '/core.extension.yml';
+
+        // We don't want configs to be shared between multisites.
+        if ($this->getConfigValue('cm.share-configs') == FALSE) {
+          $drush_uri = $this->getConfigValue('drush.uri');
```
danepowell: Let's use the `site` here rather than the Drush URI, so that it matches the directory structure in `docroot/sites`:
`$this->getConfigValue('site')`

Every BLT command operates within a specific site context. This is 'default' by default, but can be modified at run-time by passing the `site` parameter.

danepowell: @alex-moreno actually is such a large code change here even necessary? I think the functionality you desire is already supported. You should be able to just set the following parameter in each site's `site.yml`:
`cm.core.dirs.$cm_core_key.path` (or specifically maybe `cm.core.dirs.sync.path`)

src/Robo/Commands/Setup/ConfigCommand.php
```diff
@@ -39,8 +39,23 @@ public function import() {
...
+
+          // Read the config from the site folder (/config/URI/).
+          $core_config_file = $this->getConfigValue('docroot') . '/' .
+            $this->getConfigValue("cm.core.dirs.$cm_core_key.path") . '/../' .
```
danepowell: Isn't this functionally equivalent to:

```php
$core_config_file = $this->getConfigValue('docroot') . '/' .
  $drush_uri . '/core.extension.yml';
```

danepowell:
Closing due to inactivity, and the fact that I think this is already supported without any code change necessary. Reopen if necessary.

PR #2244, opened 2017-11-02
Documentation: Acsf remote aliases

alex-moreno:
Fixes documentation for #2173 (https://github.com/acquia/blt/pull/2198).


grasmash:
I prefer not to use prod as a default remote. It's better to use test and avoid the impact of running sql dumps on prod whenever a dev wants to locally refresh.

PR #2264, opened 2017-11-09
Fixes #2253: Move vm.enable to project.local.yml.

### LINKED ISSUE #2253: Update function to move vm.enable to project.local.yml
The `vm.enable` parameter should only live in project.local.yml, but this update function was putting it in project.yml: https://github.com/acquia/blt/blob/8.9.x/src/Update/Updates.php#L295

We should write a subsequent update hook to move that to project.local.yml

danepowell:
Fixes #2253


danepowell:
port to 9.x and 9.1.x

PR #2274, opened 2017-11-13
Fixes #2272: Improved getting started with DrupalVM docs.

### LINKED ISSUE #2272: Doc issue: Creating a new project documentation needs instructions for customizing DrupalVM
I thought about just writing my proposed addition and submitting a patch, but I wanted to put an issue out there first to see if others agree. It seems to me that we need a step after step 3, but before step 4 to instruct users that if they are using DrupalVM and do not want the default configuration, they need to modify box/config.yml following the instructions at http://docs.drupalvm.com/en/latest/getting-started/configure-drupalvm/.

Motivation: Some users may not understand how to properly add extras such as Solr when initializing their project.

danepowell:
Fixes #2272 

@rhuffstedtler does this look good to you?

### DIFF
--- a/readme/creating-new-project.md
+++ b/readme/creating-new-project.md
@@ -20,9 +20,11 @@
 
 1. Now its time to spin up your LAMP stack.
 
-    1. **With Drupal VM**: If you would like to use Drupal VM for local development, run the following commands:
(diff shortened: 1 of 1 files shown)


rhuffstedtler:
@danepowell - That works. I was going the route of running blt:vm config, then modifying box/config.yml, then running blt:vm, but saying no and changing the file gets the user to the same place. Thanks!

PR #2281, opened 2017-11-15
Clean up secrets file.

danepowell:
This generally cleans up the secrets include process and sample file, and most notably changes the file location to be non-environment specific.


danepowell:
@malikkotob let me know what you think. Sorry for the general refactoring which might make it a little harder to see the changes, but I think you'll get it.

settings/sample-secrets.settings.php
```diff
@@ -7,34 +7,25 @@
...
+/**
+ * The below example would set a sensitive API key for the "foo" module.
+ *
+ * In addition to $ah_env, you can use other variables defined in blt.settings.php
+ * such as $is_dev_env, $is_stage_env, $is_prod_env, and $ah_site.
+ */
+switch ($ah_env) {
+  case 'dev':
+  case 'test':
+    $settings['foo']['super_secret_key'] = 'DEV-1234';
```
malikkotob: Is this any different than?
```
$config['foo.settings']['super_secret_key']
```

settings/blt.settings.php
```diff
@@ -145,24 +145,26 @@
...
   // Store API Keys and things outside of version control.
   // @see settings/sample-secrets.settings.php for sample code.
-  $secrets_file = sprintf("/mnt/gfs/%s.%s/secrets.settings.php", $_ENV['AH_SITE_GROUP'], $_ENV['AH_SITE_ENVIRONMENT']);
+  $secrets_file = "/mnt/gfs/home/$ah_group/secrets.settings.php";
   if (file_exists($secrets_file)) {
     require $secrets_file;
   }
   // Includes secrets file for given site.
-  $site_secrets_file = sprintf("/mnt/gfs/%s.%s/$site_dir/secrets.settings.php", $_ENV['AH_SITE_GROUP'], $_ENV['AH_SITE_ENVIRONMENT']);
+  $site_secrets_file = "/mnt/gfs/home/$ah_group/$site_dir/secrets.settings.php";
```
malikkotob: Does the `$site_dir` exist for multisites at this location? I don't see it on the sub that I'm looking at. I'm not looking at a multisite, but I would've expected to see a `default` directory at a minimum.

malikkotob:
@danepowell left a couple of comments in the PR

settings/sample-secrets.settings.php
danepowell: Settings and config are different subsystems. Whether you should use `$settings` or `$config` depends on how exactly the module you are using stores this data.

Well-designed modules will use settings to store sensitive data like API keys, rather than config. Config is designed to be portable, exportable, stored in VCS, etc... and API keys should be none of those things.

ref: https://anavarre.net/from-conf-to-config-and-settings-in-drupal-8/#whentousesettingsindrupal8

settings/blt.settings.php
danepowell: The `/mnt/gfs/home/$ah_group` directory exists for all subs, but anything deeper than that you have to create yourself. This includes the root `secrets.settings.php`, as well as the site-specific `secrets.settings.php` files (as well as their parent directories).

So to answer your question, no the directory doesn't exist, you'd need to create it.

PR #2283, opened 2017-11-15
Quote ROOT_DIR

candrews:
ROOT_DIR may contain spaces so it must be quoted.

Without the quotes, if ROOT_DIR contains a space, the user will receive a "No such file or directory" error.

Fixes # .

Changes proposed:
-
-


danepowell:
Makes sense, thanks. However 8.x is no longer a maintained branch, as stated in the README. I think you want 8.9.x. I've reopened this as #2284

PR #2301, opened 2017-11-27
Updates BLT to use stable version of Acquia API SDK.

typhonius:
Changes proposed:
- Configures BLT to use a stable version of the Acquia PHP SDK library (v1.0.1)
- Alters the commands required to make use of that library.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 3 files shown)


typhonius:
Please note this passes automated tests but will need manual testing prior to being included.

composer.lock
```diff
@@ -1004,7 +1005,7 @@
                 "collections",
                 "iterator"
             ],
-            "time": "2017-07-22T10:37:32+00:00"
```
danepowell: @typhonius please update your local version of Composer: `composer global self-update`. It looks like you are using an older version that produces this deprecated timestamp format.

danepowell:
Thanks a lot, this looks good!

PR #2315, opened 2017-11-29
Update memcache.md

### LINKED ISSUE #2314: Memcache documentation needs refining
1) The [memcache readme](https://github.com/acquia/blt/blob/9.1.x/readme/memcache.md) contains the following sample code to add to `settings.php`:
```
if ($is_ah_env) {
  switch ($ah_env) {
    case 'test':
    case 'prod':
      if ($modules && isset($modules['module']['memcache'])) {
        // Use Memcached extension.
        $memcached_exists = class_exists('Memcached', FALSE);
        if ($memcached_exists) {
          $settings['memcache']['extension'] = 'Memcached';
        }

        // Use memcache as the default bin.
        $settings['cache']['default'] = 'cache.backend.memcache';

        // Enable stampede protection.
        $settings['memcache']['stampede_protection'] = TRUE;
        // Move locks to memcache.
      }
      break;
  }
}
```

The line `if ($modules && isset($modules['module']['memcache'])) {` never evaluates to true, and memcache is never used as the default cache backend. A different method for checking the existence of the module or necessary service needs to be used here until https://www.drupal.org/node/2766509 is merged.

2) The [memcache readme](https://github.com/acquia/blt/blob/9.1.x/readme/memcache.md) also contains information on SASL authentication support along with a code snippet on how to enable it:
> If using a supported Acquia product or service, SASL authentication support in the Drupal Memcache module can be enabled by updating and placing the below in correct site-specific settings.php file, ideally one that is not committed to version control.

I spoke with a member of Acquia's support team, and came away from the discussion with the conclusion that this documentation is not necessary. SASL is only relevant/enabled for sites using a per-site memcached setup (a good example would be ODEs), and for those sites, the SASL credentials are setup automatically on ACE. And again, if the site is not using it a per-site memcached setup, SASL is not enabled.

We can check if a site has per-site memcached by sshing into the server and running:
```
ps auxwww | grep memcached
```

It yields the following for a subscription that I'm currently working with:
```
site_group@server:/var/www/html/site_group.env/docroot$ ps auxwww | grep memcached
site_group     23165  0.0  0.0  22628   960 pts/0    S+   19:45   0:00 grep memcached
nobody   25767  0.0  0.0 325288  1920 ?        Ssl  Nov10   3:05 /usr/bin/memcached -p 11211 -m 64 -c 1024 -I 1m
```

We can see a lone memcached process that is running on port 11211 without a `-S` flag, which reflects that SASL is not enabled.

For a site with per-site memcached setup, we'd expect something like the following:
```
+  8238  0.0  0.0 352820  8284 ?        Ssl  Oct30   7:08 /usr/bin/memcached -p 11212 -m 64 -S
welcome   8258  0.0  0.0 347960  2388 ?        Ssl  Oct30   7:09 /usr/bin/memcached -p 11213 -m 64 -S
+  8271  0.0  0.2 389916 46124 ?        Ssl  Oct30   7:26 /usr/bin/memcached -p 11214 -m 64 -S
nobody   23738  0.0  0.3 384168 61948 ?        Ssl  Oct30  14:08 /usr/bin/memcached -p 11211 -m 128 -c 1024 -I 1m
```

Notice the three memcached processes running on different ports all with the `-S` flag. These memcached processes all have SASL enabled.

greylabel:
Fixes #2314 .

Changes proposed:
- Remove SASL docs.
- Remove errant module check.

### DIFF
--- a/readme/memcache.md
+++ b/readme/memcache.md
@@ -15,37 +15,21 @@ if ($is_ah_env) {
   switch ($ah_env) {
     case 'test':
     case 'prod':
-      if ($modules && isset($modules['module']['memcache'])) {
-        // Use Memcached extension.
-        $memcached_exists = class_exists('Memcached', FALSE);
-        if ($memcached_exists) {
-          $settings['memcache']['extension'] = 'Memcached';
-        }
-
-        // Use memcache as the default bin.
-        $settings['cache']['default'] = 'cache.backend.memcache';
-
-        // Enable stampede protection.
-        $settings['memcache']['stampede_protection'] = TRUE;
-        // Move locks to memcache.
-        $settings['container_yamls'][] = DRUPAL_ROOT . '/../vendor/acquia/blt/settings/memcache.yml';
+      // Use Memcached extension.
+      $memcached_exists = class_exists('Memcached', FALSE);
+      if ($memcached_exists) {
+        $settings['memcache']['extension'] = 'Memcached';
       }
-      break;
+
+      // Use memcache as the default bin.
+      $settings['cache']['default'] = 'cache.backend.memcache';
+
+      // Enable stampede protection.
+      $settings['memcache']['stampede_protection'] = TRUE;
+      // Move locks to memcache.
+      $settings['container_yamls'][] = DRUPAL_ROOT . '/../vendor/acquia/blt/settings/memcache.yml';
+    break;
   }
-}
-```
-
-If using a supported Acquia product or service, SASL authentication support in the Drupal Memcache module can be enabled by updating and placing the below in correct site-specific `settings.php` file, ideally one that is not committed to version control.
-
-```
-if ($is_ah_env) {
-  $settings['memcache']['sasl'] = [
-  'username' => 'yourSASLUsername',
-  'password' => 'yourSASLPassword',
-  ];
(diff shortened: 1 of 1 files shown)


malikkotob:
This looks good, @greylabel, but until https://www.drupal.org/node/2766509 is merged, do we want to:
1) make mention of the fact that if the sample setup in `settings.php` is deployed at the same time as the memcache module/config that enables it things won't work? (I think you'd have to enable the module before deploying the code that changes the cache backend to memcache)

or

2) modify the code sample in `settings.php` to handle this scenario

cc @lcatlett

greylabel:
1., memcache.md already mentions this:

> BLT treats Memcache module integration as opt-in as it can be used only in environments that have Drupal installed and the Memcache module enabled, and won't need to be reinstalled, e.g. stage and prod.

2., @grasmash and I deferred on this for now via the conversations in #2027 and #2063 (Lots of comments in the _Show outdated section of the PR conversation). Would great to see this solved in [#2766509](https://www.drupal.org/node/2766509) as you mentioned.

grasmash:
@greylabel Do you agree that this pull request is superseded by #2324?

greylabel:
@grasmash Yes, assuming #2324 gets merged.

PR #2324, opened 2017-11-30
Fixes #2314: Update memcache config and docs.

### LINKED ISSUE #2314: Memcache documentation needs refining

lcatlett:
Fixes #2314 

Changes proposed:
- Updates documentation for memcache configuration locally and on Acquia 
- Allows for rebuiding service container, deploying memcache config without installing the module, and purging stale static caches. 
- This config will be required for upcoming changes to simplesaml_auth support php 7.1 / memcached / simplesaml 1.15.0


readme/memcache.md
```diff
@@ -2,76 +2,198 @@
...
+        'arguments' => ['container', '@memcache.backend.cache.container', '@lock.container', '@memcache.config', '@cache_tags_provider.container'],
+      ],
+    ],
+  ];
+
+      // Override default fastchained backend for static bins. 
+      // See https://www.drupal.org/node/2754947
+
+      $settings['cache']['bins']['bootstrap'] = 'cache.backend.memcache';
+      $settings['cache']['bins']['discovery'] = 'cache.backend.memcache';
```
greylabel: Why would we not want to continue using APC for these bins?

settings/cache.settings.php
```diff
@@ -8,3 +8,21 @@
...
+
+/**
+ * Override static cache defaults in core.services.yml
+ * 
+ * Override the default backend set on the service definition on a per-bin basis. 
+ * 
+ * The bootstrap, discovery and config bins use the chainedfast backend by default and
+ * since the core service definition for these bins sets the expire to 
+ * CacheBackendInterface::CACHE_PERMANENT, these objects are cached permanetly by default
+ * and may result in stale configuration whjen rebuilding the service container on deploymemnts
```
greylabel: typo: `whjen`

readme/memcache.md
lcatlett: APC is enabled by default for these caches, but they are set to never expire so even when running a cr they are never purged which causes issues when you need to rebuild the service container - for example if you installed a views plugin and did a cache rebuild after deploying to 'refresh' a view using the new plugin, you'd often get a WSOD or other errors because there will be a null value due to the cache_discovery table having stale data. This snippet example was also intended to demonstrate the needed config to install Drupal with alternative cache bins by default and since this would require rebuilding the service container as modules / plugins were installed, this is not possible with the bootstrap / config / discovery using APCu because the stale cache objects often result in install errors.

greylabel: I need to revisit this, but is the cache lifetime related to the bin/cache entry, rather than the backend -- would these entries still be marked as `CACHE_PERMANENT` with memcache?

lcatlett: in core the cache lifetime is related to the backend and then various bins use those backend / `cache_permanent` settings by default. in this cache the objects now using memcache as the backend would no longer use `CACHE_PERMANENT` and can be expired.

greylabel: +1 Awesome

greylabel: Also, very awesome that this includes the code change to cache Container on bootstrap.

PR #2347, opened 2017-12-06
Fixes #2343: Use values from ci.yml during tests.

### LINKED ISSUE #2343: environment config file and travis run_tests script
I have two related issues I wanted to bring up, and potentially get some feedback on before I submit potential PRs.

I'm currently on BLT v**8.9.10** and I'm trying to set up some unique Travis behavior for testing on a custom domain & subdomain instead of http://127.0.0.1.
One minor question I have is: I'm wondering if there's any reason BLT uses localhost IP out of the box instead of setting up some kind of [custom hostname in Travis](https://docs.travis-ci.com/user/hostname)? It's working so far for me, but just wondering if I'm overlooking some gremlin that will bite me in the future...?

Anyways, since I'm using this custom domain, I need to set `tests.server.url` and `project.local.hostname` to get this to run correctly. I first tried editing my `blt/ci.yml` file, but I realized that it seems on 8.9.10, this file [is not being included](https://github.com/acquia/blt/blob/8.9.x/bin/blt-robo-run.php#L34).
When I run this command...
`$ blt config:get --define environment=ci project.local.hostname`
I am expecting to get the value in my `blt/ci.yml` file, but I don't get that. This, to me, feels like a 🐞 bug in 8.9.10 (I see this whole process was tweaked in 9.x), but I don't know offhand how to fix this. Would calling this a bug in 8.9 be fair?

Second, I realized that even if I _could_ set these properties in my `blt/ci.yml` file, it wouldn't completely help me with my problem. This is because the `vendor/acquia/blt/scripts/travis/run_tests` script does not pass `--define environment=ci` into the `$ blt tests:all` command. 
Is there any reason this script shouldn't be using that file when invoking this command? I'd like to propose a PR to fix that. What branches should I fix this against?

Thanks!

_j/w - what is the ETA around a 9.x release? Anywhere to monitor the status of that?_

danepowell:
Fixes #2343

### DIFF
--- a/bin/blt-robo-run.php
+++ b/bin/blt-robo-run.php
@@ -30,8 +30,8 @@ $processor->extend($loader->load($config->get('blt.root') . '/config/build.yml')
 $processor->extend($loader->load($config->get('repo.root') . '/blt/project.yml'));
 $processor->extend($loader->load($config->get('repo.root') . '/blt/project.local.yml'));
 
-if ($input->hasArgument('environment')) {
-  $processor->extend($loader->load($config->get('repo.root') . '/blt/' . $input->getArgument('environment') . '.yml'));
+if ($input->hasParameterOption('environment')) {
+  $processor->extend($loader->load($config->get('repo.root') . '/blt/' . $input->getParameterOption('environment') . '.yml'));
 }
 
 $config->import($processor->export());
--- a/scripts/travis/run_tests
+++ b/scripts/travis/run_tests
@@ -7,6 +7,6 @@ yaml-cli update:value blt/project.yml project.local.hostname '127.0.0.1:8888'
 
 blt validate:all
 ${BLT_DIR}/scripts/travis/tick-tock.sh blt setup --define drush.alias='${drush.aliases.ci}' --define environment=ci --no-interaction --yes -v
-blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --yes -v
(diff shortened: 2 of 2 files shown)


dpagini:
Thanks for this @danepowell... This is what I was trying locally...  it seems like the file is working now, but do you have any guesses as to why this doesn't work...?

ci.yml:
```yml
project:
  local:
    hostname: example.com
key: value
```

outputs this:
```sh
$ blt config:get -D environment=ci project.local.hostname
original.hostname.com
$ blt config:get -D environment=ci key
value
```

I guess I'm just having a problem with `project.local.hostname` key. I can get others to output correctly. Do you know of anywhere else in the 8.9.x code that might be getting overwritten or something?

danepowell:
@dpagini I can't reproduce the behavior you're seeing, the ci.yml overrides work fine for me on 8.9.x after this update. Here are the testing steps I followed:

1. Create a fresh BLT project for testing: `composer create-project acquia/blt-project --no-interaction blt2347`
1. Update to dev: `composer require "acquia/blt:8.9.x-dev"` (make sure it's commit ff27504 or later)
1. Add to the bottom of blt/ci.yml:
```
project:
  local:
    hostname: example.com
```
1. `blt config:get -D environment=ci project.local.hostname` -> `example.com`

If you're still having trouble open a new issue, this PR probably isn't the best place to troubleshoot.

dpagini:
Worked for me after I updated and tried on Travis. Maybe I had a dirty vendor from when I was trying the fix myself. Thanks again!

PR #2375, opened 2017-12-13
Fixes #2371: multisite commands do not pass uri successfully.

### LINKED ISSUE #2371: sync:refresh:all just syncs over the default database multiple times
Hello all:

I struggle with blt-driven multisite usage often.  I was really excited to see commands such as sync:refresh:all and sync:db:all that on the surface seemed to be aware that multisites tend to have different databases, config folders, and settings.  

In practice these commands tend to ignore site specific settings at exhibit behavior such as:
1. Overwrite the default "drupal" database for each site in the multisite
2. Use the default config while importing config for each site.
3. Used drush.aliases.remote to proper get data from the proper remote resource, but doesn't use drush.aliases.local to get settings about the multisite specific database or config folder.

I'm wondering, is it that the documentation is wrong?  https://github.com/acquia/blt/blob/8.x/readme/multisite.md#multisite-tasks

Or are my expectations wrong?

grasmash:


grasmash:
Relies on https://github.com/drush-ops/drush/pull/3261

PR #2377, opened 2017-12-14
Update INSTALL.md

danepowell:
* Brew repos were deprecated
* Drush install method has changed


greylabel:
You can also install `drush-launcher` with Brew now, if you want it globally instead of `drush`.

See`brew info drush-launcher`

PR #2380, opened 2017-12-18
Adding site name to local private files path to support multisite.

rjgwiz:
Changes proposed:
- ~~Move `$settings['file_private_path']` to `filesystem.settings.php`, alongside `file_public_path`~~
- In default local settings: add site_dir as a sub-folder under `/files-private` to support multisite. Prevents private file/sub-folder collisions between sites. No adjustments needed for deploy-exclude, gitignore, etc.

### DIFF
--- a/settings/filesystem.settings.php
+++ b/settings/filesystem.settings.php
@@ -5,7 +5,13 @@
  * Contains filesystem settings.
  */
 
+// Local file paths.
 $settings['file_public_path'] = "sites/$site_dir/files";
+$settings['file_private_path'] = "$repo_root/files-private/$site-dir";
+if (isset($acsf_site_name)) {
+  $settings['file_public_path'] = "sites/default/files/$acsf_site_name";
+  $settings['file_private_path'] = "$repo_root/files-private/$acsf_site_name";
+}
 
 // ACSF file paths.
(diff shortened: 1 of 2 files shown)


settings/filesystem.settings.php
```diff
@@ -5,7 +5,13 @@
  * Contains filesystem settings.
  */
 
+// Local file paths.
 $settings['file_public_path'] = "sites/$site_dir/files";
+$settings['file_private_path'] = "$repo_root/files-private/$site-dir";
```
dpagini: Just going to ask the question... should line 23 below get the same treatment? If so, how do we make this change and keep it non-disruptive to anyone with a site in prod?
Even this change... it says "local file paths", but for anyone not on Acquia hosting, this is changing their file path.

rjgwiz:
Good point @dpagini. Updated to leave `filesystem.settings.php` alone, and add `$site_dir` to the end of the default private files path in `default.local.settings.php`

dpagini:
Hmm... I think the first change makes more sense to me (in the context of multi-site), but I would be interested to hear @danepowell or @grasmash's thoughts on how to introduce a change that big into the project...?

PR #2382, opened 2017-12-19
BLT-2322: Checks if db config matches codebase config.

### LINKED ISSUE #2322: Feature request: DB vs codebase config check when running blt setup.
A nice sadness saver would be to have `blt setup` check to see if the database config matches the codebase config and prompt before continuing. Like a “Hey, FYI, your config in the DB doesn’t match the codebase. You cool with nuking that?” A force option could also be added to route around potential script issues where you really don't care what is in the database.

This could save A LOT of headaches and development time where developers accidentally wipe their sites before exporting configuration.

dooleymatt:
Fixes #2322 .

Changes proposed:
- Adds an interact hook to prevent users from accidentally overwriting their active config when running a setup command.

### DIFF
--- a/src/Robo/Hooks/InteractHook.php
+++ b/src/Robo/Hooks/InteractHook.php
@@ -100,7 +100,7 @@ class InteractHook extends BltTasks {
   }
 
   /**
-   * Confirms active config matches config in sync directory.
+   * Prompts user to confirm overwrite of active config on blt setup.
    *
    * @hook interact @interactConfigIdentical
    */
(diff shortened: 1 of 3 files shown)


dooleymatt:
@lahoosascoots Want to take a look at this? I think it's probably still a work in progress, but let me know if you have any thoughts on how to improve it.

src/Robo/Inspector/Inspector.php
```diff
@@ -761,4 +761,19 @@ protected function warnIfXdebugLoaded() {
...
+   * Determines if the active config is identical to sync directory.
+   *
+   * @return bool
+   *   TRUE if config is identical.
+   */
+  public function isActiveConfigIdentical() {
+    $result = $this->executor->drush("cex --no")
+      ->run();
+    $message = trim($result->getMessage());
+    if (strpos($message, 'The active configuration is identical to the configuration in the export directory')) {
```
grasmash: I think I'd rather rely on the exit code. Will `$result->getExitCode()` suffice?

src/Robo/Hooks/InteractHook.php
```diff
@@ -98,4 +99,31 @@ public function interactExecuteUpdates(
...
+   * Confirms active config matches config in sync directory.
+   *
+   * @hook interact @interactConfigIdentical
+   */
+  public function interactConfigIdentical(
+    InputInterface $input,
+    OutputInterface $output,
+    AnnotationData $annotationData
+  ) {
+    if ($this->getConfigValue('cm.strategy') == 'config-split') {
```
grasmash: Don't we want this for `core-only` too?

src/Robo/Hooks/InteractHook.php
```diff
@@ -98,4 +99,31 @@ public function interactExecuteUpdates(
...
+      $drupal_installed = $this->getInspector()->isDrupalInstalled();
```
grasmash: This is unused.

src/Robo/Hooks/InteractHook.php
```diff
@@ -98,4 +99,31 @@ public function interactExecuteUpdates(
...
+      $config_identical = $this->getInspector()->isActiveConfigIdentical();
+      if (!$config_identical) {
+        $this->logger->warning("The site's active config does not match the config in the sync directory.");
```
grasmash: I'd use the same language as drush `The active configuration is not identical to the configuration in the export directory`.

dooleymatt:
Awesome! Thanks for the feedback @grasmash! I'll get to work on those changes.

src/Robo/Inspector/Inspector.php
dooleymatt: @grasmash it seems like `$result->getExitCode()` won't work here. Even if the active config matches the export dir config, that results in an exit code of `1` since the config isn't exported.

lahoosascoots:
@dooleymatt Outside of the actual code issues, the is functionally exactly what I'm looking for. Thanks!

PR #2390, opened 2017-12-20
Override memcache config on Acquia Cloud.

grasmash:


settings/cache.settings.php
```diff
@@ -19,9 +20,107 @@
...
  * Uncomment the relevant lines for each bin below and configure the desired
- * See https://www.drupal.org/node/2754947.
+ *
+ * @see https://www.drupal.org/node/2754947
  */
 
 // $settings['cache']['bins']['bootstrap'] = 'cache.backend.null';
 // $settings['cache']['bins']['discovery'] = 'cache.backend.null';
 // $settings['cache']['bins']['config'] = 'cache.backend.null';.
+/**
```
greylabel: I think whitespace seems off here with these comments.

settings/cache.settings.php
```diff
@@ -19,9 +20,107 @@
...
+ *
+ * Autoload memcache classes and service container in case module is not
+ * installed. Avoids the need to patch core and allows for overriding the
+ * default backend when installing Drupal.
+ *
+ * @see https://www.drupal.org/node/2766509
+ */
+if ($is_ah_env) {
+  switch ($ah_env) {
+    case 'test':
```
greylabel: Do we want to keep this just `test` and `prod`?

settings/cache.settings.php
```diff
@@ -19,9 +20,107 @@
...
+    case 'prod':
+
+      // Check for PHP Memcached libraries.
+      $memcache_exists = class_exists('Memcache', FALSE);
+      $memcached_exists = class_exists('Memcached', FALSE);
+      $memcache_module_is_present = file_exists(DRUPAL_ROOT . '/modules/contrib/memcache/memcache.services.yml');
```
greylabel: The DrupalVM example should be updated to match this logic.

This only real difference is that on DrupalVM you need this:

```
// Include a unique prefix for each local Drupal installation.
if ($is_local_env) {
  $settings['memcache']['key_prefix'] = $site_dir;
}
```

settings/cache.settings.php
grasmash: I assume this condition is related to the text: 

> it can be used only in environments that have Drupal installed and the Memcache module enabled, __and won't need to be reinstalled, e.g. stage and prod.__

Does memcache usage cause issues when Drupal needs to be reinstalled?

lcatlett: Rather than using `test` and `prod` it will be more failsafe to use `($is_ah_env) && (isset($settings['memcache']['servers']))` in this check because it will work on any environment that has memcache servers available in settings.php which a variable set by acquia included config. This approach will also add support for any customer or non-standard environment names such as "uat" and ACSF environments.

lcatlett: @grasmash regarding memcache configuration in settings.php causing issues with installing or reinstalling drupal, the config in BLT memcache.php fixes issue by autoloading the required memcache services and classes 
and using memcache as a backend for static cache bins. If not using this configuration, there will be issues with site install, reinstalls, and adding memcache config to settings.php prior to the module being enabled since drupal boostraps the service container to the database. 

To clarify this I've submit proposed updates to the memcache module docs: https://www.drupal.org/project/memcache/issues/2934916#comment-12409521 and commented on the core issue to clarify: https://www.drupal.org/project/drupal/issues/2766509#comment-12409542

Issue #2378, opened 2017-12-15
Cannot install blt in existing project due to Class 'Symfony\Component\Debug\Exception\FatalThrowableError' not found


derhasi:
I wanted to add _blt_ in a project I currently develop locally based on drupal-composer/drupal-project and opensocial. When trying to add blt via instructions from the documentation this fails.

My composer.json right now:
```
{
    "name": "derhasi/kartslalom",
    "description": "Kartslalom.org project",
    "type": "project",
    "license": "GPL-2.0+",
    "authors": [
        {
            "name": "Johannes Haseitl",
            "role": ""
        }
    ],
    "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        },
        {
            "type": "composer",
            "url": "https://asset-packagist.org"
        }
    ],
    "require": {
        "acquia/blt": "^8.3",
        "composer/installers": "^1.0.20",
        "cweagans/composer-patches": "~1.0",
        "drupal-composer/drupal-scaffold": "^2.0.1",
        "drupal/address": "~1.0",
        "drupal/config_installer": "1.*",
        "drupal/core": "~8.0",
        "drupal/csv_serialization": "^1.0",
        "drupal/eva": "^1.2",
        "drupal/migrate_plus": "*",
        "drupal/migrate_tools": "*",
        "drupal/migrate_upgrade": "*",
        "drupal/pathauto": "^1.0",
        "drupal/redirect": "^1.0@beta",
        "drush/drush": "8.*",
        "goalgorilla/open_social": "^1.8.0",
        "league/csv": ">=7.1",
        "neitanod/forceutf8": "~2.0"
    },
    "require-dev": {
        "drupal/console": "^1.3.1",
        "drupal/devel": "^1.0@RC"
    },
    "conflict": {
        "drupal/drupal": "*"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "autoload": {
        "classmap": [
            "scripts/composer/ScriptHandler.php"
        ]
    },
    "scripts": {
        "drupal:scaffold": "DrupalComposer\\DrupalScaffold\\Plugin::scaffold",
        "post-install-cmd": [
            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
        ],
        "post-update-cmd": [
            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
        ],
        "site:install": [
            "chmod 777 docroot/sites/default docroot/sites/default/settings.php",
[...]

PR #2401, opened 2018-01-02
Fix and update readme/creating-new-project.md

TravisCarden:
`readme/creating-new-project.md` has a few issues:

1. ~~It's missing a `composer install`. Following the steps _exactly as written_ today will result in an error.~~
1. The list formatting is broken.
1. There's [a minor grammar nerd issue](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/). 😛

### DIFF
--- a/readme/creating-new-project.md
+++ b/readme/creating-new-project.md
@@ -2,7 +2,7 @@
 
 *Do not clone BLT or BLT Project as a means of using them. Just follow the directions below.*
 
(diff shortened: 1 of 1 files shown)


grasmash:
`composer install` should not be necessary?

grasmash:
`composer install` should not be necessary? I think you had to run it to workaround https://github.com/acquia/blt/issues/2378. The fix for that needs to be released.

grasmash:
Fix released. `composer install` should not be necessary now.

TravisCarden:
Thanks, @grasmash. I've removed that commit.

PR #2402, opened 2018-01-02
Replace drush.wrapper until we support Drush 9

TravisCarden:
Reverts https://github.com/acquia/blt/pull/1987.


grasmash:
Ack, this should have into 8.9.x, not 9.1.x.

PR #2408, opened 2018-01-04
Fixing the exit_early command.

### LINKED ISSUE #2403: Travis exit_early script makes it impossible to merge a PR
I'm using BLT and Travis. I updated some docs and opened a PR. Travis is reporting the build as a failure based on the output of exit_early. 

Should `exit_early` return 0 to indicate the build should be marked as Success if triggered? Or maybe use the `travis_terminate 0` command?

```sh
if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(\.md$)'
then
  if [ "$TRAVIS_PULL_REQUEST" != "false" ]
  then
    echo "Only docs were updated, stopping build process."
    travis_terminate 0
  fi
fi
```

Got this thought from a [stackoverflow post](https://stackoverflow.com/questions/44087114/how-to-kill-travis-build-with-exit-code-0-using-command). Not sure if it works, can't find this command documented anywhere... there is also some thought/mention to using a `set +e` flag in this file.

---

My Travis build output:
image

dpagini:
Closes #2403.

First commit shows a `README.md` change only to show Travis builds stalling with existing code.

Second commit (_coming after failure_) introduces a new method of exiting early.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,3 +107,5 @@ Copyright (C) 2016 Acquia, Inc.
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation.
 
 This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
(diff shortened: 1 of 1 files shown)


dpagini:
Alright, well the exit_early command works fine in this repo, so I'm going to close. I have no idea why this is failing on my Travis...

PR #2423, opened 2018-01-08
Fix xdebug and nodejs on beetbox.

danepowell:
The nodejs and xdebug settings in the box config are completely ignored if you use beetbox. This is the root cause: https://github.com/geerlingguy/drupal-vm/issues/1649

tl;dr: you have to add components to installed_extras in order to configure them (regardless of whether beetbox provides them by default).


danepowell:
Sorry I missed the code style issue with the first commit. It looks like the later test fails are a systemic problem though, unrelated to this PR.

PR #2447, opened 2018-01-12
Remove export from setup:composer:install

steveworley:
Other composer installs from BLT do not enforce failure on patch which
can cause some issues eg. blt deploy will work but blt setup will fail
at composer install. Given that composer patches doesn't fully support ignoring patches dependent projects can introduce patches that may break throughout a build which could cause issues with developer on-boarding.

For consistency all composer installs should follow the same process.

- This can be enforced per project in composer.json with "composer-exit-on-patch-failure": true.

Changes proposed:

- Add `composer:install` command
- Update `setup:composer:install` to invoke `composer:install`.
- Update `deploy` to invoke `composer:install` without dev.
- Add `project.yml` key to control composer cache clearing (for deploy
and setup).
- Remove `export COMPOSER_EXIT_ON_PATCH_FAILURE=1` from the
`composer:install` command ([this can be added in composer.json](https://github.com/cweagans/composer-patches#error-handling) per
project).

### DIFF
--- a/src/Robo/Commands/Composer/ComposerCommand.php
+++ b/src/Robo/Commands/Composer/ComposerCommand.php
@@ -4,6 +4,7 @@ namespace Acquia\Blt\Robo\Commands\Composer;
 
 use Acquia\Blt\Robo\BltTasks;
 use Acquia\Blt\Robo\Exceptions\BltException;
+use Robo\Contract\VerbosityThresholdInterface;
 
 /**
  * Defines commands in the "composer:*" namespace.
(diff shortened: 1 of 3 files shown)


src/Robo/Commands/Composer/ComposerCommand.php
```diff
@@ -61,4 +62,37 @@ public function requirePackage($package_name, $package_version, $options = ['dev
...
+   *
+   * @param bool $dev
+   *   Whether or no dev packages should be installed.
+   *
+   * @return bool
+   *   If the command was successful.
+   *
+   * @command composer:install
+   */
+  public function install($dev = TRUE) {
```
grasmash: I think we want to use dev as an option rather than an argument. 

```
$options = ['dev' => TRUE]
```

See https://github.com/consolidation/annotated-command#option-default-values

PR #2448, opened 2018-01-12
Fix chrome on Travis.

grasmash:


grasmash:
See https://github.com/travis-ci/travis-ci/issues/8836

dpagini:
Is the real fix here just the change to sudo in the Travis file @grasmash?

We're still on `8.9.11` I believe, so are any of the other updates in this file _needed_ to get this to work? Or cleanup of some kind?

PR #2451, opened 2018-01-16
Adding command to generate stub config_split configuration/directories.

pixlkat:
@grasmash / @danepowell 

This PR creates a BLT command which creates configs/directories for standard environment-based splits (Local, CI, Dev, Stage, Prod). It will skip any split that already has a config file present in the default config directory. It also does not touch the config in `project.yml` for toggling modules on/off.

### DIFF
--- a/scripts/config-split/templates/config_split.config_split.env.yml.twig
+++ b/scripts/config-split/templates/config_split.config_split.env.yml.twig
@@ -1,0 +1,14 @@
+uuid: {{ uuid }}
+langcode: en
+status: true
+dependencies: {  }
+id: {{ id }}
+label: {{ name }}
+folder: ../config/envs/{{ id }}
+module: {  }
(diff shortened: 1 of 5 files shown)


scripts/config-split/templates/config_split.config_split.env.yml.twig
```diff
@@ -0,0 +1,14 @@
+uuid: {{ uuid }}
```
danepowell: I worry about what will happen if/when config split updates their configuration schema, and we also need to update this template to match.

I'm not sure of a better alternative though.

grasmash:
> we should remove the pre-existing CI directory

Makes sense to remove it.

pixlkat:
@grasmash - I removed the `ci` directory and moved the README up a level and changed the wording a bit.

cc @danepowell

PR #2457, opened 2018-01-18
Add command to clear local solr server.

lahoosascoots:
Changes proposed:
- Adds command to clear local solr server.

### DIFF
--- a/src/Robo/Commands/Solr/ClearLocalSolrIndexCommand.php
+++ b/src/Robo/Commands/Solr/ClearLocalSolrIndexCommand.php
@@ -17,11 +17,11 @@ class ClearLocalSolrIndexCommand extends BltTasks {
    */
   public function clearSolr() {
     $this->taskExecStack()
-      ->exec("curl -s http://local.ebsco.com:8983/solr/d8/update --data '<delete><query>*:*</query></delete>' -H 'Content-type:text/xml; charset=utf-8' > /dev/null")
+      ->exec("curl -s {$this->getConfigValue('project.local.uri')}:8983/solr/d8/update --data '<delete><query>*:*</query></delete>' -H 'Content-type:text/xml; charset=utf-8' > /dev/null")
       ->run();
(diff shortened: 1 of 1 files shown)


src/Robo/Commands/Solr/ClearLocalSolrIndexCommand.php
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * Clear local solr index.
+   *
+   * @command solr:clear
+   * @description Clears the local solr index.
+   */
+  public function clearSolr() {
+    $this->taskExecStack()
+      ->exec("curl -s http://local.ebsco.com:8983/solr/d8/update --data '<delete><query>*:*</query></delete>' -H 'Content-type:text/xml; charset=utf-8' > /dev/null")
```
greylabel: `ebsco.com`?

lahoosascoots: Ah whoops! Good catch. I should pull that from config.

src/Robo/Commands/Solr/ClearLocalSolrIndexCommand.php
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * Clear local solr index.
+   *
+   * @command solr:clear
+   * @description Clears the local solr index.
+   */
+  public function clearSolr() {
+    $this->taskExecStack()
+      ->exec("curl -s {$this->getConfigValue('project.local.uri')}:8983/solr/d8/update --data '<delete><query>*:*</query></delete>' -H 'Content-type:text/xml; charset=utf-8' > /dev/null")
```
grasmash: Is this always the correct port?

lahoosascoots: This is the standard solr port which we do not have a config variable for

Issue #2466, opened 2018-01-20
Running BLT VM did not generate Drush Alias File After Drush 9 Update


mikemadison13:
My system information:
* Operating system type: Mac OS
* Operating system version: 10.11.6
* BLT version: 9.0.0-alpha2

Output of `blt doctor`:
```
blt doctor
Drupal VM was detected. Running blt doctor inside of VM...
Executing command cd /var/www/<project>/docroot/.. && /var/www/<project>/docroot/../vendor/bin/drush cc drush && /var/www/<project>/docroot/../vendor/bin/drush --include=/var/www/<project>/docroot/../vendor/acquia/blt/drush blt-doctor -r /var/www/<project>/docroot/../docroot inside of Drupal VM...
 [success] 'drush' cache was cleared.

                     
  [Exception]        
  Bootstrap failed.  
                     

blt:doctor [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--no] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--druplicon] [--notify] [--xh-link XH-LINK] [--] <command>

Connection to 127.0.0.1 closed.
Attempting to run doctor on host machine...

```

When I run this command:
```
blt vm
```

And I expected this to happen:
I expected that the Drush 9 alias file would be generated for the new local / vm, but it was not.

mikemadison13:
digging into this, i believe the issue is that I had drush_server_home() in my old site alias file. 

I think we need to expand our documentation in drush/sites/README.md to help people troubleshoot this. It is unclear to me if BLT VM was necessary in this instance, or if blt update should have addressed this (re: https://github.com/acquia/blt/issues/2437). However, as far as I can tell, I had to manually remove drush_server_home() from the alias file, then move the file, then run the drush command to get a local alias.

see: https://github.com/acquia/blt/pull/2241/files

mikemadison13:
note that once i have the valid alias file, i am not getting:
`Permission denied (publickey,password).`

This may be related to https://github.com/geerlingguy/drupal-vm/issues/1595 ?

grasmash:
The upgrade hook should have actually created new Drush alias files in a Yaml format. Did that happen?

mikemadison13:
[...]

PR #2467, opened 2018-01-20
BLT-2374: re-enabling features import with drush 9 compatible version…

### LINKED ISSUE #2374: Re-implement or remove checkFeaturesOverrides
https://github.com/acquia/blt/blob/9.1.x/src/Robo/Commands/Setup/ConfigCommand.php#L90

mikemadison13:
… of features.

Fixes #2374  .

Changes proposed:
- add features to composer (suggested?) at hash: 2e82a6ad751fc88f771be5b4e53b98a8c6eb66b5 which includes https://www.drupal.org/project/features/issues/2913651
-remove early return in checkFeaturesOverrides() function in src/Robo/Commands/Setup/ConfigCommand.php to allow features import to continue running

### DIFF
--- a/composer.suggested.json
+++ b/composer.suggested.json
@@ -15,7 +15,8 @@
     "drupal/memcache": "^2.0-alpha5",
     "drupal/seckit": "^1.0.0-alpha2",
     "drupal/security_review": "*",
-    "drupal/shield": "^1.0.0"
+    "drupal/shield": "^1.0.0",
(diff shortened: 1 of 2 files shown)


mikemadison13:
@grasmash this "should" work but I was unable to fully test it due to https://github.com/acquia/blt/issues/2466 (and my inability to do an alias into my local environment with drush). 

also, do you have an opinion as to which of the composer files this goes into? i put it in suggested for now.

grasmash:
I think they going in the suggested composer file works fine, but we may want to check to ensure that features exists before we execute any commands.

PR #2487, opened 2018-01-24
Cleaning up wizard. (#2484)

grasmash:
* Porting remaining commands, adding aliases.

* Clean up aliases.

* Fixing validate command.


src/Robo/Commands/Setup/WizardCommand.php
```diff
@@ -100,9 +100,30 @@ protected function askForAnswers() {
...
+      ];
+      $answers['cm']['strategy'] = $this->askChoice('Choose a configuration management strategy:', $strategy_options, 'config-split');
+    }
+    else {
+      $answers['cm']['strategy'] = 'none';
+    }
+
+    $profile_options = [
+      'lightning' => 'Lightning',
+      'minimal' => 'Minimal',
```
greylabel: @grasmash I think `testing` should also be an option here.

PR #2508, opened 2018-01-31
make vm vs non-vm more clear

CashWilliams:

### DIFF
--- a/readme/onboarding.md
+++ b/readme/onboarding.md
@@ -25,7 +25,7 @@ You have probably been linked to this documentation by a project that is using B
 1. Run `composer install` (you must already have Composer installed).
 1. Install `blt` alias: `composer run-script blt-alias`. At this point you might need restart your shell in order for the alias work.
 
-If your project uses a virtual development environment such as DrupalVM:
+If your project has an existing virtual development environment configured, such as DrupalVM:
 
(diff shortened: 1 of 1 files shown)


readme/onboarding.md
```diff
@@ -25,7 +25,7 @@ You have probably been linked to this documentation by a project that is using B
 1. Run `composer install` (you must already have Composer installed).
 1. Install `blt` alias: `composer run-script blt-alias`. At this point you might need restart your shell in order for the alias work.
 
-If your project uses a virtual development environment such as DrupalVM:
+If your project has an existing virtual development environment configured, such as DrupalVM:
```
grasmash: Hm, even if you don't have a configured VM, you can still run this and it will configure it for you.

CashWilliams: thats fair. what was unclear to me when I originally read this was that this command is to start your _already configured_ VM. I read it was it would make new VM config for me, but the project I pulled down already had VM config and I wanted to use that, so this command felt like the wrong one based on this wording.

CashWilliams: instead I saw the VagrantFile and figured I should just run `vagrant up` which broken things

grasmash:
https://github.com/acquia/blt/commit/4cc786569e9c9fea9654e64dcc02f435ba1d286e

PR #2521, opened 2018-02-05
Make allow_failure step in .travis.yml reachable

dpagini:
Posting this as PR b/c I need it as a patch.

Not sure if this should be the extent of the change, or if other bash scripts should change as well, and what other branches this should be opened against...? Any direction here is welcome...

Here is my issue and what I'm trying to solve.
I use a Travis `after_failure` script to upload any behat errors to a 3rd party site. At some point, this stopped working, and I began trying to figure out why. I started by pairing my Travis file down to the bare minimum, and `after_failure` was executing for me. I found that by adding back my `install` scripts (which include the `setup_project` and `setup_environment` scripts), my `after_failure` script stopped running. Based on some Travis reading, a call to `set -e` in any file is changing the behavior of the entire shell, and any error bubbles up and stops the script. I found that by ending these scripts with a `set +e` that it is allowing me to correctly finish the build and hit the `after_failure` script, and still let the build end in error.

I hope that was a clear explanation, but I can try to give more details if possible.

### DIFF
--- a/scripts/travis/.travis.yml
+++ b/scripts/travis/.travis.yml
@@ -55,7 +55,7 @@ install:
   - source ${BLT_DIR}/scripts/travis/setup_project
 
 script:
-  - source ${BLT_DIR}/scripts/travis/run_tests
+  - ${BLT_DIR}/scripts/travis/run_tests
 
 deploy:
    - provider: script
--- a/scripts/travis/deploy_branch
+++ b/scripts/travis/deploy_branch
@@ -4,4 +4,4 @@ set -ev
 
 blt deploy --commit-msg "Automated commit by Travis CI for Build ${TRAVIS_BUILD_ID}" --branch "${TRAVIS_BRANCH}-build" --ignore-dirty --no-interaction -v
 
-set +v
+set +ev
--- a/scripts/travis/deploy_tag
+++ b/scripts/travis/deploy_tag
@@ -4,4 +4,4 @@ set -ev
 
 blt deploy --commit-msg "Automated commit by Travis CI for Build ${TRAVIS_TAG}" --tag "${TRAVIS_TAG}-build" --ignore-dirty -v
 
-set +v
+set +ev
--- a/scripts/travis/run_tests
+++ b/scripts/travis/run_tests
@@ -9,4 +9,4 @@ blt validate:all
 ${BLT_DIR}/scripts/travis/tick-tock.sh blt setup --define drush.alias='${drush.aliases.ci}' --define environment=ci --no-interaction --yes -v
 blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --define environment=ci --yes -v
 
-set +v
+set +ev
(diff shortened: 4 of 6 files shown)


dpagini:
Ok, so I don't think this is a total fix, but I still think it's relevant...

I learned a bit about bash while working on this. So using `$ source filename` syntax, you are running the script in the context of the same shell you are currently in. I think this makes sense, for example, in the `setup_environment` script, where we are setting the path. For the `run_tests` command, I'm not sure this is the best choice. It is here where using `set -e` _bleeds_ to the current shell environment.

I think [this line](https://github.com/acquia/blt/blob/9.x/scripts/travis/.travis.yml#L57) should execute the script as a standalone program instead of "sourcing" the script.

That would change:
```
script: 
  - source ${BLT_DIR}/scripts/travis/run_tests
```
to just:
```
script:
  - ${BLT_DIR}/scripts/travis/run_tests
```

dpagini:
Just updated this PR to include the suggested Travis file change as well. @grasmash any thoughts on this approach? If this works for you, I can open a PR against 9.x as well...? Thanks!

grasmash:
So I think that part of the reason we use `source` is so that travis shows the individual commands in the script that are called. Once you remove it, the whole script gets wrapped up in a singe line call that can be harder to debug.

dpagini:
Hmm... so what about changes to the `set -e` logic? Is it possible maybe these scripts have some checks at the bottom for exit codes instead of using that directive at the top?
Or what about just a BLT command to fire `run_tests` instead of a custom script? So something like `$ blt tests:ci:travis` that then calls `validate:all && setup && tests:all`? That command could output what's needed and set the appropriate flags.

Sorry, just getting some ideas out there. I think you understand my problem though based on the issue title change. LMK if any of those ideas sound appropriate and I can pursue...

grasmash:
>  Is it possible maybe these scripts have some checks at the bottom for exit codes instead of using that directive at the top?

I'd be fine with that. You could check for failure and return from the script prematurely.

dpagini:
Hmm... I just don't know how to do that piece. Would this be better as a closed PR and open issue request?
I think I'm still seeing the same thing in `9.0` now where I can't get `after_failure` to trigger in Travis.

danepowell:
> So I think that part of the reason we use source is so that travis shows the individual commands in the script that are called. Once you remove it, the whole script gets wrapped up in a singe line call that can be harder to debug.

I just tested this and can't seem to reproduce this observation, I can see no difference in the formatting of the Travis output using `source` or not.

Does anyone know if perhaps Travis changed something? I'd like to merge this, but want to make sure I'm not overlooking something in relation to Matt's concern.

danepowell:
Ah... regardless, this would need to be reopened against 9.x, so I'll close this for now. Feel free to reopen if we can confirm the Travis behavior.

dpagini:
I'll take a look at this again if I can... It's difficult to encourage contribution when a PR sits out there for 6 months at a time and then gets closed. =/

danepowell:
I'm sorry about that, and I totally understand your frustration, I'd feel the same way. We're trying to catch up on the backlog of issues but realistically it will probably take a while.

PR #2539, opened 2018-02-11
Allow Chromium for Behat tests.

kviolka:
On Debian we have Chromium instead of Google Chrome. This patch allows using it.


kviolka:
I'm using BLT 8.x right now, but it should work also with 9, right? :)

grasmash:
8.x is a deprecated branch. Please use 8.9.x.

PR #2552, opened 2018-02-13
Moving cloud hook logic to PHP.

grasmash:


danepowell:
This fails to correctly detect ACSF multi-stack tenants, and tries to run Cloud Hooks leading to errors like https://github.com/acquia/blt/issues/3565

I'm wondering if there's a reason we didn't use the existing logic to detect ACSF envs here? https://github.com/acquia/blt/pull/2313

PR #2566, opened 2018-02-15
Fixes #2565 to allow config to be imported for existing sites.

### LINKED ISSUE #2565: Allow site to be installed from existing config.
My system information:
* Operating system type: [replace]
* Operating system version: [replace]
* BLT version: 9.x-dev

When I run this command:
```
blt setup:config-import
```

I get the following output:
```
Site UUID in source storage does not match the target storage. 
```

OR (depending on approach)
```
[notice] Starting Drupal installation. This takes a while. Consider using the --notify global option.
 [error]  The "taxonomy_term" entity type does not exist. 
```

(yes, i trimmed the output). 

after extensively experimenting with both https://github.com/acquia/blt/wiki/Upgrading-from-Drush-8-to-Drush-9#installing-site-from-existing-configuration and https://lightning.acquia.com/blog/using-configuration-installer-lightning, I continue to hit major issues with existing configuration.

**Proposal**
Let's drush cset a UUID in-flight during config-import. 

This is fairly straight forward code...
```
$cm_uuid = $this->getConfigValue('cm.uuid');
      if ($cm_uuid != NULL) {
        $task = $this->taskDrush()
          ->drush("config:set system.site uuid $cm_uuid");
      }
```
...that can easily be run as part of the blt setup:config-import command so that we don't have to do all of the installation profile, minimal profile, lightning config, whatever other changes would be required to make an existing project work. This allows configuration to be imported as it was with Drush 8 / BLT 8 without any other fancy solutions. 

I've tried to figure out a way to automatically write the UUID into the project file based on what is in system.site.yml (not sure if that can be done during an update). But otherwise, all we'd have to do is have the UUID written into the project yml file once at the beginning of the project and theoretically no other fanciness would be required.

mikemadison13:
Fixes #2565  .

Changes proposed:
- sets the UUID in current drupal install to match "config" version as defined in project / blt yml file
- adds cm.uuid to the templates/blt.yml file

### DIFF
--- a/src/Robo/Commands/Setup/ConfigCommand.php
+++ b/src/Robo/Commands/Setup/ConfigCommand.php
@@ -66,6 +66,14 @@ class ConfigCommand extends BltTasks {
         // necessary configuration file(s) as part of the db update.
         ->drush("updb");
 
+      // Retrieve the config UUID and set site UUID to allow config import
+      // for existing sites.
+      $cm_uuid = $this->getConfigValue('cm.uuid');
+      if ($cm_uuid != NULL) {
+        $task = $this->taskDrush()
+          ->drush("config:set system.site uuid $cm_uuid");
+      }
+
       switch ($strategy) {
         case 'core-only':
           $this->importCoreOnly($task, $cm_core_key);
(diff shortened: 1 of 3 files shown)


grasmash:
Replaced by #2590

PR #2570, opened 2018-02-23
Update patch to clear Twig caches on deploys

anavarre:
The patch BLT is using to clear Twig caches on deploys is very outdated and even failed tests. It would seem using the latest patch (which passes tests) would be better.

Before committing this change **it'd be ideal to confirm everything works as anticipated**.

### DIFF
--- a/composer.required.json
+++ b/composer.required.json
@@ -50,7 +50,7 @@
     "enable-patching": true,
     "patches": {
       "drupal/core": {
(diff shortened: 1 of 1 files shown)


anavarre:
@balsama FYI.

grasmash:
I'm going to commit since we're not cutting stable releases from this branch yet. Cloud testing will be part of prerelease testing.

PR #2571, opened 2018-02-23
Fixes #1486: Add option to allow validating frontend files.

### LINKED ISSUE #1486: Frontend linting via git commit hooks
Currently we run PHPCS on all code committed, but we don't do anything similar for frontend code.  We have coding standards for JS, and SCSS.  We should come up with a way to easily add them to the git commit hooks.

One idea would be to add a custom hook to project.yml that can be invoked by the git commit hooks.

nikunjkotecha:
Changes proposed:
- Allow validation of frontend files.

Ref: https://github.com/acquia/blt/issues/1486

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -160,6 +160,11 @@ target-hooks:
     dir: ${docroot}
     # E.g., '[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" && nvm use 4.4.4 && npm build'
     command: null
+  frontend-validate:
+      # E.g., ${docroot}/sites/all/themes/custom/mytheme.
+      dir: ${docroot}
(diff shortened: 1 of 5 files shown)


nikunjkotecha:
This is not working as expected as commands that are executed inside VM don't return exit code.

nikunjkotecha:
Verified it works great, might be able to fix many other issues too.

danepowell:
As mentioned in the upstream issue, we already have a frontend-test hook that should serve basically the same role.

However, what's currently missing is the ability to run validation/tests on git commits. I think this is worth having, and I would suggest opening a new issue to this effect. That would probably require a separate validation hook, since tests might take too long to run on commit.

Regardless, I unfortunately have to close this PR because 8.9.x is not getting new features. A new PR would have to go against 9.x.

PR #2583, opened 2018-02-27
Disable Behat extension for Drupal autoloading.

grasmash:


balsama:
@grasmash I _think_ this is unnecessary now that Lightning has moved its subcontexts. See https://github.com/acquia/lightning-core/pull/18

PR #2592, opened 2018-03-01
Use HEAD of main Lightning branch

balsama:
HEAD of the main Lightning branch (8.x-3.x - from which we will cut the 3.1.0 tag) now requires core ~8.5.0 - so we can use that instead of the 3.1.x testing branch.

### DIFF
--- a/composer.required.json
+++ b/composer.required.json
@@ -22,7 +22,7 @@
     "dmore/chrome-mink-driver": "2.4.0",
     "dmore/behat-chrome-extension": "^1.0.0",
     "mikey179/vfsStream": "~1.2",
-    "sensiolabs-de/deprecation-detector": "^0.1.0-alpha4"
+    "sensiolabs-de/deprecation-detector": "dev-master"
   },
   "autoload-dev": {
     "psr-4": {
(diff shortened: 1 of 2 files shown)


balsama:
Looks like Lightning Core is causing this failure. Let's see if this fixes it:
https://github.com/acquia/lightning-core/pull/28

balsama:
Ok. Looks like we have a real conflict here.

`sensiolabs-de/deprecation-detector` requires `phpdocumentor/reflection-docblock: ~2.0`. Lightning needs > 3.x of `phpdocumentor/reflection-docblock` because 3.x introduced `DocBlockFactory`.

@grasmash Can you give some insight into `sensiolabs-de/deprecation-detector`? Can that be solved any other way? Otherwise, we'll have to rewrite Lightning's update system.

balsama:
@grasmash Looks like HEAD of deprecation-detector allows docblock 3.x or 4.x - they just haven't tagged a release in two years 👎 

https://github.com/sensiolabs-de/deprecation-detector/commit/6f85e00abca65cd211bcc2df6eee49b1efb8d2c4

Can we chase HEAD of that maybe?

balsama:
https://github.com/sensiolabs-de/deprecation-detector/issues/142

PR #2600, opened 2018-03-05
Stub out quickdev command.

grasmash:


grasmash:
Closing given that I'm transitioning away from BLT work and I have no plans to follow through on this PR.

PR #2601, opened 2018-03-05
Require PHP 7 minimum.

grasmash:


japerry:
https://github.com/acquia/blt/issues/2661

PR #2611, opened 2018-03-08
Require PHP 7.1 in all the right places.

grasmash:
This is considered a bug fix. 9.0.0 should have required this since it's already the default PHP version in Drupal VM and acquia-pipelines.yml. Without this fix, it's possible to build dependencies in the VM that cannot be installed on Travis CI or run in the cloud.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "type": "composer-plugin",
     "license": "GPL-2.0-only",
     "require": {
-        "php": ">=7",
+        "php": ">=7.1",
         "composer-plugin-api": "^1.0.0",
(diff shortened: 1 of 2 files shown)


scor:
If blt 8.9 is going to EOL in 2 months (5/8/18), that means blt users will have no choice but to upgrade from PHP 5.6 to PHP 7.1 after that point. Afaik, there are still a lot of site running PHP 5.6, and I don't see them moving to PHP 7.1 that easily. What is the rush for making such a sudden jump in PHP requirement? Drupal core still supports PHP 5 and won't require PHP 7 until Drupal 8.7.

Issue #2661, opened 2018-03-21
BLT needs to support our minimum php version requirement (5.6)


japerry:
My system information:
* Operating system type: Linux 
* Operating system version: W/e, its running PHP 5.6
* BLT version: 9.0.0-beta3 -> 9.0.5

When I run this command:
```
composer update acquia/blt
```

I get the following output:
```
Gathering patches for root package.
Loading composer repositories with package information
Updating dependencies (including require-dev)                                 
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for acquia/blt ~9.0.5 -> satisfiable by acquia/blt[9.0.5].
    - acquia/blt 9.0.5 requires php >=7.1 -> your PHP version (5.6.30) does not satisfy that requirement.
  Problem 2
    - Conclusion: don't install phpdocumentor/reflection-docblock 2.0.5
    - Conclusion: don't install phpdocumentor/reflection-docblock 3.0.3|install phpdocumentor/reflection-docblock 2.0.5
    - Conclusion: remove acquia/lightning 3.1.0
    - Conclusion: don't install acquia/lightning 3.1.0
    - acquia/lightning 3.1.x-dev requires drupal/lightning_core 2.x-dev -> satisfiable by drupal/lightning_core[2.x-dev].
    - Conclusion: don't install phpdocumentor/reflection-docblock 3.0.2
    - Conclusion: don't install phpdocumentor/reflection-docblock 2.0.4
    - Installation request for acquia/lightning ^3.1 -> satisfiable by acquia/lightning[3.1.0, 3.1.x-dev].
    - Conclusion: don't install phpdocumentor/reflection-docblock 3.0.1|install phpdocumentor/reflection-docblock 2.0.4
    - Installation request for sensiolabs-de/deprecation-detector ^0.1.0-alpha4 -> satisfiable by sensiolabs-de/deprecation-detector[0.1.0-alpha4].
    - sensiolabs-de/deprecation-detector 0.1.0-alpha4 requires phpdocumentor/reflection-docblock ~2.0 -> satisfiable by phpdocumentor/reflection-docblock[2.0.0, 2.0.0a1, 2.0.0a2, 2.0.0a3, 2.0.1, 2.0.2, 2.0.3, 2.0.4, 2.0.5].
    - Can only install one of: phpdocumentor/reflection-docblock[3.0.0, 2.0.0].
    - Can only install one of: phpdocumentor/reflection-docblock[3.0.0, 2.0.0a1].
[...]

PR #2665, opened 2018-03-21
Revert back to requiring php 5.6 as the minimum php version

japerry:
This reverts back all the php7 changes that was performed after beta3 was released.

Note: We should be documenting that if you're using php7 locally, you'll need to manually update configuration files to support php7. While its unfortunate that we support both versions of PHP, we must continue to do so until our product roadmap makes it EOL, which is slated for the end of 2018.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: php
 dist: trusty
 
 php:
-  - 7.1
+  - 5.6
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,5 +83,5 @@ deploy:
     skip_cleanup: true
     on:
       branch: $DEPLOY_SOURCE_BRANCH
-      php: 7.1
+      php: 5.6
       condition: $DRUPAL_CORE_VERSION = default
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: trusty
 
 php:
   - 5.6
+  - 7.1
 
 env:
   global:
--- a/scripts/drupal-vm/config.yml
+++ b/scripts/drupal-vm/config.yml
@@ -74,7 +74,7 @@ npm_config_prefix: "/home/{{ drupalvm_user }}/.npm-global"
 installed_extras: ${installed_extras}
 
 # PHP 7
-php_version: "7.1"
+php_version: "5.6"
 php_packages_extra:
   - "php{{ php_version }}-bz2"
   - "php{{ php_version }}-imagick"
--- a/scripts/travis/.travis.yml
+++ b/scripts/travis/.travis.yml
@@ -1,4 +1,4 @@
-sudo: true
+sudo: false
 language: php
 dist: trusty
 
@@ -25,6 +25,7 @@ cache:
   - "$HOME/.drush/cache"
   - "$HOME/.npm"
   - "$HOME/.nvm"
+  - "vendor"
   - ".rules"
   # Cache front end dependencies to dramatically improve build time.
   # - "docroot/themes/custom/mytheme/node_modules"
@@ -62,16 +63,16 @@ deploy:
      skip_cleanup: true
      on:
        branch: develop
-       php: 7.1
+       php: 5.6
    - provider: script
      script: "${BLT_DIR}/scripts/travis/deploy_branch"
      skip_cleanup: true
      on:
        branch: master
-       php: 7.1
+       php: 5.6
    - provider: script
      script: "${BLT_DIR}/scripts/travis/deploy_tag"
      skip_cleanup: true
      on:
        tags: true
-       php: 7.1
+       php: 5.6
--- a/scripts/travis/.travis.yml
+++ b/scripts/travis/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
 language: php
 dist: trusty
 
@@ -28,7 +28,6 @@ cache:
   - "$HOME/.drush/cache"
   - "$HOME/.npm"
   - "$HOME/.nvm"
-  - "vendor"
   - ".rules"
   # Cache front end dependencies to dramatically improve build time.
   # - "docroot/themes/custom/mytheme/node_modules"
--- a/scripts/travis/.travis.yml
+++ b/scripts/travis/.travis.yml
@@ -2,8 +2,11 @@ sudo: false
 language: php
 dist: trusty
 
+# Adjust the version of PHP to match your production environment.
+# Making this version number greater than the production environment can unintended consequences including a
+# non-functional prod environment.
 php:
-  - 7.1
+  - 5.6
 
 matrix:
   fast_finish: true
@@ -57,6 +60,7 @@ install:
 script:
   - source ${BLT_DIR}/scripts/travis/run_tests
 
+# The version below MUST match the php version indicated at the top of this file, otherwise the artifact won't build.
 deploy:
    - provider: script
      script: "${BLT_DIR}/scripts/travis/deploy_branch"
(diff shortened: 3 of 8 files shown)


grasmash:
I think we'd need to change Drupal VM's default version of PHP, or else somehow enforce a composer platform requirement of PHP 7 when Drupal VM is used.

https://github.com/acquia/blt/blob/9.x/scripts/drupal-vm/config.yml

grasmash:
Also need to change PHP version in:

- https://github.com/acquia/blt/blob/9.x/.travis.yml
- https://github.com/acquia/blt/blob/9.x/acquia-pipelines.yml

japerry:
Ahh yup yup, I've gone and made those updates. I didn't see anything else in a search/replace for 7.1 or 7 although that had a lot of noise. We'll see what travis thinks ;)

src/Robo/Inspector/Inspector.php
```diff
@@ -788,7 +788,7 @@ public function issueEnvironmentWarnings($command_name) {
    * @throws \Acquia\Blt\Robo\Exceptions\BltException
    */
   public function warnIfPhpOutdated() {
-    $minimum_php_version = 7;
+    $minimum_php_version = 5.6;
```
dpagini: Why not keep a warning in here, but don't throw an exception? Maybe suggest users to move off 5.6 at their convenience...

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: php
 dist: trusty
 
 php:
-  - 7.1
+  - 5.6
```
dpagini: Isn't it possible to test with both versions? via the matrix? I don't know enough about it, but just seems like it would be worth considering.

src/Robo/Inspector/Inspector.php
greylabel: Or, have a warning and suggest users move _to_ 5.6 if they really need it, but not encourage it as a default. Otherwise, this entire issue feels like a step backwards.

grasmash:
Weird, the last build exited prematurely.

grasmash:
Cleared caches and restarted builds.

grasmash:
Passing.

However, I'm still not convinced that this should be merged. Can we continue the conversation here? https://github.com/acquia/blt/issues/2661

I would like to understand what the blockers are for updating to PHP 7, and also why extending LTS for 8.9.x is not sufficient.

scripts/travis/.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: true
+sudo: false
```
grasmash: Need to update line 6. Also, is this sudo change necessary?

scripts/drupal-vm/config.yml
```diff
@@ -74,7 +74,7 @@ npm_config_prefix: "/home/{{ drupalvm_user }}/.npm-global"
 installed_extras: ${installed_extras}
 
 # PHP 7
-php_version: "7.1"
+php_version: "5.6"
```
grasmash: @geerlingguy Are more changes than this required?

grasmash:
Pinging @danepowell for visibility.

scripts/drupal-vm/config.yml
geerlingguy: No not for the VM... but do we really want to *default* to 5.6? Shouldn’t we default to whatever Acquia Cloud currently defaults to (7.1 IIRC)?

grasmash: The trouble is that if your host is 5.6 and your VM is 7, you get some issues with Composer. Maybe we should look into defaulting the "config" in composer.json to PHP 7. I know that adding it to `template/composer.json` and `composer.required.json` do not work. We'd need to actually modify the generated `composer.json` after project creation.

dpagini: I actually run 5.6 on host and 7 in VM. TBH, I haven't really had a problem with it... I usually run `$ composer` commands on the host, and BLT will run composer in the VM... knock on wood but I haven't seen much yet. Just wanted to share... ha.

grasmash: Do you use the platform config setting in composer.json?

danepowell: Most people I've seen (including myself) run a higher PHP version on the host than the VM. i.e. 7.2 on the host, and 5.6 or 7.1 in the VM. The reason being pretty obvious, that the VM replicates an Acquia Cloud or other hosting environment that generally lags behind general desktop software in terms of PHP releases.

I'd really like to see the VM default to the highest version possible in order to match the host machine, or at least never be "behind" the host, which leads to some nasty problems. For instance, running `composer update` on the host breaks everything.

dpagini: > Do you use the platform config setting in composer.json?

I don't. I mean, below is my output, haha. I haven't upgraded my host b/c I've just been lazy, TBH, and I haven't seen a need to. I do _a lot_ of composer updates for our project too, and just haven't run into any issues...

If you need me to try anything out, LMK - I'm happy to report back.

```sh
$ php -v
PHP 5.6.31 (cli) (built: Aug 17 2017 16:41:40) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.5.3, Copyright (c) 2002-2017, by Derick Rethans

$ vagrant ssh -c 'php -v'
PHP 7.0.20-2~ubuntu16.04.1+deb.sury.org+1 (cli) (built: Jun 14 2017 05:33:01) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.20-2~ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.5.0, Copyright (c) 2002-2016, by Derick Rethans

```

scripts/travis/.travis.yml
grasmash: Line 6 of this file also needs to be 5.6.

scripts/travis/.travis.yml
```diff
@@ -25,6 +25,7 @@ cache:
   - "$HOME/.drush/cache"
   - "$HOME/.npm"
   - "$HOME/.nvm"
+  - "vendor"
```
grasmash: Why add this back?

japerry: I was just reverting the changes associated with php7, if this wasn't meant to be part of that commit (or was two commits in one), then this should be taken out of the PR.

scripts/drupal-vm/config.yml
japerry: Hmm IMHO, except for grabbing the initial code you really should never run composer on the host environment. The part that matters here is that your VM matches your Production environment. If I could have my host not have ANY dev tools (php, drush, composer, etc) I would. 

While Acquia Cloud -defaults- to 7.1 for new instances, it certainly isn't what most of our clients are running. And until ACE+ACSF are fully on 5.6, we shouldn't be shipping a VM that can break prod.. which is the whole point of Vagrant in the first place. Since 7.0 much more backwards compatible with 5.6 than visa-versa, we need to use the least common denominator.

src/Robo/Inspector/Inspector.php
japerry: This function is not currently a warning but rather an exception, causing BLT to fail. I agree -- giving some warning that you're running 5.6 would be good, although Drupal already does that.

scripts/travis/.travis.yml
japerry: Fixed! I also added a doc comment to advise the user this should match whatever version their prod environment is in.

scripts/travis/.travis.yml
grasmash: Let's not add it back.

PR #2683, opened 2018-03-26
Remove over-aggressive "Icon" .gitignore rule (9.x)

TravisCarden:
The `Icon` rule in `blt/scripts/blt/deploy/.gitignore` is too aggressive. It causes the `docroot/core/lib/Drupal/Core/Layout/Icon` directory that's now a part of Drupal core to be omitted from the build artifact.

c.f. https://github.com/acquia/blt/pull/2684 for PR to 8.9.x branch.

### DIFF
--- a/scripts/blt/deploy/.gitignore
+++ b/scripts/blt/deploy/.gitignore
@@ -16,7 +16,6 @@ reports
 .DS_Store
 .AppleDouble
 .LSOverride
-Icon
 
 # Thumbnails
 ._*


TravisCarden:
It sounds good in theory, but I couldn't get it to work. It looks like we've done something similar in https://github.com/acquia/blt/blob/5d0079de5bd0972cda13dbeccfdddf1ef54f4cac/scripts/simplesamlphp/gitignore.txt, but it's not what the article led me to expect, and duplicating it in this case didn't get results.

PR #2690, opened 2018-03-28
Fixes #2686: Add deployment_identifier to project gitignore.

### LINKED ISSUE #2686: Local deployment_identifier
My system information:
* BLT version: 9.0.5

I'm seeing a deployment_identifier file in my repo root that gets updated periodically, which is annoying because it's tracked by Git.

I know we originally added this file for cache-busting purposes, but only for deploys. My guess is that on local installs, this file shouldn't be present, or should be gitignored.

This is a fairly new installation of BLT 9.0.5.

malikkotob:


src/Update/Updates.php
```diff
@@ -534,4 +534,22 @@ public function update_9000000() {
     $this->updater->getOutput()->writeln($formattedBlock);
     $this->updater->getOutput()->writeln("");
   }
+
+  /**
+   * 9.0.6.
```
grasmash: This will actually be going into 9.1.0.

malikkotob: 👍 fixed

PR #2694, opened 2018-03-29
BLT always reports configuration overridden

danepowell:
Whenever I run `blt setup`, it always complains that configuration has been overridden, even though it hasn't:

>[warning] The active configuration is not identical to the configuration in the export directory.
[warning] This means that you have not exported all of your active configuration.
[warning] Run drush cex to export the active config to the sync directory.
[warning] Continuing will overwrite the active configuration.

The problem is that getMessage() in the inspector class returns null when configuration is not overridden, but BLT expects it to return a string. I'm not sure if this ever worked, or if it used to work what changed to break it--maybe something in Drush or how Robo is returning strings. Might be worth looking into.

Anyway, this PR fixes it.

I'm running drush 9.2.1 and blt 9.0.5

### DIFF
--- a/src/Robo/Inspector/Inspector.php
+++ b/src/Robo/Inspector/Inspector.php
@@ -814,7 +814,7 @@ class Inspector implements BuilderAwareInterface, ConfigAwareInterface, Containe
   public function isActiveConfigIdentical() {
     $result = $this->executor->drush("config:status")->run();
     $message = trim($result->getMessage());
-    $identical = strstr($message, 'No differences between DB and sync directory') !== FALSE;
+    $identical = empty($message);
 
     return $identical;
   }


grasmash:
Related: https://github.com/acquia/blt/pull/2680

grasmash:
Wait, why doesn't Drush return the string `No differences between DB and sync directory`? Maybe we need to add `--format=json`? Or would an exit code be better?

Relying on a null output feels a little unstable.

danepowell:
Drush _does_ return the "No differences between DB and sync directory" string if you just run `drush config:status`, but it's not getting slurped up by Robo / Inspector for whatever reason.

Robo sees a null string if configuration is identical, and a list of changes of configuration is not identical.

grasmash:
Let's figure out why we're getting a null string. I suspect that output redirection may fix this. E.g., `2>&1`.

grasmash:
Closing in favor of https://github.com/acquia/blt/pull/2680.

Issue #1568, opened 2017-06-01
Use deployment identifier to prevent problems with ApcClassLoader cache


danepowell:
Drupal core uses the [Symfony ApcClassCloader](https://github.com/symfony/class-loader/blob/master/ApcClassLoader.php) (when APCu cache is available) to cache class file locations. This can create fatal errors when classes are moved or deleted, as when updating contributed modules.

For instance, I recently upgraded Config Filter from 1.0-beta3 to 1.0-rc1 by simply running `composer update drupal/config_filter` on my local install. Turns out Config Filter [removed a class file](http://cgit.drupalcode.org/config_filter/commit/?h=8.x-1.x&id=1149cfcb37d3b51311521b1451cc74dc212b0167) in this update, and the file location was still cached in my DrupalVM APCu cache, which caused the entire site to be unstable until I restarted the VM. Subsequently, the problem also occurred after deploying the update to Acquia Cloud, which required manually restarting PHP-FPM.

Fortunately, Drupal uses the deployment identifier to seed the APCu cache. Simply changing the deployment identifier when updating modules will prevent this problem. Accordingly, I propose the following changes to BLT:

1. For local installs, BLT should parse the file hash from `composer.lock` and set this as the deployment identifier. Currently, the deployment identifier isn't set at all locally.
2. For deployments, BLT should set the deployment identifier to a random value or hash on every commit. Currently, BLT only updates the deployment identifier when creating a tag.

@grasmash I'm happy to take a crack at a PR if you think this is a valid and worthwhile solution.

grasmash:
> For local installs, BLT should parse the file hash from composer.lock and set this as the deployment identifier. 

Is it feasible that the file has will be the same if no composer dependencies have changes? E.g., what if a class is moved in a custom module. I assume that will not cause a composer hash change.

danepowell:
If you moved a class in a custom module, then other local installs on your team would be liable to hit the APC error when they updated their code (although only if they updated while their VM was running). It's not ideal, but I'm not sure of a better way to track any arbitrary change to the codebase.

grasmash:
@danepowell Why not simply use the commit id as the deployment identifier?

danepowell:
I'm not sure how you could use the commit id. I hope I'm not making this more complicated than it needs to be, but this is actually a quite complex issue. Thinking both about remote deployments and local installs separately:

**For deployments**
You'd presumably want to write the deployment_identifier to disk as part of the deploy.

[...]

PR #2726, opened 2018-04-11
Fixes #2720: Write deployment identifier using robo task.

### LINKED ISSUE #2720: Error on deploy builds: unterminated quoted string
My system information:
* BLT version: 9.0.5

It looks like the deployment identifier can have quotes in it that aren't escaped properly before writing to disk. I got this error recently on a Pipelines build:

```
[ExecStack] Running echo 'MMv[Nwm'' > deployment_identifier in /tmp/artifact
sh: 1: Syntax error: Unterminated quoted string```

### PR DESCRIPTION
**Pull request by malikkotob**

### CODE REVIEW

File: src/Robo/Commands/Deploy/DeployCommand.php [hunk 1]
```diff
@@ -401,11 +401,19 @@ protected function composerInstall() {
    * Creates deployment_identifier file.
    */
   protected function createDeployId($id) {
-    $this->taskExecStack()->exec("echo '$id' > deployment_identifier")
-      ->dir($this->deployDir)
-      ->stopOnFail()
+    $deployment_identifier_file = $this->getConfigValue('repo.root') . '/deployment_identifier';
```
grasmash: Let's move this logic to a method that is shared so that we don't duplicate code.

src/Robo/Commands/Deploy/DeployCommand.php
```diff
@@ -401,11 +401,19 @@ protected function composerInstall() {
...
+      ->line($id)
       ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
       ->run();
+
+    if (!$result->wasSuccessful()) {
+      $filepath = $this->getInspector()->getFs()->makePathRelative($deployment_identifier_file, $this->getConfigValue('repo.root'));
+      throw new BltException("Unable to write deployment identifier to $filepath.");
+    }
+
+    return $result->getExitCode();
```
grasmash: We don't need to return anything. Unfortunately, this can cause Robo to actually print the exit code to screen.

src/Robo/Commands/Deploy/DeployCommand.php
malikkotob: Where would that live, ideally? Does exposing the `createDeployId` method as a command (maybe `source:build:deployment-identifier`) make sense? Then we could add it to the list of commands invoked in `artifact:build`?

src/Robo/Commands/Deploy/DeployCommand.php
malikkotob: Ack

src/Robo/Commands/Deploy/DeployCommand.php
grasmash: That sounds like a good idea!

malikkotob: Hmm, `createDeployId` actually seems like it's "on the same level", as `drupal:hash-salt:init`, in a sense. I wonder if it makes more sense to move it into the `drupal` namespace under `drupal:deployment_identifier:init`. The `createDeployId` method is currently invoked from the `drupal:install` command, so it seems like there's definitely an argument that it can fit into the `drupal` namespace, since BLT needs it whenever it installs `drupal`. We could remove it from `drupal:install` and instead add it to the list of commands invoked in `blt setup`:
```
    $commands = [
      'source:build',
      'drupal:hash-salt:init',
      'drupal:deployment-identifier:init`,
    ];
```

and, again, to the list of commands invoked in `artifact:build`:
```
    $this->invokeCommands([
      // Execute `blt source:build:frontend` to ensure that frontend artifact
      // are generated in source repo.
      'source:build:frontend',
      // Execute `drupal:hash-salt:init` to ensure that salt.txt exists.
      // There's a slim chance this has never been generated.
      'drupal:hash-salt:init',
      'drupal:deployment-identifier:init`,
    ]);
```

Thoughts?

src/Robo/Commands/Setup/AllCommand.php
```diff
@@ -25,6 +26,9 @@ public function setup() {
       'source:build',
       'drupal:hash-salt:init',
     ];
+    $commands['drupal:deployment-identifier:init'] = [
+      'id' => RandomString::string(8, FALSE, NULL, 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890!#%^&*()_?/.,+=><'),
```
grasmash: It looks like this command will create a random ID for you if you don't specify one. Why create a random one here?

src/Robo/Commands/Setup/SettingsCommand.php
```diff
@@ -286,4 +286,29 @@ public function hashSalt() {
...
+  /**
+   * Writes a deployment_identifier to ${repo.root}/deployment_identifier.
+   *
+   * @command drupal:deployment-identifier:init
+   * @aliases ddii
+   *
+   * @throws \Acquia\Blt\Robo\Exceptions\BltException
+   */
+  public function createDeployId($id = '') {
+    if (!$id) {
```
grasmash: Arguments should be required. If you want an option, use an $options array. See https://github.com/consolidation/annotated-command#option-default-values

tests/phpunit/src/BltProjectTestBase.php
```diff
@@ -208,6 +208,7 @@ protected function importDbFromFixture($root = NULL, $uri = 'default') {
     $drush_bin = $this->sandboxInstance . '/vendor/bin/drush';
     $this->execute("$drush_bin sql-drop --root=$root --uri=$uri", NULL, FALSE);
     $this->blt('drupal:hash-salt:init');
+    $this->blt('drupal:deployment-identifier:init');
```
grasmash: Why do we do this here? Is it necessary outside of a deployment? If so, should we generate one for all users after initial setup, rather that only in the test case?

src/Robo/Commands/Setup/AllCommand.php
malikkotob: @grasmash that's what we're currently doing as part of the [`drupal:install` command](https://github.com/acquia/blt/blob/ceb6e8e3becf5338a7709d2d528a0fb7128729cc/src/Robo/Commands/Setup/BuildCommand.php#L40), so I just followed suit. Do you know why we were doing that here, previously?

tests/phpunit/src/BltProjectTestBase.php
malikkotob: I believe it is, necessary. @danepowell, can you confirm? As mentioned in an earlier comment, we were previously creating this as part of the [`drupal:install` command](https://github.com/acquia/blt/blob/ceb6e8e3becf5338a7709d2d528a0fb7128729cc/src/Robo/Commands/Setup/BuildCommand.php#L40) (in addition to as part of putting together the build artifact).

Also, this PR does generate one for all users as part of setup. It stops invoking it from the `drupal:install` command and adds it as a command to invoke [as part of `setup:all`](https://github.com/acquia/blt/pull/2726/files#diff-2214b274c5704bd37ecb1d740ac17a20R29).

src/Robo/Commands/Setup/AllCommand.php
grasmash: Because a Drupal account password is more restrictive it term of the characters that can be used.

malikkotob: Hmm, I'm not sure what a Drupal account password has to do with this deployment identifier. Am I missing something?

src/Robo/Commands/Setup/SettingsCommand.php
malikkotob: Ack

src/Robo/Commands/Setup/AllCommand.php
malikkotob: If deployment identifiers don't need to have character restrictions I'll just kill the `id` option altogether. Thoughts?

grasmash:
Last thing, can we add relevant tests to assert that the `deployment_identifier` file exists? I'd guess that we need a test in DeployTest and SetupCommandTest, if both of those commands are expected to generate it. We should also assert that it is non-empty.

tests/phpunit/src/BltProjectTestBase.php
danepowell: The deployment identifier is most important for deployments (duh), but it's also useful for preventing unexpected behavior in local environments due to stale twig/apc caches. See discussion here: https://github.com/acquia/blt/issues/1568

malikkotob:
Added, @grasmash. Anything else I should add on or change here?

PR #2732, opened 2018-04-12
Throw exception when cloud hooks are used for ACSF.

grasmash:


grasmash:
This is not ready to merge. The conditional statement needs to be revised by @lcatlett.

grasmash:
Duplicate of https://github.com/acquia/blt/pull/2768

PR #2735, opened 2018-04-17
Updating BLT README template to include more basic info.

mikemadison13:
Changes proposed:
- Updates the README with more useful information about basic BLT workflow (and Git workflow)

### DIFF
--- a/template/README.md
+++ b/template/README.md
@@ -2,19 +2,90 @@
 
 A brief description of My Project.
 
+## Using This Template
+
+Remove this section after initial setup!
+
+Search for and replace the following placeholders:
+
+* \<organization>
+* \<project>
+* \<project local url>
+* \<local environment>
+* \<project key>
+* \<branch>
+* \<provide link>
+
 ## Getting Started
 
-This project is based on BLT, an open-source project template and tool that enables building, testing, and deploying Drupal installations following Acquia Professional Services best practices.
+This project is based on BLT, an open-source project template and tool that enables building, testing, and deploying Drupal installations following Acquia Professional Services best practices. While this is one of many methodologies, it is our recommended methodology. 
 
-To set up your local environment and begin developing for this project, refer to the [BLT onboarding documentation](http://blt.readthedocs.io/en/latest/readme/onboarding/). Note the following properties of this project:
-* Primary development branch: develop
-* Local environment: DrupalVM
-* Local drush alias: @my-project.local
-* Local site URL: http://local.my-project.com
+* Review the [Required / Recommended Skills](http://blt.readthedocs.io/en/latest/readme/skills) for working with a BLT project
+* Ensure that your computer meets the minimum installation requirements (and then install the required applications). See the [System Requirements](http://blt.readthedocs.io/en/latest/INSTALL/#system-requirements).
+* Request access to organization that owns the project repo in GitHub (if needed)
+* Fork the project repository in GitHub
+* Request access to the Acquia Cloud Environment for your project (if needed)
+* Setup a SSH key that can be used for GitHub and the Acquia Cloud (you CAN use the same key)
+    * [Setup GitHub SSH Keys](https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/)
+    * [Setup Acquia Cloud SSH Keys](https://docs.acquia.com/acquia-cloud/ssh/generate)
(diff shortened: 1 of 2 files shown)


template/README.md
```diff
@@ -2,15 +2,110 @@
...
 ## Getting Started
 
 This project is based on BLT, an open-source project template and tool that enables building, testing, and deploying Drupal installations following Acquia Professional Services best practices.
 
-To set up your local environment and begin developing for this project, refer to the [BLT onboarding documentation](http://blt.readthedocs.io/en/latest/readme/onboarding/). Note the following properties of this project:
-* Primary development branch: develop
-* Local environment: DrupalVM
-* Local drush alias: @my-project.local
-* Local site URL: http://local.my-project.com
+* Ensure that your computer meets the minimum installation requirements (and then install the required applications). See the [System Requirements](http://blt.readthedocs.io/en/8.x/INSTALL/).
```
grasmash: This should be 9.x.

template/README.md
```diff
@@ -2,19 +2,124 @@
...
+* Fork the project repository in GitHub
+* Request access to the Acquia Cloud Environment for your project (if needed)
+* Setup a SSH key that can be used for GitHub and the Acquia Cloud (you CAN use the same key)
+    * [Setup GitHub SSH Keys](https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/)
+    * [Setup Acquia Cloud SSH Keys](https://docs.acquia.com/acquia-cloud/ssh/generate)
+* Clone your fork
+```
+$ git clone git@github.com:<account>/<project>.git
+```
+* Add the parent repository as an upstream
```
grasmash: Wouldn't this already be done by the `git clone`?

template/README.md
```diff
@@ -2,19 +2,124 @@
...
+```
+$ git remote add upstream git@github.com:<organization>/<project>.git
+```
+* Install Composer Dependencies (warning: this can take some time based on internet speeds)
+```
+$ composer install
+```
+* Setup Local Environment 
+
+Note: Drupal VM is OPTIONAL but templated within BLT. If using Drupal VM, BLT 9 assumes all commands are run INSIDE the VM (this is a change from BLT 8).
```
grasmash: I would link to http://blt.readthedocs.io/en/9.x/readme/local-development/#using-drupal-vm-for-blt-generated-projects, which lists system level requirements for using Drupal VM.

template/README.md
```diff
@@ -2,19 +2,124 @@
...
+
+* Run the initial Setup
+```
+$ vagrant ssh
+$ blt setup
+``` 
+* Access the site and do necessary work at \<project local url> by running
+
+```
+$ drush @<project>.local uli
```
grasmash: We're trying to move away from using drush aliases for the VM. This also assumes that drush is globally installed.

template/README.md
```diff
@@ -2,19 +2,124 @@
...
+
+Additional [BLT documentation](http://blt.readthedocs.io) may be useful. You may also access a list of BLT commands by running:
+```
+$ blt
+``` 
+
+Note the following properties of this project:
+* Primary development branch: \<branch>
+* Local environment: \<local environment>
+* Local drush alias: @\<project>.local
```
grasmash: Again, I wouldn't use a local drush alias.

template/README.md
```diff
@@ -2,19 +2,124 @@
...
+* Local site URL: \<project local url>
+
+## Working With a BLT Project
```
grasmash: This doesn't feel like it belongs in a read me. Isn't this type of information in the onboarding doc?

mikemadison13: @grasmash this level of specificity is NOT in the onboarding doc, but i'm happy to link to that in this section instead and move this detail over. i think it needs to be in our docs "somewhere" but could be convinced that "here" isn't the right place.

template/README.md
mikemadison13: @grasmash will update these. although, you still need... @self to run in the vm, right?

template/README.md
mikemadison13: good call, we can then also include other locals in the future as/if we begin supporting them.

template/README.md
mikemadison13: will make this clearer in regards to origin == fork and upstream == original repository

mikemadison13:
@grasmash  this should be ready for another review, i addressed all previous comments

PR #2749, opened 2018-04-18
Fixes #2747 to define target_env for BLT 8.x

### LINKED ISSUE #2747: Can't install new sites on ACSF
My system information:
* BLT version: 9.0.5

Currently, if you install a new site on ACSF (via the ACSF management console), BLT attempts to do its normal post-install UUID-setting / config-importing goodness but fails because `$target_env` is not defined here:
https://github.com/acquia/blt/blob/bc97176ea060d85bb942b896c10d19d83a8bc4f3/settings/acsf/post-install/post-install.php#L23

mikemadison13:
Fixes #2747 .

Changes proposed:
- defines $target_env variable based on existing $site and $env
- version 8.9.x.

### DIFF
--- a/settings/acsf/post-install/post-install.php
+++ b/settings/acsf/post-install/post-install.php
@@ -15,6 +15,7 @@
 
 $site = $_ENV['AH_SITE_GROUP'];
 $env = $_ENV['AH_SITE_ENVIRONMENT'];
+$target_env = $site . $env;
 
 // The public domain name of the website.
 // Run updates against requested domain rather than acsf primary domain.


settings/acsf/post-install/post-install.php
```diff
@@ -15,6 +15,7 @@
 
 $site = $_ENV['AH_SITE_GROUP'];
 $env = $_ENV['AH_SITE_ENVIRONMENT'];
+$target_env = $site.$env;
```
grasmash: Given that there is no test coverage for this file, can we verify that this works?

PR #2769, opened 2018-04-24
[WIP]Fixes #2663 to add chromium as a supported headless browser.

### LINKED ISSUE #2663: Add Support for Chromium for Behat Testing
Currently, we're forcing users to use a proprietary `google-chrome` binary. On Linux, we can have on par functionality by using `chromium` instead.

Here's me hacking things around to demonstrate it works well:

```
$ ls -l /usr/local/bin/google-chrome 
lrwxrwxrwx 1 root staff 17 Mar 21 18:07 /usr/local/bin/google-chrome -> /usr/bin/chromium
```

In `./vendor/acquia/blt/src/Robo/Commands/Tests/BehatCommand.php` I commented out the `checkChromeVersion()` method entirely and also commented out `$this->checkChromeVersion($chrome_bin);` in the `launchChrome()` method to not get the version check to kick in, since Chromium's version doesn't look the same as Chrome's.

```
$ chromium --version
Chromium 64.0.3282.119 built on Debian 9.3, running on Debian 9.4
```

And a test run:

```
$ blt tests:behat:run
[Robo\Common\ProcessExecutor] Running 'google-chrome' --headless --disable-web-security --remote-debugging-port=9222  http://localhost in /var/www/html/blt
[Testing\Behat] Running behat  --format pretty /var/www/html/blt/tests/behat --colors --no-interaction --stop-on-failure --strict --config /var/www/html/blt/tests/behat/local.yml --profile local --tags '~ajax&&~experimental&&~lightningextension' -v
@example
Feature: Web drivers
  In order to verify that web drivers are working
  As a user
  I should be able to load the homepage
  With and without Javascript

  @javascript
  Scenario: Load a page with Javascript # features/Examples.feature:9
    Given I am on "/"                   # Drupal\DrupalExtension\Context\MinkContext::visit()
    Then I should be on "/"             # Drupal\DrupalExtension\Context\MinkContext::assertPageAddress()

  Scenario: Load a page without Javascript      # features/Examples.feature:13
    Given I am on "/"                           # Drupal\DrupalExtension\Context\MinkContext::visit()
    Then the response status code should be 200 # Drupal\DrupalExtension\Context\MinkContext::assertResponseStatus()

2 scenarios (2 passed)
4 steps (4 passed)
0m0.46s (20.97Mb)
```

So ideally we would:

* Check for either the `google-chrome` or `chromium` binary
* Check the `chromium` version
* In `tests/behat/local.yml`, make sure the keys currently accepting `chrome` as a value can also accept `chromium` in the future
* Make the code a little less opinionated about Google Chrome

mikemadison13:
Fixes #2663 .

Changes proposed:
- adds chromium ports to config/build.yml
- replicates all chrome functions as chromium

### DIFF
--- a/src/Robo/Commands/Tests/BehatCommand.php
+++ b/src/Robo/Commands/Tests/BehatCommand.php
@@ -59,6 +59,16 @@ class BehatCommand extends TestsCommandBase {
   protected $chromeArgs;
 
   /**
+   * @var string
+   */
+  protected $chromiumPort;
+
+  /**
+   * @var string
+   */
+  protected $chromiumArgs;
+
+  /**
    * This hook will fire for all commands in this command file.
    *
    * @hook init
@@ -68,6 +78,8 @@ class BehatCommand extends TestsCommandBase {
     $this->behatLogDir = $this->getConfigValue('reports.localDir') . "/behat";
     $this->chromePort = $this->getConfigValue('behat.chrome.port');
     $this->chromeArgs = $this->getConfigValue('behat.chrome.args');
+    $this->chromiumPort = $this->getConfigValue('behat.chromium.port');
+    $this->chromiumArgs = $this->getConfigValue('behat.chromium.args');
     $this->seleniumPort = $this->getConfigValue('behat.selenium.port');
     $this->seleniumUrl = $this->getConfigValue('behat.selenium.url');
     $this->serverPort = $this->getConfigValue('tests.server.port');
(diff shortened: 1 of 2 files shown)


src/Robo/Commands/Tests/BehatCommand.php
```diff
@@ -162,6 +174,72 @@ protected function launchWebDriver() {
...
+   *
+   * @param string $bin
+   *   Absolute file path to the chromium bin.
+   *
+   * @throws \Acquia\Blt\Robo\Exceptions\BltException
+   *   Throws exception if minimum version is not met.
+   */
+  protected function checkChromiumVersion($bin) {
+    $version = (int) $this->getContainer()->get('executor')
+      ->execute("'$bin' --version | cut -f3 -d' ' | cut -f1 -d'.'")
```
anavarre: I don't think this one will work.

Just tried locally:

```
$ chromium --version | awk '{print $2}' | cut -f1 -d '.'
64
```

Alternatively, to follow your approach:

```
$ chromium --version | cut -f2 -d ' ' | cut -f1 -d '.'
64
```

mikemadison13:
@anavarre i've adjusted the function per your recommendation

anavarre:
Patch from the PR applied successfully:

```
$ patch -p1 < d67974dda70246097d7fe6d61566ea5135596944.patch 
patching file config/build.yml
patching file src/Robo/Commands/Tests/BehatCommand.php
```

However, while the code changes are here, I seem to still get blocked on the missing `google-chrome` binary.

```
$ grep '\$chromium_bin' vendor/acquia/blt/src/Robo/Commands/Tests/BehatCommand.php
    $chromium_bin = $this->findChromium();
    $this->checkChromiumVersion($chromium_bin);
      ->execute("'$chromium_bin' --headless --disable-web-security --remote-debugging-port={$this->chromiumPort} {$this->chromiumArgs} $chromium_host")
$ blt behat
[error]  Could not find Google Chrome. Please add an alias for "google-chrome" to your CLI environment.
```

mikemadison13:
@anavarre did you update your behat configuration to use chromium instead of behat?

i'm not 100% sure where https://github.com/acquia/blt/pull/2769/files#diff-a240de0034a3c5a228b0531256268671R177 is checking for this from... ```behat.web-driver``` possibly goes in the blt/blt.yml file?

anavarre:
So, that's the thing. I was expecting for an OOTB `chromium` detection (if present) but the code is still very much opinionated for `google-chrome`. Should this be taken into account as part of the BLT `tests:behat:init:config` command?

Under `config/build.yml` I found how to make this work.

```
# May be chrome, selenium, phantomjs.
  web-driver: chromium
```

(BTW this should be `# May be chrome, chromium, selenium, phantomjs.`)

Editing `tests/behat/local.yml` (which is generated as part of the `init` command) doesn't seem to override `config/build.yml`'s configuration above.

mikemadison13:
@grasmash it sounds like he had to manually change a file in BLT to make chromium work... any thoughts on how to do this in configuration? I would have thought that changing the blt/blt.yml file would work.

src/Robo/Commands/Tests/BehatCommand.php
```diff
@@ -162,6 +174,72 @@ protected function launchWebDriver() {
     elseif ($this->getConfigValue('behat.web-driver') == 'chrome') {
       $this->launchChrome();
     }
+    elseif ($this->getConfigValue('behat.web-driver') == 'chromium') {
+      $this->launchChromium();
+    }
+  }
+
+  /**
+   * Launches a headless chromium process.
```
anavarre: Chromium

src/Robo/Commands/Tests/BehatCommand.php
```diff
@@ -162,6 +174,72 @@ protected function launchWebDriver() {
...
+      ->execute("'$chromium_bin' --headless --disable-web-security --remote-debugging-port={$this->chromiumPort} {$this->chromiumArgs} $chromium_host")
+      ->background(TRUE)
+      ->printOutput(TRUE)
+      ->printMetadata(TRUE)
+      ->run();
+    $this->getContainer()->get('executor')->waitForUrlAvailable("$chromium_host:{$this->chromiumPort}");
+  }
+
+  /**
+   * Kills headless chromium process running on $this->chromiumPort.
```
anavarre: Chromium

src/Robo/Commands/Tests/BehatCommand.php
```diff
@@ -162,6 +174,72 @@ protected function launchWebDriver() {
...
+  }
+
+  /**
+   * Finds the local Chromium binary.
+   *
+   * @return null|string
+   *   NULL if Chromium could not be found.
+   *
+   * @throws \Acquia\Blt\Robo\Exceptions\BltException
+   *   Throws exception if google-chrome cannot be found.
```
anavarre: s/google-chrome/Chromium

src/Robo/Commands/Tests/BehatCommand.php
```diff
@@ -162,6 +174,72 @@ protected function launchWebDriver() {
...
+      return 'chromium';
+    }
+    throw new BltException("Could not find Chromium. Please add an alias for \"chromium\" to your CLI environment.");
+  }
+
+  /**
+   * Verifies that Chromium meets minimum version requirement.
+   *
+   * @param string $bin
+   *   Absolute file path to the chromium bin.
```
anavarre: Chromium binary

grasmash:
Please add a test for this new functionality.

PR #2775, opened 2018-04-27
Add merged composer deploy scripts

christopher-hopper:
## Motivation

We've been maintaining `composer deploy` scripts in our BLTed projects and I thought it may be worth sharing with everyone. They're a quick way to call `blt deploy` with sensible defaults for the commit message and a branch or tag name. 

Not sure if this is the right place, but sharing is caring! right? or not. You tell me. I thought at least here they'll get merged into the root project composer.json.

## Changes proposed

- Add `composer deploy:branch` script 
- Add `composer deploy:tag` script 

## Usage examples

For a branch, you do something like this

```
git checkout master
composer install --no-interaction
composer deploy:branch
```

For a tag, you do something like this

```
git checkout master
git tag -a v1.0.0 -m "Release version 1.0.0"
composer install --no-interaction
composer deploy:tag
```

### DIFF
--- a/composer.required.json
+++ b/composer.required.json
@@ -50,6 +50,8 @@
   },
   "scripts": {
     "blt-alias": "blt blt:init:shell-alias -y --ansi",
+    "deploy:branch": "blt deploy --commit-msg=\"Automated build by $(git config user.name) for commit $(git rev-parse HEAD)\" --branch=\"$(git rev-parse --abbrev-ref HEAD)-build\" --ignore-dirty --no-interaction --verbose",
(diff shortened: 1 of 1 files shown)


grasmash:
It seems like these should be BLT commands rather than Composer scripts. Composer scripts are typically used for scenarios when we cannot be certain that BLT is available for execution.

Also, we will need to add tests for any new functionality.

christopher-hopper:
Sure, sounds sensible. 

Does anyone have a view on whether these would be welcome additions as BLT commands? If so I could take a look at adding them somewhere. I'd need some guidance on the command names to use. 

The goal of these is to have a command, that takes no arguments or options, but can deploy code as either a branch or a tag. If you need options or arguments, the existing BLT deploy command provides that. I wouldn't want to replace the existing BLT deploy command.

christopher-hopper:
I'm going to close this and create a new PR that contains these as a blt/robo command instead.

PR #2780, opened 2018-04-27
Allowing flexibility with PHPUnit version.

### LINKED ISSUE #2771: Request: Update PHPUnit library

dpagini:
Fixes #2771

Honestly, I haven't really tested this much, just wanted to put something up... maybe see how this reacts to existing tests? I'm not even sure the best way to test this right now? But this is what I was thinking we need to do...


dpagini:
K, so here's an approach... I had to push up some fixes based on the test results, but this looks promising. There is a shim file here to allow PHPUnit 4 through 6 to run out of the same project. I found the suggestion [here](https://stackoverflow.com/questions/6065730/why-fatal-error-class-phpunit-framework-testcase-not-found-in).

I took a guess at the best place for this "shim file" but if you want another approach, happy to make any changes...

grasmash:
I think I'd prefer to just bump the version and release a BLT 10 version around the time that Drupal 8.6 is released.

dpagini:
I should also point out this would be a problem for anyone currently running PHP 7.2. Maybe I'm the only one...  =)
Also, if you do go away from this approach (which I get it, it's a little hacky...) then which PHPUnit version would you envision bumping to? PHPUnit 5 is EOL, and PHPUnit 6 does not support PHP < 7.0 anymore. We're back at the PHP version support discussion. (Sorry, haha)

grasmash:
Drupal 8.6 will support PHP 5.6, so we should support PHP 5.6 until Drupal 8.7 is released.

I think that means we need to support PHPUnit 5, but we can drop support for PHPUnit 4.

composer.json
```diff
@@ -31,7 +31,7 @@
         "grasmash/yaml-cli": "^1.0.0",
         "grasmash/yaml-expander": "^1.2.0",
         "oomphinc/composer-installers-extender": "^1.1",
-        "phpunit/phpunit": "^4.8",
+        "phpunit/phpunit": "^4.8|^5.7|^6.5",
```
grasmash: Let's drop 4.* support.

scripts/phpunit/bootstrap.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+// Compatibility for PHPUnit < 6 and PHPUnit 6+.
+if (class_exists('PHPUnit_Framework_TestSuite') === true && class_exists('PHPUnit\Framework\TestSuite') === false) {
```
grasmash: Is `===` necessary? I think `class_exists()` always returns a boolean. You should be fine with:
```
if (class_exists('PHPUnit_Framework_TestSuite') 
  && !class_exists('PHPUnit\Framework\TestSuite')) {
```

grasmash: I'd like a comment here describing what this condition indicates. Does this mean it's PHPUnit 4, for instance?

tests/phpunit/bootstrap.php
```diff
@@ -4,6 +4,9 @@
  * @file
  */
 
+// Alias PHPUnit classes for backwards compatibility.
+include '../../scripts/phpunit/bootstrap.php';
```
grasmash: Is this necessary? Do internal tests break without it?

dpagini: Hmm... I _think_ this is needed for the PHP 5.6 tests. When PHP is 5.6, composer will only install PHPUnit 5, which does not have `PHPUnit\Framework\TestCase` as a valid class. This file aliases those namespaces to the old classes.

scripts/phpunit/bootstrap.php
dpagini: > Is === necessary? I think class_exists() always returns a boolean.

Makes sense to me... I think I copied the code from a comment. I'll update.

Are you saying you'd like more detail in the comment on line 3? Based on your other comment, if we drop PHPUnit 4 support, then this file is a shim to make PHPUnit 5 work with PHPUnit 6 namespaced classes. You're just asking for that statement in a more verbose comment?

tests/phpunit/bootstrap.php
grasmash: Got it.

dpagini: Actually, Drupal core is [doing something _similar_](https://github.com/drupal/drupal/blob/8.6.x/core/tests/bootstrap.php#L197) that I just came across, because I was confused how they were going to support both versions. I guess it's just a weird in-between time with this PHP support decision.

dpagini:
Based on the feedback above... are you still looking to get this merged in for a later BLT major release? I only ask b/c, if you don't mind, I'd like to leave this open and _as-is_ in case anyone is, like me, trying to run BLT with PHP 7.2. That will give them a [patch](https://github.com/acquia/blt/pull/2780.patch) to pull in before BLT 10 comes out, and I could open up a new PR with all the requested changes...?

_(Actually, as I say that, I'm not sure this will work as a patch b/c of the composer.json change...)_

composer.json
danepowell: Agreed, it's crazy deprecated, I don't see any reason to keep it, and I've verified that there shouldn't be any conflicts with other dependencies if 5.x is supported.

danepowell: @ba66e77 and I have agreed that we should maintain support for 4.x until we release a new version of BLT. I'll review this PR today.

danepowell:
We're trying to get this merged into 9.x (without dropping support for 4.x). I addressed some of the feedback and opened #2982

PR #2782, opened 2018-04-27
Adding a postDbCopy cloud hook command.

### LINKED ISSUE #2781: BLT 9 Cloud Hooks changes
My system information:
* BLT version: 9.0.5

dpagini:
Fixes #2781 

First pass at 2781. Adding a hook to run when the database is copied down to run deployed updates.


dpagini:
There was a little discussion on 2781 on this... I wonder if another option here would be to accept the PHP code into BLT, and we can just make the use of it optional (via not including a default .sh script). Thoughts?

dpagini:
Can someone with permission restart this build? These failures look suspect... cc: @mikemadison13

mikemadison13:
restarted @dpagini

dpagini:
Any thoughts on this approach... good/bad?

trwill:
Looks good to me!

rjgwiz:
@dpagini any ideas on how to make `post-db-copy` for ACE sites only update the database that was copied? Since only one db at a time can be copied, updates should only run for the corresponding site, not all multisites

The current chain for ACE is: 
- `postDbCopy()`
- `$this->updateSites($site, $target_env)`
- `$this->updateAceSites($target_env)`
- `$this->invokeCommand('artifact:update:drupal:all-sites');`

What's the cleanest way to have `postDbCopy()` take `$db_name` and find which multisite it belongs to? 

Since the `update*` functions are used by other cloud hooks (which aren't specific to 1 site/database), does it make sense to add optional params, for example:
- `updateSites($site, $target_env, $multisite_name = NULL)`
- `updateAceSites($target_env, $multisite_name = NULL)` 

To help determine whether `updateAceSites()` should call:
- `invokeCommand('artifact:update:drupal:all-sites')`
or
- `invokeCommand('artifact:update:drupal', ["--site=$multisite_name"])`
?

dpagini:
I don't really have any suggestion on this b/c BLT doesn't know which database name belongs to which multisite... they aren't really connected in any way w/o maybe evaluating a drush sql statement maybe. Does this command need to be multisite aware to go in? Isn't this a different issue?

rjgwiz:
> Does this command need to be multisite aware to go in? Isn't this a different issue?

The only concern I would have is `artifact:update:drupal:all-sites` is being run on every ACE db copy. If someone has a lot of multisites configured, running update on all sites will take longer (imagine 10+ multisites). And if multiple dbs are copied within a say a minute of each other, can two `artifact:update:drupal:all-sites` run concurrently without stepping on each others toes? I could run some tests on this today

I don't think there's any good way to dynamically lookup `site` from `db_name`. One good thing is hosted environments typically have static db names so we could map it in config. I could see this living is in `blt.yml`... somewhere. I thought about in a drush site file under `options` but that seems too disconnected. blt.yml could map `$db_name` to a multisite. `postDbCopy()` could look for this config pair and if it exists, call `updateSite()`, otherwise call `updateSites()` to default to all sites.

danepowell:
I have strong feelings both for and against the idea of post-db-copy hooks.

On the one hand, when I "stage" down environments from prod, I prefer to deploy the code first and then copy down the db / files. Currently this operation is very fast and I am also guaranteed to get a bit-level clone of the production environment (since nothing happens post-db-copy). Adding post-db-copy commands would make both of those aspects marginally worse.

On the other hand, I can't tell you how many times I've copied a db somewhere and then wondered why configuration that's environment-dependent (such as Shield / http auth) didn't get imported. Adding post-db-copy would fix that.

I guess on the whole this is probably a good thing to have, since it will provide some consistency and prevent confusion like in the Shield use case. I just wonder if it will make staging deployments harder.

dpagini:
Yes, I've had the same pros/cons list in my head on this one too. I'm sure there are other ways to do it, but we utilize things like environment config or toggle-modules to make sure our non-prod environments are protected (from sending live emails, for example), so once you copy down the database, if you don't do those updates, there is some risk your nonprod server acts like prod for some undetermined amount of time. It would be awesome if, upon copying the database down in the UI, Acquia could provide a checkbox to just not run the hooks, giving you the option to have an untouched version of production databases.

Anyways, just thinking of some alternate ideas... It seems to make sense to me to provide the hook, even if it were to be empty. If we fire the robo hook, we can then hook in other robo commands, and maybe this should turn into a recipe that users can run if they want this functionality, instead of defaulting it for everyone...? I haven't looked much into the recipe stuff, other than knowing it exists, but it actually came up for me earlier today.... we have a job that runs before the code update on production and takes a backup of the database immediately after code is pushed (in case you forgot) but before any commands are run. That feels like something that would fit into the recipe model to me... Sorry to get a little rambly there.

dpagini:
Update and slight approach tweak in #3202.

PR #2787, opened 2018-04-30
Adding support for factory hooks.

grasmash:


src/Robo/Commands/Artifact/FactoryHooksCommand.php
```diff
@@ -39,18 +39,45 @@ public function updateAcsfSites($site, $target_env) {
...
-      ->run();
-    if (!$result->wasSuccessful()) {
-      throw new BltException("Unable to get list of ACSF sites.");
-    }
-    $output = $result->getMessage();
-    // @todo Populate. Update ACSF tools to drush 9.
-    $sites = [];
+    $sites = $this->getAcsfSites();
     $this->getConfig()->set('multisites', $sites);
     $this->invokeCommand('artifact:update:drupal:all-sites');
```
grasmash: @lcatlett @danepowell Do we actually want to update all multisites when a give db-update.sh hook is called, or do we want to just execute updates for the specific [`$site`]9https://github.com/acquia/blt/blob/9.x/scripts/factory-hooks/db-update/db-update.sh#L13(https://github.com/acquia/blt/blob/9.x/scripts/factory-hooks/db-update/db-update.sh#L13) defined in the db-update.sh ?

grasmash: I guess I'm asking for context around when this db-update.sh hook is executed. Is it called once per site in ACSF, or is it called one for the factory? If it's called once per site, then the code executed by the hook doesn't require awareness of any sites beyond the single one that in being updated at that time.

danepowell: I know that db-update is called when an administrator deploys code via the management console and selects the option to run db updates... however I have no idea if it's called once for the entire deploy or on a site-by-site basis... I'll try to find out.

danepowell: db-update.sh is called once per site during code deploys, in a format like this:
`/mnt/www/html/foo.01dev/factory-hooks/db-update/db-update.sh' 'foo' '01dev' 'blah123dev' 'site1.dev-foo.acsitefactory.com'`

grasmash: So I don't think we even need this `getAcsfSites()`? BLT wouldn't need a list of ACSF sites.

BLT just needs to call `artifact:update:drupal`, which in turn calls `drupal:config:import` and `drupal:toggle:modules` on the target site.

BLT only need to know the site so that it can:

1. Pass it as --uri to drush, and 
2. So it can load site-specific BLT config, which isn't really possible with ACSF anyway since the `docroot/sites` dir doesn't actually contain the ACSF site dirs, right?

danepowell: Note that although db-update only runs for individual sites at a time, we still need to the ability to run updates on _all_ sites when post-code-update is called.

The post-code-update support is necessary for dev environments that typically run a development code branch that gets automatically updated as new code is merged. Note that using post-code-update is officially unsupported on ACSF, but it _should_ be safe to use in most cases because only dev environments should be running code branches (as opposed to tags), and dev environments typically have far fewer sites than prod.

That's just my opinion though, I'm open to other input as long as we can support the standard dev workflow.

PR #2790, opened 2018-05-01
Fixes #2789: Require minimum Composer API version.

### LINKED ISSUE #2789: Minimum Composer version
Lately we've seen a few user reports of BLT project creation failing because Drupal Scaffold isn't running, so things like `docroot/index.php` aren't getting created.

The root cause seems to be older Composer versions (including 1.1.3 and 1.3.2) that apparently don't work with Drupal Scaffold: https://github.com/drupal-composer/drupal-scaffold/issues/72

We should at least document the minimum required Composer version (I'm not exactly sure what that is, but maybe just recommend the latest), and if possible do some sort of code-level check for the correct Composer version when creating projects.

grasmash:


grasmash:
Composer's docs say that's the current version of composer-plugin-api?

bhasselbeck:
Does this roll out at least 1.4.1 of composer?

> The root cause seems to be older Composer versions (including 1.1.3 and 1.3.2) that apparently doesn't work with Drupal Scaffold: drupal-composer/drupal-scaffold#72

grasmash:
I know Composer != composer-plugin-api, but I'm having trouble getting a list of composer-plugin-api versions and corresponding Composer versions.

grasmash:
https://github.com/composer/composer/blob/14c77130e8f717ceb154720e1abefafe99124cb9/src/Composer/Plugin/PluginInterface.php#L30

grasmash:
Yeah so this doesn't do the trick...

grasmash:
I'm not even sure that it's possible to require a specific version of Composer. Wish we had more detail on what was broken.

grasmash:
This won't fix the issue, but it won't hurt either.

PR #2791, opened 2018-05-01
Update RELEASE.md

danepowell:
Seems like this document needs some cleanup. The only thing I know for sure can be updated is the 8.x/9.x test branch.

I'm not sure if we still use the blted repo, it hasn't had any commits since last year so I assume not?


grasmash:
We really should still use it, but I'm not sure what its status is.

PR #2795, opened 2018-05-03
Issue #2794: Fix db update failures on ACSF.

### LINKED ISSUE #2794: DB updates fail on ACSF, breaking deploys
My system information:
* Operating system type: 9.x-dev

When I deploy my BLT-based project to ACSF via the management console and select the option to apply db updates as part of the deploy, the deployments always fail. The task log for each SiteUpdate task contains an error like this:
> The db-update.sh script has encountered an error... Running updates for site foo in environment 01dev.> artifact:update:drupal... The specified database connection is not defined: default

Clearly BLT _should_ be targeting the specific site specific by the `$domain` provided in the db-update hook, not `default`.

I think what's happening here is that this call to switchSiteContext() is clobbering the domain passed via the CLI: https://github.com/acquia/blt/blob/9.x/src/Robo/Commands/Artifact/FactoryHooksCommand.php#L41

danepowell:
Fixes #2794 

This feels like a ham-fisted approach, but it does seem to fix the problem. Note that `artifact:acsf-hooks:db-update` almost immediately invokes `artfifact:update:drupal`, which also invokes `switchSiteContext()`. It seems wrong to call that twice, I don't if that could have been part of the problem here.

### DIFF
--- a/src/Robo/Commands/Artifact/FactoryHooksCommand.php
+++ b/src/Robo/Commands/Artifact/FactoryHooksCommand.php
@@ -38,7 +38,6 @@ class FactoryHooksCommand extends BltTasks {
       ->drush("cc drush")
       ->run();
     $this->say("Running updates for site <comment>$site</comment> in environment <comment>$target_env</comment>.");
-    $this->switchSiteContext($site);
     $this->invokeCommand('artifact:update:drupal');
   }
 
--- a/src/Robo/Commands/Artifact/FactoryHooksCommand.php
+++ b/src/Robo/Commands/Artifact/FactoryHooksCommand.php
@@ -10,34 +10,42 @@ use Acquia\Blt\Robo\BltTasks;
 class FactoryHooksCommand extends BltTasks {
 
   /**
-   * Execute updates against an artifact hosted in AC Cloud.
+   * Execute updates against an artifact hosted in ACSF.
    *
-   * This is intended to be called from post-code-deploy.sh cloud hook.
+   * This is intended to be called via the db-update factory hook whenever
+   * code is deployed from the management console and the user has opted to
+   * apply DB updates.
+   *
+   * Note that ACSF calls this hook once per site being updated.
    *
    * @param string $site
-   *   The site name. E.g., site1.
+   *   The "ACSF site" name (actually subscription name). E.g., foo.
    * @param string $target_env
-   *   The cloud env. E.g., dev
+   *   The cloud env. E.g., 01dev
+   * @param string $db_role
+   *   Internal to ACSF.
+   * @param string $domain
+   *   Domain for the site being updated. E.g., foo1.foo.acsitefactory.com.
    *
    * @command artifact:acsf-hooks:db-update
    */
   public function dbUpdate($site, $target_env, $db_role, $domain) {
-    $this->updateAcsfSites($site, $target_env);
+    $this->updateAcsfSites($domain, $target_env);
   }
 
   /**
-   * Executes updates against all ACSF sites in the target environment.
+   * Executes updates against a single ACSF site in the target environment.
    *
-   * @param $site
+   * @param $domain
    * @param $target_env
    *
    * @throws \Acquia\Blt\Robo\Exceptions\BltException
    */
-  public function updateAcsfSites($site, $target_env) {
+  public function updateAcsfSites($domain, $target_env) {
     $this->taskDrush()
       ->drush("cc drush")
       ->run();
-    $this->say("Running updates for site <comment>$site</comment> in environment <comment>$target_env</comment>.");
+    $this->say("Running updates for site <comment>$domain</comment> in environment <comment>$target_env</comment>.");
     $this->invokeCommand('artifact:update:drupal');
   }
 


mikemadison13:
@danepowell I know this code was just added like 2 days ago by @grasmash in https://github.com/acquia/blt/pull/2787. 

i'm a little concerned about removing $site in this context, as the commentary on the function perhaps is no longer accurate (if we aren't driving directly to $site in the command, should we "say" we are running on site if we aren't "actually" running on $site)?

danepowell:
I think I'm the first person to test #2787 in a production environment, and we expected it to need a little remediation like this.

You are probably right, we should refactor the rest of this function. I can do that. Basically, `$site` is a really misleading name for a variable in this context, it should more aptly be called `$subscription`. What we colloquially think of as a "site" (i.e. a specific URI or DB) is actually specified by the `$domain` (and possibly `$db_role`) here.

mikemadison13:
this looks good to me, although i have no way of testing it in practice. 

@danepowell thanks for the cleanup of the docs as well!

lcatlett:
This is related to #2794 and issues with actually running the db updates as intended are due to Drush 9.2.1 rather than ACSF or BLT bug. 

This issue does however bring up a very important topic regarding ACSF and general multisite deploys that should be discussed further re: whether `--site` or `--uri` passed as a parameter into `blt artifact:acsf-hooks:db-update` should take precedence during deploys/updates. 

Historically when `--uri` has been hard coded or managed in config it has presented a problem with ACSF deploys and site installs since the uri may not be known prior to executing the BLT task which is why I used `$_SERVER['HTTP_HOST'];` in the post-install factory hook. A lot of strange issues can also result on updates if tasks such as Drupal and Vernish cache rebuilds are executed against domains that are not set as the primary ACSF domain. The `$4` parameter provided by the db-update factory hooks was used to set `--uri` to account for this scenario since ACSF outputs the primary public domain rather than secondary or the internal factory domain.

danepowell:
Yes, as the title and OP suggests this is the fix for #2794. As I mention there though, I don't see how this is in any way related to Drush versions. Let's restrict discussion on that topic to either this PR or that parent issue so we're not splitting attention.

I hear your concern about `--uri` vs `--site` but that doesn't sound like it should block merging this fix.

@mikemadison13 we've been using this fix successfully for about a week now, what other sort of validation would you like to see to get it merged?

lcatlett:
There are a few issues causing db update to fail which have been addressed in https://github.com/acquia/blt/pull/2846. Regarding this proposed fix, I was unable to get the same results when running the testing steps or when creating a site on ACSF, as default was still returned as the site (see below). Note that below I was manually running the commands executed by the Factory Hooks.

```
$/mnt/www/html/sandbox01live/vendor/acquia/blt/bin/blt artifact:acsf-hooks:db-update sandbox 01live tlvwux401 demo1.sandbox.acsitefactory.com --environment=01live --define drush.uri=demo1.sandbox.acsitefactory.com  --verbose --yes


 Deploying updates to default...
> drupal:config:import
[info] Skipped pre-config-import target hook. No hook is defined.
[Acquia\Blt\Robo\Tasks\DrushTask] Running /mnt/www/html/sandbox01live/vendor/bin/drush @self cc drush --no-interaction --ansi in /mnt/www/html/sandbox01live/docroot
 [success] 'drush' cache was cleared.
[Acquia\Blt\Robo\Tasks\DrushTask] Done in 0.634s
[Acquia\Blt\Robo\Tasks\DrushTask] Running /mnt/www/html/sandbox01live/vendor/bin/drush @self cache-rebuild --no-interaction --ansi in /mnt/www/html/sandbox01live/docroot
 [error]  The specified database connection is not defined: default
[Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1  Time 2.389s
[error]  Failed to import configuration!
[error]  Command `drupal:config:import ` exited with code 1.
3.062s total time elapsed.
```

Additionally, updating configuration on sites via site splits does not work as expected since --site is not passed and $acsf_site_name doesnt use the correct value in blt.settings.php. 

When adding --site explicitly, the update gets a bit further, but still fails: 

```
 /mnt/www/html/sandbox01live/vendor/acquia/blt/bin/blt artifact:acsf-hooks:db-update sandbox 01live tlvwux401 demo1.sandbox.acsitefactory.com --environment=01live --define drush.uri=demo1.sandbox.acsitefactory.com --site=demo1 --verbose --yes

 Deploying updates to demo1...
> drupal:config:import
[info] Skipped pre-config-import target hook. No hook is defined.
[Acquia\Blt\Robo\Tasks\DrushTask] Running /mnt/www/html/sandbox01live/vendor/bin/drush @self cc drush --uri=demo1 --no-interaction --ansi in /mnt/www/html/sandbox01live/docroot
 [success] 'drush' cache was cleared.
[Acquia\Blt\Robo\Tasks\DrushTask] Done in 0.537s
[Acquia\Blt\Robo\Tasks\DrushTask] Running /mnt/www/html/sandbox01live/vendor/bin/drush @self cache-rebuild --uri=demo1 --no-interaction --ansi in /mnt/www/html/sandbox01live/docroot
 [error]  The specified database connection is not defined: default
[Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1  Time 2.313s
[error]  Failed to import configuration!
[error]  Command `drupal:config:import ` exited with code 1.
2.805s total time elapsed.

```

In looking at the `artifact:deploy:update` and `artifact:acsf-hooks:db-update` tasks further, I am not sure either should be used on ACSF since ACSF already manages site context and overriding values such as the drush uri via `switchSiteContext()` appears to result in additional issues since the current implementation assumes a traditional multisite architecture.

Closing this issue to move discussion into https://github.com/acquia/blt/pull/2846.

PR #2798, opened 2018-05-04
Remove pipes from deploy sanitize find command

christopher-hopper:
Changes proposed:

- No need to pipe through three commands when one will do. 
- Improve performance for git directory remove.


mikemadison13:
my bash foo isn't awesome, but this seems to work fine and is consistent with docs.

PR #2799, opened 2018-05-04
Fixes #2797: Pass option instead of argument for deploy id

### LINKED ISSUE #2797: Deploying with a tag fails due to an undefined id
The latest `9.x` branch seems to have a bug with how it passes options through to the new `drupal:deployment-identifier:init` command. When deploying a tag it will always fail with the error output below.

Stepping through the code it looks as though the command is registered as the run invocation expects. I'm not entirely sure if this is an issue with my branch or something deeper, but I just wanted to bring this up.

```
if (!empty($this->tagName)) {
  $commands['drupal:deployment-identifier:init'] = ['id' => $this->tagName];
}
else {
  $commands[] = 'drupal:deployment-identifier:init';
}
```

My system information:
BLT version: 9.1.0

When I run this command:
```
blt deploy
blt artifact:deploy
```

I get the following output:
```
 Would you like to create a tag? (y/n) y
 Enter a valid commit message [Commit.] 
 Enter the tag name for the deployment artifact. E.g., 1.0.0-build 1.1.0-build

> drupal:deployment-identifier:init
[error]  The "id" argument does not exist. 
```

- Successfully deploy the tag

malikkotob:


mikemadison13:
tested locally with blt 9.x-dev and confirmed errors are resolved and id is now passed.

PR #2802, opened 2018-05-04
Issue #2796: Fixes failing deploys due to config status check.

### LINKED ISSUE #2796: Test for config overrides fails on ACSF, breaking deploys
My system information:
* Operating system type: 9.x-dev

I'm deploying code to ACSF using the management console, and selecting the option to apply DB updates. This results in the following well-known error, despite the fact that configuration _is not_ in fact overridden:
> Configuration in the database does not match configuration on disk...

The problem is that Inspector.php::isActiveConfigIdentical() calls `$this->executor->drush("config:status 2>&1")->run();`, and this does not use the current `drush.uri` passed via the CLI. So it runs against the default site (which isn't valid on ACSF), resulting in a 'command not found' error from Drush that BLT interprets as a failure.

danepowell:
Fixes #2796 

This Drush command needs to use whatever Drush context is provided, especially the URI in the case of site factory. Really, all of the Drush calls in this file should probably be refactored to be more consistent (like we do with the Drush task for setup and similar commands), but this solves the immediate problem.

I'm not sure if it's the best approach, but I've confirmed it works on ACSF and doesn't break anything locally.

### DIFF
--- a/src/Robo/Inspector/Inspector.php
+++ b/src/Robo/Inspector/Inspector.php
@@ -811,7 +811,8 @@ class Inspector implements BuilderAwareInterface, ConfigAwareInterface, Containe
    *   TRUE if config is identical.
    */
   public function isActiveConfigIdentical() {
-    $result = $this->executor->drush("config:status 2>&1")->run();
+    $uri = $this->getConfigValue('drush.uri');
+    $result = $this->executor->drush("config:status --uri=$uri 2>&1")->run();
     $message = trim($result->getMessage());
     $identical = strstr($message, 'No differences between DB and sync directory') !== FALSE;
 


lcatlett:
To ensure context is not lost when using vendored drush on the cli I think `--root=/absolute/path/to/site` needs to also be used. see https://github.com/drush-ops/drush/issues/2067

danepowell:
This works fine as-is. I haven't seen any problems with loss of context on ACSF. Adding the root probably wouldn't hurt, but might be unnecessary complexity. I leave that decision to @grasmash .

Like I said, we probably want to ultimately refactor how Drush is called here so we're not passing all of these arguments in a one-off way.

PR #2804, opened 2018-05-07
Force vagrant to use the virtualbox provider.

gheydon:
With the 2 box options provided, beet/box only provides a virtualbox box and the geelingguy/ubuntu1604 has not updated the vmware box in over 2 years. This change to how vagrant forces vagrant to use the virtualbox provider instead of the default provider.

Fixes # .

Changes proposed:

- This change uses the [vargant](https://www.vagrantup.com/docs/providers/default.html) environment variable VAGRANT_DEFAULT_PROVIDER to force virtualbox as the provider instead of using the local system default provider.

### DIFF
--- a/src/Robo/Commands/Vm/VmCommand.php
+++ b/src/Robo/Commands/Vm/VmCommand.php
@@ -173,8 +173,8 @@ class VmCommand extends BltTasks {
     $this->checkRequirements();
     $confirm = $this->confirm("Do you want to boot Drupal VM?", TRUE);
     if ($confirm) {
-      $this->say("In future, run <comment>vagrant up</comment> to boot the VM.");
-      $result = $this->taskExec("vagrant up")
+      $this->say("In future, run <comment>VAGRANT_DEFAULT_PROVIDER=virtualbox vagrant up</comment> to boot the VM.");
+      $result = $this->taskExec("VAGRANT_DEFAULT_PROVIDER=virtualbox vagrant up")
         ->dir($this->getConfigValue('repo.root'))
         ->printOutput(TRUE)
         ->run();
@@ -182,7 +182,7 @@ class VmCommand extends BltTasks {
         $this->logger->error("Drupal VM failed to boot. Read Drupal VM's previous output for more information.");
         $confirm = $this->confirm("Do you want to try to re-provision the VM? Sometimes this works.", TRUE);
         if ($confirm) {
-          $result = $this->taskExec("vagrant provision")
+          $result = $this->taskExec("VAGRANT_DEFAULT_PROVIDER=virtualbox vagrant provision")
             ->dir($this->getConfigValue('repo.root'))
             ->printOutput(TRUE)
             ->run();


geerlingguy:
I would rather leave this option up to the individual end user. I stopped updating the boxes for VMware because I was tired of paying another $200+ for the license and plugin license every year or so for something that nowadays only provides a marginal performance boost.

This change adds a lot of conceptual weight to the whole "run `vagrant up` and you have an environment" command... and could lead to some weird situations when someone remembers it sometimes, doesn't other times.

Maybe somewhere in the docs we can mention that if you:

  - Have VMware Fusion / Desktop on your workstation
  - Have the VMware plugin for Vagrant installed (this is a paid option... so very few users have it)
  - Don't want to customize the base box to use one of the maintained ubuntu1604 base boxes for VMware

Then you could set this option to force using VirtualBox. But in the end, I don't think it's worth adding this in as an official recommendation for the very tiny percentage of users to whom all of the above applies.

mikemadison13:
I defer to @geerlingguy's thoughts on this matter. It does seem like from a BLT perspective we don't want to force all users to change their workflow if it has a minimal impact on most day to day use cases.

gheydon:
IMO there is no boxes for any provider other than virtualbox, so executing ```blt vm``` will always fail if your default provider is not virtualbox which in this case would be very confusing for new users, especially it doesn't tell why it has not worked.

I am not asking to have the vmware box updated and I understand that few people use it. I myself have it for other clients and not always for drupal developers, also in my personal opinion I prefer VMWare over virtual box as I find it faster and much more stable.

Another option which may be a nicer fix is with the ```blt vm``` command is to add a --provider option which will default to virtualbox unless specified.

geerlingguy:
You can also set that env var in your profile, or set the default provider in a Vagrantfile.local file in your codebase to prevent having to enter it as part of the command all the time.

Also, you can use any minimal Ubuntu 16.04 base box with any provider (doesn’t have to be geerlingguy or beet); there are a few maintained ones for VMware still, I just can’t guarantee everything Drupal VM does will work with those boxes (though it should).

gheydon:
I think you have the wrong end of the stick. This is not about having VMWare boxes, it is about the fact that if you do a ```blt vm``` out of the box and your default provider is not VMWare it is going to crash.

If you follow the [onboarding](http://blt.readthedocs.io/en/latest/readme/onboarding/) document and run a ```blt vm```, follow the prompts given it will fail if you do not have virtualbox as the default provider.

I have made another change to add a --provider option which can be altered on the command line, but it defaults to virtualbox, so the onboarding will work without issue. I know you could use another box that has a vmware image but the problem is if you are setting up by using ```blt vm``` there is no way to add in a random box that you think will work. You need to manually edit the config to specify the box you want and then run ```vagrant up```

This change will make vagrant work if you are using the defaults, and I think that if you do not have virtualbox installed the errors from vagrant will make more sense.

I am not sure how to add it to this pull request now that it is closed.

PR #2807, opened 2018-05-08
Fix issue 2806, wrong files path in ACSF envs.

### LINKED ISSUE #2806: ACSF doesn't copy files when staging to others envs due to an error in blt.settings.php
My system information:
* Operating system version: Ubuntu 16.04.4 LTS
* BLT version: 9.1.0

This only happens in ACSF infrastructure when staging pro/live env to others envs. 

file_public_path in ACSF envs should be under '/g/' directory for ACSF to manage files properly. If files are stored in classic '/sites/default/files' (or probably other dirs) files are not copied when staging pro/live env to other envs.

@sdelbosc, from Acquia, has tracked down the problem to blt.settings.php file.

This check:
`$acsf_db_name = isset($GLOBALS['gardens_site_settings']) && $is_acsf ? $GLOBALS['gardens_site_settings']['conf']['acsf_db_name'] : NULL;
`
Should be:
`$acsf_db_name = isset($GLOBALS['gardens_site_settings']) && $is_acsf_env ? $GLOBALS['gardens_site_settings']['conf']['acsf_db_name'] : NULL;
 `

To see problem exceute this PHP code in an ACSF env:
`echo Drupal\Core\Site\Settings::get("file_public_path");`

If response is 

> sites/default/files

Then BLT is not detecting the ACSF env properly.

PR in short.

rsanzante:
Credit to @sdelbosc, he proposed the patch.

Fixes #2806.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -65,7 +65,7 @@ $is_ah_ode_env = (preg_match('/^ode\d*$/', $ah_env));
 $is_acsf_env = (!empty($ah_group) && file_exists("/mnt/files/$ah_group.$ah_env/files-private/sites.json"));
 // @todo Maybe check for acsf-tools.
 $is_acsf_inited = file_exists(DRUPAL_ROOT . "/sites/g");
-$acsf_db_name = isset($GLOBALS['gardens_site_settings']) && $is_acsf ? $GLOBALS['gardens_site_settings']['conf']['acsf_db_name'] : NULL;
+$acsf_db_name = isset($GLOBALS['gardens_site_settings']) && $is_acsf_env ? $GLOBALS['gardens_site_settings']['conf']['acsf_db_name'] : NULL;
 
 /**
  * Pantheon envs.


mikemadison13:
code looks good, @danepowell or @lcatlett do you have any thoughts to the actual fix before we merge it?

danepowell:
Phew... this seems like it could affect a lot more than just #2806. I have no idea where all the ACSF db name is used. I think we're going to have to test this pretty extensively.

rsanzante:
> Phew... this seems like it could affect a lot more than just #2806. I have no idea where all the ACSF db name is used. I think we're going to have to test this pretty extensively.

I would love to have those test done as now we are developing a site in ACSF with that patch. If I can help please tell me.

lcatlett:
This was a regression introduced by https://github.com/acquia/blt/commit/9fb689391729b087932a8192b05a32a06dd5884d and it  has caused a substantial number of issues. Merging now since this impacts a number  of other in-progress ACSF issues.

danepowell:
Per my previous comment, I'd really prefer if we could analyze and test what happens with this fix in place before merging. The fix itself is tiny, but if you look a few lines down, it activates a giant nested conditional that I don't really understand, and can't predict what sort of side effects (re)activating that currently unused code might be.

danepowell:
I've tested this to at least verify that it's not going to cause a regression with code / db updates on ACSF, which satisfies my primary concern. Although it looks like it's failing automated tests now.

rsanzante:
Error is:

```
 [Composer\Downloader\TransportException]                                     
  The "http://packagist.org/p/provider-2015%24e5bf6b2d88f60bb21836bd864264e09  
  0a98635172c92598cdb1a776fa018007c.json" file could not be downloaded (HTTP/  
  1.1 404 Not Found) 
```

Not sure if that error is related to the patch.

danepowell:
I restarted the test, looks good now. @mikemadison13 do you want to review?

lcatlett:
I've confirmed that this is working as expected on multiple projects - this will be required for an upcoming PR that will resolve #2806 #2796 and #2794

PR #2810, opened 2018-05-08
Cleaning up onboarding doc

danepowell:
In BLT 9, you have to run all commands inside the VM. Updated docs to reflect that.


danepowell:
Also it seems like if we're going to force people to run commands inside the VM, we should just go whole hog and get rid of the `blt vm` setup to make the process a little simpler. There's no reason to set the VM as the default local alias if you're not supposed to run commands from outside of it anyway, right?

PR #2811, opened 2018-05-09
Run simplesaml config command on all composer calls.

dpagini:
I am using SimpleSamlPhp in my project, but I almost never run `$ blt setup` any longer. I am usually pulling from our test site with `$ blt sync:refresh`. The only problem is the sync:refresh (or 9 equivalent command) will not copy in the simplesamlphp config for a developer with a new set up. Basically, if I delete my vendor folder, or update my simplesamlphp library, I am left with a broken site until I run the specific simplesaml copy config command.

This change ensures the config file gets copied into the saml folder on every run of composer install by BLT.

This is related to #2679 where I first proposed this change on BLT v8.9.


mikemadison13:
code looks good, would defer to @grasmash or others on feasibility

dooleymatt:
Agreed. I looked at this yesterday, and while I don't know how many projects use simplesaml to login to their local envs, this seems like it would improve the experience for those who do. Additionally, as far as I can tell, it shouldn't have any negative impact on projects that are not using SSO locally.

mikemadison13:
@jstephendix here is that SAML PR I mentioned to you yesterday. any thoughts?

dpagini:
ping @ba66e77 - any thoughts on this approach? Is there any chance of this getting in?

danepowell:
Looks good to me. @jstephendix or @dooleymatt have you actually tested this? As long as someone has actually tried it out, I'd be okay merging it.

ba66e77:
This looks good to my testing. Just needs to be rebased onto the latest 9.x changes. It appears it's past the point that github will allow me to do that in the UI.

danepowell:
I wasn't able to update it either so I reopened it as #3048

PR #2812, opened 2018-05-09
Cleaning up INSTALL.md

danepowell:
Fixes a number of issues, and generally cleans up / simplifies instructions.

For instance, most projects these days use PHP 7.1 instead of 5.6, and the installation method for PHP and Drush have changed due to the deprecation of the homebrew-php repo.


mikemadison13:
@danepowell this looks good, if you can please rebase i think we can merge it.

danepowell:
Thanks, should be good now.

PR #2816, opened 2018-05-11
Fix composer patches creating extraneous core directories.

danepowell:
Follow-up to #2809

This will have no effect today, but as soon as Composer Patches 1.6.5 or later is released, this configuration will stop the duplicate core/b and core/core directories appearing, as well as failing core patches.

We _could_ update composer patches to 1.x-dev to take advantage of this fix today, but I assume that's not desirable.

I've tested to ensure this works with 1.x-dev.


christopher-hopper:
Just a comment on composer patches, something I recently found out is that composer-patches has very little test coverage. I recently fixed a bug in the `patches-ignore` feature, which seemed to have been hanging around for months after the feature was added to a stable release. 

I'd proceed with caution when relying on it. We could do well to help out by writing some tests for it so that composer patches is more robust and can be improved over time.

danepowell:
There's a new release of Composer Patches that now supports this feature: https://github.com/cweagans/composer-patches/releases/tag/1.6.5

I think we should go ahead and merge this. I've tested it extensively with the new release of Composer Patches, it's pretty low risk.

@christopher-hopper I understand and share the concern about Composer Patches more generally. I don't think we can get away from it at this point, but we're open to any ideas to help support it.

PR #2818, opened 2018-05-15
Improvements to sanitize the deploy build artifact

christopher-hopper:
## Proposed Changes

The following changes have been made:

- Use Symfony Finder instead of Linux/BSD find command
- Preserve LICENSE.txt files for compliance
- Remove PATCHES.txt files
- Remove more common markdown and text files
- Remove larger tests fixtures

## Motivation

The `find` command behaves differently on Linux vs BSD/macOS. Symfony Finder should behave the same on all platforms, perhaps even on Windows :astonished: 

LICENSE files should generally not be removed from builds. It may even break the license conditions, though I'm no Open Source license expert let alone a lawyer. 

PATCHES.txt files disclose sensitive information about the code deployed in the build. They could, if exposed through the web server, provide information on possible attack vectors, installed code versions, etc. 

Lots of other common text files can be removed to reduce disclosures of sensitive information and reduce the overall build size. Using both `.txt` and `.md` file extensions to catch all variants.

Larger tests fixtures are removed for two reasons. Firstly, they're large and likely not wanted or needed in a production ready build. Secondly, some of them contain Drupal version information that can be used to determine specific versions of installed core and community modules. This was pointed out to me by a Pentest report on a Drupal 7 project. Same issues exist in Drupal 8 tests fixtures.

## Remaining Tasks

Looking for feedback and hopefully acceptance of most or all of this. I think it should be a good improvement.

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -404,38 +404,79 @@ class DeployCommand extends BltTasks {
   protected function sanitize() {
     $this->say("Sanitizing artifact...");
 
-    $this->logger->info("Removing .git subdirectories...");
-    $this->taskExecStack()
-      ->exec("find '{$this->deployDir}/vendor' -type d -name '.git' -exec rm -fr \\{\\} \\+")
-      ->exec("find '{$this->deployDir}/docroot' -type d -name '.git' -exec rm -fr \\{\\} \\+")
-      ->stopOnFail()
-      ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
-      ->run();
+    $this->logger->info("Find Drupal core text files...");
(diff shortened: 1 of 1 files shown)


danepowell:
I generally support the changes here, but I have three concerns:
1. I agree that taking out test fixtures seems risky. I'm not sure how many projects actually run those tests that require fixtures on dev/test (as opposed to just in a CI environment), but I'd rather play it safe.
2. Removing anything from the vendor directory seems superfluous, since that directory shouldn't be web-accessible anyway. I'd prefer not to mess with it in order to minimize the chances of breaking anything and improve performance.
3. Could removing README and other text files from module directories break things? For instance, I know that at least in the past, modules would basically do `file_get_contents(README.txt)` in order to generate administrative help pages. Not sure how widespread that is, but it just seems risky to me.

danepowell:
I made the minor improvements mentioned and re-rolled this as #3423

PR #2821, opened 2018-05-16
#2820 Add use statement for ClassLoader to memcache settings.

### LINKED ISSUE #2820: Memcache config is never loaded
My system information:
* BLT version: 9.x

I am surprised this has not been noticed.

In [memcache.settings.php](https://github.com/acquia/blt/blob/9.x/settings/memcache.settings.php) there is a check for `if (class_exists(ClassLoader::class)) {...}` gating the include of `memcache` related settings. This check will not be satisfied as `ClassLoader` is not defined.

greylabel:
Fixes #2820.

Changes proposed:
- Add use statement for ClassLoader to memcache settings.

### DIFF
--- a/settings/memcache.settings.php
+++ b/settings/memcache.settings.php
@@ -5,6 +5,8 @@
  * Contains memcache configuration.
  */
 
+use Composer\Autoload\ClassLoader;
+
 // Check for PHP Memcached libraries.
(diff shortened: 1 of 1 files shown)


greylabel:
@mikemadison13 Done.

PR #2824, opened 2018-05-17
Prevent PHPCS from scanning markdown files.

aweingarten:
Changes proposed:
- BLT is scanning twig files which is a headache
- According to this issue the exclude patterns are necessary to prevent PHPCS from merging. The file lists.

https://www.drupal.org/node/2867601#comment-12075633


mikemadison13:
@aweingarten you mentioned Twig in the issue but excluded md files in the PR. Was this intentional? Should we also add .twig?

aweingarten:
@mikemadison13 this was intended just to cover twig.  Title has been updated!

phpcs.xml.dist
```diff
@@ -34,5 +34,6 @@
   <exclude-pattern>*/node_modules</exclude-pattern>
   <exclude-pattern>*/vendor</exclude-pattern>
   <exclude-pattern>*/default.local.settings.php</exclude-pattern>
+  <exclude-pattern>*md</exclude-pattern>
```
dpagini: There is a section above which may be a better fit...?
```xml
<arg name="ignore" value="*.css,*.md,*.txt"/>
```

ba66e77: I wonder if this pattern is too broad. I'd have to verify, but I expect that a directory named `shazamd` would also have its contents ignored with this. As dpagini pointed out, previous sections use a `*.md` pattern.

ba66e77: I've confirmed by suspicion. With the `  <exclude-pattern>*md</exclude-pattern>` rule, if I put an improperly formatted css file inside a directory named `shabamd` (or anything else ending in md, phpcs stops seeing the css file. If I then rename the directory to `shabam`, phpcs sees the css file again.

ba66e77:
It looks to me like the test really needs to be `<exclude-pattern>*\.md</exclude-pattern>` to get .md files and not get other files/directories ending in md.

mikemadison13:
this should be covered in 10.x already, see https://github.com/acquia/blt/blob/10.0.x/phpcs.xml.dist#L11

PR #2841, opened 2018-05-31
Fixed Solr in DrupalVM.

danepowell:
Currently, if you spin up a fresh DrupalVM using BLT and try to use Solr, you'll find it doesn't work. The Solr service isn't even running, because there's some bug in the Ansible role that prevents it from starting using systemd.

BLT can work around this by adjusting its provision script.

See:
- https://github.com/geerlingguy/ansible-role-solr/pull/81
- https://github.com/geerlingguy/drupal-vm/issues/1546

### DIFF
--- a/scripts/drupal-vm/configure-solr.sh
+++ b/scripts/drupal-vm/configure-solr.sh
@@ -6,6 +6,9 @@
 # Drupal Solr module's configurations. This shell script presumes you have
 # `solr` in the `installed_extras`, and is currently set up for the D8 versions
 # of Search API Solr.
+#
+# It's also intended for Solr 4.5. For other versions of Solr, refer to the
+# example scripts provided by DrupalVM.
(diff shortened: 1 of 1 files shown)


danepowell:
This works great when you first provision your VM. Unfortunately, when you bring it up subsequently, Solr still fails to start.

mikemadison13:
I can confirm that this works great when setting up a NEW vm, however if you have an existing VM that was provisioned prior to this fix, you will likely need to execute:

```
$ vagrant ssh
$ sudo /etc/init.d/solr start
```

to get the service up and running.

geerlingguy:
@mikemadison13 / @danepowell - FYI, the issue was fixed upstream and the fix is in Drupal VM 4.9.0; we may want to revert these changes as the `service` command should work fine now.

PR #2847, opened 2018-06-02
Reverted changes to solr config script for DrupalVM.

danepowell:
As Jeff noted in https://github.com/acquia/blt/pull/2841#issuecomment-393915733, #2841 has been fixed upstream in DrupalVM so we no longer need to work around it. Woot!

I've tested and verified that Solr runs as expected with DrupalVM 4.9.0 and BLT patched with this PR.


mikemadison13:
rebasing, see https://github.com/acquia/blt/pull/2852

PR #2862, opened 2018-06-08
Redundant Drush cache-clear on deployments

danepowell:
Drush clears caches three times in a row during deployments, isn't terribly performant on ACSF (since the Drush cache lives on Gluster) and can exacerbate #2867

This removes the first cache clear, which should be totally unneccessary.


danepowell:
I've verified that the root cause of this is having the Drush home directory on a shared volume (Gluster) with reduced file i/o performance, and running multiple cache clears in parallel. This can be reproduced completely independently of BLT by simply logging in to your ACSF server in 3-4 simultaneous SSH sessions and running `drush cc drush` simultaneously in each. I'll follow up on this with either the ACSF or Drush teams.

Nevertheless, I think this PR is good to merge because it will partially mitigate the problem.

PR #2888, opened 2018-06-20
Fix typo: add missing "have"

hugovk:


ba66e77:
Thanks for catching that, @hugovk!

PR #2891, opened 2018-06-20
Creating a setup:all command.

dpagini:
I'm proposing a `$ blt setup:all` command, similar to the `$ blt drupal:sync:all-sites`. The goal is to have a developer pull down the code, and run a command to get all sites in a multisite up and running.

Some questions:
* Is there any reason we would not want this to be the default behavior of a multisite `$ blt setup` call? Why do we _just_ install the lone `default` site for people with multisite? I could obviously change this PR if that's desired.
* Why is `setup` its own command after the renaming of all the commands? Maybe knowing that would help me figure the right answer for this...


ba66e77:
@dpagini, I'm sorry it's been so long since this got any attention. I you wouldn't mind rebasing it so it's up with current 9.x I'd like to get it moving. I commented on #2921 some time ago that I think this is the better approach, but I'm open to the discussion if others feel strongly.

danepowell:
I rebased on 9.x and reopened this is #3049 so David doesn't have to waste his time on that.

dpagini:
I've rebased this branch as well. I am unable to re-open though.

danepowell:
There is something wrong with this PR (a Github bug it seems) preventing me from reopening it as well. See my comment here: https://github.com/acquia/blt/pull/3048#issuecomment-417865097

Probably easiest to just open a new PR yourself if you don't mind.

PR #2896, opened 2018-06-21
Add Acquia Cloud support to BLT alias and command

TravisCarden:
This adds support for Acquia Cloud Hosting to the `create-alias` command and to the bash alias it creates. I have tested it and confirmed that it works on the Cloud and doesn't break the local functionality.

### DIFF
--- a/src/Robo/Commands/Blt/AliasCommand.php
+++ b/src/Robo/Commands/Blt/AliasCommand.php
@@ -24,7 +24,7 @@ class AliasCommand extends BltTasks {
       if (is_null($config_file)) {
         $this->logger->warning("Could not find your CLI configuration file.");
         $this->logger->warning("Looked in ~/.zsh, ~/.bash_profile, ~/.bashrc, ~/.profile, and ~/.functions.");
-        $created = $this->createOsxBashProfile();
+        $created = $this->createBashProfile();
         if (!$created) {
           $this->logger->warning("Please create one of the aforementioned files, or create the BLT alias manually.");
         }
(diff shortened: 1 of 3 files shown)


src/Robo/Commands/Blt/AliasCommand.php
```diff
@@ -166,10 +166,13 @@ protected function getClosingBracketPosition($contents, $start_pos) {
...
-   * Creates a ~/.bash_profile on OSX if one does not exist.
+   * Creates a ~/.bash_profile on supporting systems if one does not exist.
+   *
+   * @throws \Acquia\Blt\Robo\Exceptions\BltException
    */
-  protected function createOsxBashProfile() {
-    if ($this->getInspector()->isOsx()) {
+  protected function createBashProfile() {
+    $inspector = $this->getInspector();
+    if ($inspector->isOsx() || $inspector->isAhEnv()) {
```
anavarre: As a Linux user, am I being excluded from this condition? IIRC the alias is being set up just fine when I run `blt blt:init:shell-alias` FYI.

ba66e77:
@TravisCarden, what's the use case behind this; why do you want to be able to use the `blt` alias on Acquia Cloud?

src/Robo/Commands/Blt/AliasCommand.php
danepowell: Hey @anavarre , this conditional does exclude Linux users (like you and me). However, that generally shouldn't be a problem because most Linux distros include at least a stub .bashrc or .bash_profile that BLT will happily modify, making this function unnecessary. Mac OS doesn't typically have one of these files OOTB, so BLT has to create it.

danepowell:
@ba66e77 @mikemadison13 I've reviewed this, I'd be okay with merging it if you agree.

The use case here is to do things like initial site installs on Acquia Cloud. The problem is that the codebase on ACE is not a Git clone, so the BLT alias fails.

I think this is pretty safe to merge since it only attempts to add support for Cloud, and shouldn't affect existing (local) installs.

ba66e77:
No dissent from me.

anavarre:
Filed #2972

PR #2911, opened 2018-06-25
Fixes #2880: Re-add ACSF tools.

### LINKED ISSUE #2880: Re-add ACSF tools
We removed ACSF Tools in https://github.com/acquia/blt/pull/2786 because it wasn't compatible with Drush 9

There is now a Drush9-compatible branch for ACSF Tools: https://github.com/acquia/acsf-tools/tree/9.x-dev

We should test and reintegrate it if it works so that we can have feature parity with 8.9.x.

danepowell:
Fixes #2880 

I've tested this and it works great.

I'm still in the process of testing ACSF tools itself, there do appear to be a few gaps in the 9.x branch (for instance it doesn't yet generate Drush aliases for you). But regardless, I think it's worth merging this and continuing work on ACSF Tools upstream.

### DIFF
--- a/scripts/blt/deploy/deploy-exclude.txt
+++ b/scripts/blt/deploy/deploy-exclude.txt
@@ -17,7 +17,7 @@
 /files-private
 /docroot/sites/*/private
 /docroot/themes/contrib
-/drush/contrib
+/drush/Commands
 /README.md
 /readme
 /sites/development.services.yml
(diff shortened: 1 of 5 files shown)


danepowell:
I think we need an update hook to adjust people's drush contrib path in their composer.json.

ba66e77:
Either that or we need big, bold messaging when `acsf:init:all` completes saying that you have to make that change in order to use ACSF Tools

danepowell:
My understanding is that `drush/contrib` is no longer a valid command path in Drush 9, so we should update everyone's `composer.json` regardless of whether they use ACSF tools or not. So I added an update function.

danepowell:
I've tested and verified the latest code here, ready for review again.

ba66e77:
Thanks, Dane.  I'll check it this evening.

PR #2918, opened 2018-06-27
Adding a postDbCopy cloud hook command, with additional params.

### LINKED ISSUE #2781: BLT 9 Cloud Hooks changes

rjgwiz:
Fixes #2781

Expanding on #2782 from @dpagini to add additional params that `post-code-deploy` and `post-code-update` utilize: `--environment=$target_env -v --yes --no-interaction`

To summarize why these are also required for this cloud hook (same reason it's needed for the others): without the environment override the `post-db-copy` cloud hook will try to toggle modules from `modules.local`. Verbose output is helpful, and we don't want cloud hooks to hang on any prompts.

### DIFF
--- a/scripts/cloud-hooks/hooks/common/post-db-copy/post-db-copy.sh
+++ b/scripts/cloud-hooks/hooks/common/post-db-copy/post-db-copy.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+#
+# db-copy Cloud hook: post-db-copy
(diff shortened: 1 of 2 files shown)


dpagini:
Updated #2782 with suggested changes here.

rjgwiz:
Closing in favor of #2782

PR #2921, opened 2018-06-28
Alter setup command to execute multisite.

dpagini:
Alternative to #2891.
This will run site setup against all multisites in a BLT project as the default behavior. #2891 creates a new command called `$ blt setup:all`, but this feels like more natural functionality to me. I believe this command would be most often run by new developers in which case they'd want to create all sites in their project at once.

### DIFF
--- a/src/Robo/Commands/Setup/AllCommand.php
+++ b/src/Robo/Commands/Setup/AllCommand.php
@@ -13,9 +13,20 @@ class AllCommand extends BltTasks {
    * Executes source:build:* and installs Drupal via setup.strategy.
    *
    * @command setup
+   *
    * @executeInVm
    */
-  public function setup() {
+  public function allSites() {
+    foreach ($this->getConfigValue('multisites') as $multisite) {
+      $this->switchSiteContext($multisite);
+      $this->setup();
+    }
+  }
+
+  /**
+   * Executes source:build:* and installs Drupal via setup.strategy.
+   */
+  protected function setup() {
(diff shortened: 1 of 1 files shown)


ba66e77:
I'm not averse to the idea that setting up all sites should maybe be the default behavior (though I can see a few use cases which argue against it). However, implementing this for setup would take it out of alignment with other commands like sync which use the default site and provide an `:all` option to handle multisites. As a result, I'm inclined to favor #2891 over this approach.

dpagini:
I do agree with the "out of alignment" comment, but I think the rest of the project may not have been thinking of multisite context either. It would be easy enough to run in "single-site" context by just having only 1 sites directory, or defining `multisites` as a single item array with "default". Alternatively, it would be good if `$ blt drupal:sync` or `$ blt setup` could easily be passed a flag to run only a single site... (maybe `--site site1` for example).

danepowell:
I like the idea of giving projects the ability to install all multisites with one command, but it needs to be configurable per project. Some projects may only have 2 sites and so it's no big deal to install them "all", while others might have hundreds and it would be wildly impractical to do so.

Additionally, I think the default behavior needs to remain installing/syncing a single site with the setup/sync commands.

If you open a PR meeting those requirements I'd be happy to review. I apologize that you've had to bounce around between a few PRs with this, it might be better to discuss in an issue before writing any code if you are still interested.

dpagini:
No worries, these were pretty low effort attempts. I like the idea of the config option and "opt-in". I also see the other PR had some positive feedback as well, to add a `setup:all` command. Maybe I'll try to incorporate both ideas. Thanks again.

PR #2922, opened 2018-06-29
Fixed #2867: Cache clear errors on ACSF deploys.

### LINKED ISSUE #2867: Deploy error on ACSF due to drush cache clear
About 75% of the time (on a per-site basis*) when deploying code on ACSF, I get the following error related to BLT calling `drush cc drush` as part of the db-update factory hook:
```
The db-update.sh script has encountered an error.
Completed in 5 seconds. Exit code: 1. Standard out:
[error] Failed to remove directory "/home/foo/.drush/cache/factory":
rmdir(/home/foo/.drush/cache/factory): Directory not empty
```

The root cause of this appears to be a bad interaction between the Symfony Filesystem remove() method (which gets invoked by `drush cc drush`) and the Gluster filesystem on which the Drush cache is stored on ACSF: https://github.com/symfony/symfony/issues/27578

As a workaround, I'd recommend that users of BLT on ACSF move their Drush cache directory off of Gluster and onto a local or EC2 filesystem.

One way to do this is to SSH into each server on the factory sites subscription, run `mkdir /mnt/tmp/foo01env/.drush` (replace foo01env with your current environment name), then `ln -s /mnt/tmp/foo01env/.drush`.

We should at least document this workaround and make sure users see it, maybe as part of the `acsf:init` command? Or in the ACSF readme? I'm open to suggestions for alternate workarounds as well.

*While this happens about 75% of the time _per site_, this means that as soon as you have 2-3 sites in a factory, it becomes basically impossible to deploy code.

danepowell:
Fixes #2867 

Drush is about to introduce an environment variable that allows you to specify the cache directory: https://github.com/drush-ops/drush/pull/3597

We can use this to set the cache directory to local ephemeral (block) storage rather than Gluster in order to avoid race conditions. I've verified that this fixes #2867 when using the patched version of Drush.

Ideally we would shoehorn that Drush PR into BLT somehow, but I'm not sure how to do it since we don't patch anything else in our composer.json. I suppose people who really want the patch before the next stable Drush release can apply it themselves.

### DIFF
--- a/scripts/factory-hooks/db-update/db-update.sh
+++ b/scripts/factory-hooks/db-update/db-update.sh
@@ -33,6 +33,6 @@ IFS='.' read -a name <<< "${uri}"
 # Set Drush cache to local ephemeral storage to avoid race conditions. This is
 # done on a per site basis to completely avoid race conditions.
 # @see https://github.com/acquia/blt/pull/2922
-export DRUSH_PATHS_CACHE_DIRECTORY=/tmp/.drush/${DB_ROLE}
+export DRUSH_PATHS_CACHE_DIRECTORY=/tmp/.drush/${db_role}
 
 $blt drupal:update --environment=$env --site=${name[0]} --define drush.uri=$domain --verbose --yes
--- a/scripts/factory-hooks/db-update/db-update.sh
+++ b/scripts/factory-hooks/db-update/db-update.sh
@@ -30,4 +30,7 @@ echo "$site.$target_env: Running BLT deploy tasks on $uri domain in $env environ
 
 IFS='.' read -a name <<< "${uri}"
 
+# Set Drush cache to local ephemeral storage to avoid race conditions.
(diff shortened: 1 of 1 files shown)


weitzman:
If you guys could also poke the Acquia Cloud to set this globally that would be much appreciated. I cant think of a downside.

danepowell:
I've tweaked the patch to isolate caches per site, so it's pretty much "impossible" (at the risk of jinxing myself...) to run into race conditions. A few of us have tested this patch, it should be good to merge.

bkosborne:
@danepowell Do you know if anything ever came of this as a platform-wide enhancement in Acquia Cloud?

danepowell:
It does look like some changes related to DRUSH_PATHS_CACHE_DIRECTORY were added to ACSF some time in 2020. I'm sorry, I can't tell you much more than that (3 years ago is ancient history to me! 😄 )

PR #2923, opened 2018-06-29
Recommended memcache updates

lcatlett:
Fixes # .

Changes proposed:
- Enables compression by default since Drupal's Memcache module with PHP 5.6 uses the memcache PECL extension which defaults compression to on, while PHP 7.x uses the memcached extension which defaults compression to off.
- Adds post-settings-php Factory Hook to use BLT/Acquia memcache settings with support for ACSF memcache site flag. This ensures that memcache can be enabled per-site rather than on all sites by default
- Adds BLT update hook to refresh Factory Hooks with new memcache settings and other required Factory Hook updates in BLT 9.1.0 
- Minor phpcs updates to memcache settings files


settings/memcache.settings.php
```diff
@@ -12,78 +12,82 @@
...
       ],
-    ];
+    ],
+  ];
+
+  // Override default fastchained backend for static bins.
+  // @see https://www.drupal.org/node/2754947
+  $settings['cache']['bins']['bootstrap'] = 'cache.backend.memcache';
+  $settings['cache']['bins']['discovery'] = 'cache.backend.memcache';
+  $settings['cache']['bins']['config'] = 'cache.backend.memcache';
```
greylabel: @lcatlett Question, have you done any performance comparisons of using `memcache` vs the default `chainedfast` for `cache.backend` on these three bins? In my initial testing, using `memcache` is results in worse performance. 

I can always open a separate issue for this, but would be curious to hear if you have noticed the same.

lcatlett: @greylabel if you are using PHP 7, it is very likely that performance seems degraded with memcache vs. chainedfast because of large objects in the discovery and config bins - since compression is disabled by default t

greylabel: I am very interested to try this with the compression settings added in this PR. I also was just looking at the `compressed_cache`, interesting as well. If I find time I will do some more formal benchmarking.

ba66e77:
The syntax error which Travis complained about is corrected in #2936. Closing this in favor of that one.

PR #2924, opened 2018-06-29
ACSF Saml Configuration

### LINKED ISSUE #2031: Add ACSF compatibility to BLT saml task
Currently the templates used by the simplesamlphp tasks assume ACE and not ACSF. These tasks should be updated to add ACSF support which would require dynamically setting the database based on the available gardens variables.

lcatlett:
Fixes #2031 

Changes proposed:
- Adds support for ACSF/multisite to BLT saml config by default
- Fixes typo in BLT saml htaccess patch task


ba66e77:
Subsumed by #2953. Closing

PR #2927, opened 2018-06-29
Updates for drush 9 files and removing irrelevant drush 8 references.

rjgwiz:
Expanding upon the upgrade to Drush 9 to support more drush 9 files and remove drush 8 files where possible (e.g. not retroactively touching update hooks)

### Changes proposed:

- Add support for optional `drush/sites/local.drush.yml`, for global local drush config settings/overrides that aren't related to the default site drush config (`sites/default/local.drush.yml`)
    - No error occurs if the file is not present because Drush is smart enough to check if the file exists or not before trying to load it
- gitignore `local.site.yml` (local only site aliases that would live in `/drush/sites`) as well as `drush/sites/local.drush.yml` from above (local only global drush config)
- Remove references to drush 8 config/alias files, except where used in the drush 8 to 9 upgrade function
    - Does anyone have a use case for keeping these old files?

### Questions: 

**1) What is `docroot/sites/$site/local.drush.yml` used for?** 

The default site `local.drush.yml` is loaded by BLT default in `drush/drush.yml`. Loading `options.uri` into drush's global config for the project (not on a `site` level). It seems to be needed for `$ drush` and `$ drush @self` to work. But there can only be one `options.uri` set so there's not really multisite support here. And the only places I can find the non-default `docroot/sites/$site/local.drush.yml`s being read is in `Acquia\Blt\Robo\Commands\Doctor\WebUriCheck::performAllChecks()`. 

Unless I am overlooking something, I feel like `settings/default.local.drush.yml` should be copied to `/drush/local.drush.yml` instead of `docroot/sites/$site/local.drush.yml`, with the uri default being populated by the default site's `'${project.local.uri}'`. Multisite is supported by site aliases (`$site.site.yml`) providing their own URI, or by passing uri as a command line param

**2) Why does `drush/drush.yml` use drush 8?**

It looks like `drush/drush.yml` is only used by`DrushTest::testDrushConfig()`? which verifies `drush status` works when using that drush config file. `drush.paths.include` includes the acquia cloud drush commands, and I don't think `drush.paths.drush-script` is valid, so it's not changing the remote drush command to drush 8, I think it would have to be `drush.options.drush-script`

It doesn't seem that config file is copied anywhere so it may be internal to BLT, but curious what the files intentions are

### Observation:

During the drush upgrade (called in BLT in `update_9000000()`), `drush site:alias-convert` brought along our `drush8` value from `['path-aliases']['%drush-script']` for our remote aliases. This is probably more of a drush issue than BLT, but I feel somewhere there should be a notice or prompted confirmation of if drush9 is inheriting a command path, is that still desired"? Seems if no command path is specified, it defaults to the project's `/vendor/drush` folder (if I am understanding drush's preflight correctly)

### DIFF
--- a/docs/onboarding.md
+++ b/docs/onboarding.md
@@ -35,7 +35,7 @@ If your project uses a virtual development environment such as Drupal VM:
 If your project does not use a virtual development environment:
 
 1. Setup your local LAMP stack with the webroot pointing at you project's `docroot` directory.
-1. Run `blt blt:init:settings` This will generate `docroot/sites/default/settings/local.settings.php` and `docroot/sites/default/local.drushrc.php`. Update these with your local database credentials and your local site URL.
+1. Run `blt blt:init:settings` This will generate `docroot/sites/default/settings/local.settings.php` and `docroot/sites/default/local.drush.yml`. Update these with your local database credentials and your local site URL.
 1. Run `blt setup`. This will build all project dependencies and install drupal.
 
 Please see [Local Development](local-development.md) for more information on setting up a local \*AMP stack or virtual development environment.
--- a/scripts/blt/deploy/deploy-exclude.txt
+++ b/scripts/blt/deploy/deploy-exclude.txt
@@ -29,7 +29,8 @@
 .git
 example.*
 local.settings.php
-local.drushrc.php
+local.drush.yml
+local.site.yml
 node_modules
 /vendor
 local.blt.yml
--- a/src/Robo/Commands/Doctor/DrushCheck.php
+++ b/src/Robo/Commands/Doctor/DrushCheck.php
@@ -14,7 +14,7 @@ class DrushCheck extends DoctorCheck {
   }
 
   /**
-   * Checks for local.drushrc.php file and prints messaging to screen.
+   * Checks for local.drush.yml file and prints messaging to screen.
    */
   protected function checkLocalDrushFile() {
     $drush_site_yml = $this->getConfigValue('docroot') . "/sites/default/local.drush.yml";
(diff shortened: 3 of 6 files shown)


dpagini:
I'm not sure I understand what a `local.site.yml` file is? What would go in `local.site.yml` that wouldn't go in `local.drush.yml`?

rjgwiz:
`.drush.yml`s are drush config files, `.site.yml`s are site alias files

dpagini:
So BLT out of the box will create you some aliases like `example_site.site.yml` - right? That would give you `$ drush @example_site.env status` I believe. If you have a `local.site.yml` file, the you would have something `$ drush @local.site status` I think. Wouldn't instead you want something like a `example_site.local.site.yml` or something along those lines to keep all your drush aliases under "example_site"?

I only ask b/c I google "local.site.yml" and can't really find any results on that file as a suggestion for drush.

rjgwiz:
`example_site.local.site.yml` site file won't work after Drush 9 dropped support for [alias groups](https://github.com/drush-ops/drush/issues/3107). On my local I can get a `*.*.site.yml` alias to show up under `drush sa`, but it is not usable. Alias files expect an environment name key, so you can't just do `uri: 'http://local.example.com'` in the file. That results in a `[preflight] The alias @example_site.local could not be found` error. So inside the file it would need to be:
```
local: 
  uri:  'http://local.example.com'
```
Which means referencing it as `@example_site.local.local` (different alias so it's not overriding any), and that doesn't work either: `Command "@example_site.local.local" is not defined.` which I believe is due to alias group support being removed

Where `local.site.yml` can be useful is to set any drush overrides needed to help with local dev, debugging, etc., and to make BLT commands use this override, set `drush.aliases.local` in `local.blt.yml` to an alias in `local.site.yml` (e.g. `local.example_site`).

PR #2928, opened 2018-06-30
Fix broken links in Getting Started section

kmbremner:
Path to files is /docs/ rather than /readme/. 

Changes proposed:
Fix by using relative paths.


ba66e77:
Good idea. Thanks, @kmbremner!

PR #2936, opened 2018-07-09
Recommended memcache updates, redux

ba66e77:
Changes proposed:
- Adds a single commit to #2923 to correct a syntax error.

Changes from #2923:
```
   * Enables compression by default since Drupal's Memcache module with PHP 5.6 uses the memcache PECL extension which defaults compression to on, while PHP 7.x uses the memcached extension which defaults compression to off.
    * Adds post-settings-php Factory Hook to use BLT/Acquia memcache settings with support for ACSF memcache site flag. This ensures that memcache can be enabled per-site rather than on all sites by default
    * Adds BLT update hook to refresh Factory Hooks with new memcache settings and other required Factory Hook updates in BLT 9.1.0
    * Minor phpcs updates to memcache settings files
```


scripts/factory-hooks/post-settings-php/memcache.php
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @file
+ * Example implementation of ACSF post-settings-php hook.
+ *
+ * @see https://docs.acquia.com/site-factory/tiers/paas/workflow/hooks
+ */
+
+// Use ACSF feature flag to indicate if a site should use memcache backend
+if ($is_ah_env &&
```
danepowell: We already have logic to include `memcache.settings.php` here: https://github.com/acquia/blt/blob/9.x/settings/cache.settings.php#L38

Can we move this logic over to cache.settings.php so it's all in one place instead of split between ACE / ACSF?

danepowell:
I think we need some documentation on how to use this:
- How to use the "site flag" to enable memcache on a site (BLT docs)
- What existing users need to do to switch to memcache (i.e. clear APCu... does it actually hurt anything to not clear this?) (BLT docs)
- When should you enable or not enable memcache for a given site, what are the resource constraints (i.e. 64MB shared between all sites), how many sites does that actually support, etc... ? (either Cloud or BLT docs)

scripts/factory-hooks/post-settings-php/memcache.php
lcatlett: blt settings are included in the pre-settings-php hook and ACSF has config in settings.php which needs to be overridden by the blt memcache.settings.php settings, hence the post-settings.php hook

scripts/factory-hooks/post-settings-php/memcache.php
```diff
@@ -0,0 +1,20 @@
...
+  !empty($site_settings['flags']['memcache_enabled'] &&
```
alexku: This line has a syntax error. Should be:

`!empty($site_settings['flags']['memcache_enabled']) &&`

ba66e77:
@danepowell, are the changes you requested still pending or have they been addressed?

settings/memcache.settings.php
```diff
@@ -85,5 +85,10 @@
 
     // Move locks to memcache.
     $settings['container_yamls'][] = DRUPAL_ROOT . '/../vendor/acquia/blt/settings/memcache.yml';
+
+    // Enable compression for PHP 7.
+    $settings['memcache']['options'] = [
+      Memcached::OPT_COMPRESSION => TRUE,
```
danepowell: According to the [PHP docs](http://php.net/manual/en/memcached.constants.php), the default value for this should already be true.

However, I can see that the Drupal Memcache module overrides this to FALSE, which probably isn't great for a multisite environment with shared hardware. So I agree that we should make it TRUE.

It looks like there's [already a push to enable compression](https://www.drupal.org/project/memcache/issues/2958403) with Drupal Memcache. I'd like add a comment here that links to that issue so that we know that this code can be removed once that patch is committed.

greylabel:
While we're on the subject...

Testing with compression enabled, I see moving the below bins to `memcache` causing a performance regression over using the default `cache.backend.chainedfast`. Does anyone have data showing otherwise?

```
     // Override default fastchained backend for static bins.
     // @see https://www.drupal.org/node/2754947
     $settings['cache']['bins']['bootstrap'] = 'cache.backend.memcache';
     $settings['cache']['bins']['discovery'] = 'cache.backend.memcache';
     $settings['cache']['bins']['config'] = 'cache.backend.memcache';
```

scripts/factory-hooks/post-settings-php/memcache.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @see https://docs.acquia.com/site-factory/tiers/paas/workflow/hooks
+ */
+
+// Use ACSF internal settings site flag to apply memcache settings.
+if (getenv('AH_SITE_ENVIRONMENT') && 
+	isset($site_settings['flags']['memcache_enabled']) &&
+	isset($settings['memcache']['servers'])
+) {
+  require DRUPAL_ROOT . '/../vendor/acquia/blt/settings/memcache.settings.php';
```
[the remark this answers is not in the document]
lcatlett: I believe this issue is addressed directly in #3006. BLT settings are included in a pre-settings-php hook intentionally to set the request variable values used at various stages of the bootstrap, install, and update process.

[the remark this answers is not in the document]
lcatlett: the documentation was broken out into separate tickets because the changes were so significant. Over the last few months it has been code and content reviewed by hosting and other relevant stakeholders, and published after approvals in https://backlog.acquia.com/browse/PUB-17066. These updates are live now and split across multiple pages at https://docs.acquia.com/acquia-cloud/performance/memcached/enable/ , https://docs.acquia.com/acquia-cloud/performance/memcached/pecl/ , https://docs.acquia.com/acquia-cloud/performance/memcached/drupal/ etc. 

The ACSF specific docs are in https://backlog.acquia.com/browse/PUB-17068 with additional internal and external information about  `memcache_enabled`, however it is an internal flag used in hosting and ACSF in the Acquia database config so we should only link to that page once it is published. 

Both docs sections include this exact settings configuration because it is also becoming the baseline configuration for all Acquia customers very soon and has gone through many months of production testing on ACSF and ACE on BLT and non-BLT projects so any changes to actual config, hook order, etc need to go through hosting and CL review.

settings/memcache.settings.php
lcatlett: Acquia is currently supporting that particular issue as part of their investment in a stable memcache release. (see https://www.drupal.org/project/memcache/issues/2989594). I have been more than involved with that effort so will coordinate any needed blt settings updates immediately before.

scripts/factory-hooks/post-settings-php/memcache.php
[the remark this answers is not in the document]
lcatlett: circling back to this because we made a lot of progress this weekend on the beta release which required testing this config on ACSF and the team was unable to reproduce this issue - @danepowell can you clarify how you tested and deployed this PR to ACSF including the environment (prod / 01live vs. a lower environment and if you followed the steps about enabling the flag from internal docs?

lcatlett:
@greylabel there are some benchmarks in https://www.drupal.org/project/memcache/issues/2958403#comment-12719231, and I've been finding while testing is that the bigger issue to resolve is that the Drupal 8 version of the module allows sets that are larger than 1MB which is the limit for memcached. Compressing these larger objects by default, some views of which can be 20+ MB, and then uncompressing them then seems to consume more memory than is saved by compressing them in the first place. I am working on putting together more metrics and benchmarks as part of the beta release so if I come up with more convincing data I'll add it to the issue queue.

lcatlett:
@ba66e77 and @danepowell what is needed for this PR to merge?

scripts/factory-hooks/post-settings-php/memcache.php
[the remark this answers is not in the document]
danepowell: Just to close the loop on documentation... sorry if I wasn't clear before, but the reason we need documentation is because we need to know how to enable the "site flag" for memcache in order to test this. I finally found a Confluence page that contains more details, anyone who needs a link can ping me internally.

danepowell:
My only remaining concern is the mixed results that people are reporting for compression. Ultimately it's a numbers question: how likely are sites to run out of memory with and without compression, and how much of a performance impact will they see? Without those numbers, I can't really judge.

My hunch (based only on the very shaky benchmarks in 2958403) is that there's probably not a ton of risk in using compression, so we can probably go ahead and merge this. But again, that's not based on very good data.

danepowell:
Actually based on a very recent customer experience that @ba66e77 is aware of, I'm thinking this is worth expediting.

lcatlett:
the memcacache module version in blt should be updated to the alpha6 releease before this is merged @danepowell and @ba66e77 - it has bugfixes regarding key prefixing and locking (https://www.drupal.org/project/memcache/issues/2995907)

cosmicdreams:
Additional issue reported in https://github.com/acquia/blt/issues/3055.

On 9/4 a rename / refactor / reorganization of objects was committed by the memcache module. I've uploaded the patch of that change here : https://github.com/acquia/blt/issues/3055#issuecomment-418774428

settings/memcache.settings.php
```diff
@@ -85,5 +85,8 @@
 
     // Move locks to memcache.
     $settings['container_yamls'][] = DRUPAL_ROOT . '/../vendor/acquia/blt/settings/memcache.yml';
+
+    // Enable compression for PHP 7.
+    $settings['memcache']['options'][Memcached::OPT_COMPRESSION] = TRUE;
```
anavarre: FYI compression was enabled by default in https://www.drupal.org/project/memcache/releases/8.x-2.0-alpha7

danepowell:
This PR and the related modules / services (i.e. Memcache) have evolved so much from when it was opened, I think it's better to just close it and start fresh.

See #3096

PR #2939, opened 2018-07-10
Fixes #2932: Inline example of how to require additional settings

### LINKED ISSUE #2932: Inline example of how to require additional settings
BLT downloads a stock `settings.php` file for Drupal and appends its own settings require to the end:
`require DRUPAL_ROOT . "/../vendor/acquia/blt/settings/blt.settings.php";`

However, folks sometimes need to add their own custom settings. We provide a `docroot/sites/default/settings` directory to help organize these files but don't provide an example of how to include them.

I think it would be good to include an example custom settings file, and/or an example of how to include this at the end of `settings.php`, i.e.:

```
require DRUPAL_ROOT . "/../vendor/acquia/blt/settings/blt.settings.php";
# Add any custom settings to files in the settings subdirectory and include them here, i.e.:
# require 'settings/memcache.settings.php'
```

ba66e77:
Fixes #2932: Inline example of how to require additional settings

Changes proposed:
- [x] Add a default.includes.settings.php with instructions for use, similar to the default.local.settings.php

- [x] Add a section to the docs (still need to determine where) to elaborate when and how to use the includes file

Additional tasks based on [Dane's comment below](https://github.com/acquia/blt/pull/2939#issuecomment-404009853)
- [x] have a "default" example file for global settings, like you've started here for site settings
- [x] cross-reference the site and global settings file from one another so people know they exist
- [x] add a note to the end of settings.php saying "don't modify this!"
- [x] revise docs above to speak to this additional detail


danepowell:
Interesting, I learn something new every day... BLT will automatically pull in any settings files in `docroot/sites/settings`: https://github.com/acquia/blt/blob/a3006192fa8fbdfa80ef0b5d45ac0a65e54c2bf6/settings/blt.settings.php#L236

They can be named anything in the pattern of `*.settings.php`.

Then on a per-site basis BLT also includes `sites/$site_dir/settings/includes.settings.php`.

That seems a little wonky, why not include _all_ settings of the pattern `sites/$site_dir/settings/*.settings.php`, same as for the global includes?

I think we should have a "default" example file for both global and local settings, and cross-reference them from one another so people know they exist. And then add a note to the end of settings.php saying "don't modify this!". That's especially important for ACSF users, because modifications do settings.php won't get picked up, but anything in those settings directories will (thanks to BLT's crafty use of factory hooks).

ba66e77:
Yeesh. The documentation I just added to this PR doesn't feel anywhere near sufficient given that discovery.

The only reason I can think why we wouldn't want to glob the `sites/$site_dir/settings` file is that the `local.settings.php` file and it's default variant are in there (as will be the default.includes.settings.php file, after the PR).

dpagini:
I think I'm responsible for the `includes.settings.php` piece, and I believe that was before the glob'd settings was introduced. The reason I would give for being a fan of the includes approach is that it gives a team control over which settings files and in which order to load them. With the glob approach, you'd default to alpha-order, I'm guessing?

greylabel:
IIRC, I added the glob include to have a hook to include multisite-friendly `settings.php` files since there is not an easy way to modify the main `blt.settings.php` since it in buried in `/vendor`. @dpagini's `includes.settings.php` servers a different purpose as it's site specific.

danepowell:
Thanks for the history everyone. I'd propose that we just document the existing system for now, and then consider globbing site settings as a separate issue. I can see the advantage to both globbing and not globbing, but I feel like we should be consistent in the approach we choose.

ba66e77:
I've edited the task to account for these additional details/tasks

ba66e77:
@danepowell, @mikemadison13 either of you see why these tests are failing?

danepowell:
That's pretty bizarre. I've verified that the settings files are written correctly. There's no problem that I can see in the way they are formatted. When I test this locally, everything works fine.

The other weird thing is that it's only certain builds in the matrix that are failing (specifically those excluding the "long" phpunit tests). I have no idea why that would be, but I restarted one just to see if it changes the result.

danepowell:
I still don't know what's the root cause of the test failure, but it has something to do with this line:
`$blt_glob_settings_file => $default_glob_settings_file,`

When I remove that, Travis tests all pass: https://travis-ci.org/acquia/blt/builds/423259441

src/Robo/Commands/Setup/SettingsCommand.php
```diff
@@ -75,6 +89,14 @@ public function generateSiteConfigFiles() {
       $default_local_settings_file = "$multisite_dir/settings/default.local.settings.php";
       $project_local_settings_file = "$multisite_dir/settings/local.settings.php";
 
+      // Generate default.includes.settings.php.
+      $blt_includes_settings_file = $this->getConfigValue('blt.root') . '/settings/default.includes.settings.php';
+      $default_includes_settings_file = "$multisite_dir/settings/default.includes.settings.php";
+
+      // Generate sites/settings/global.settings.default.php.
+      $blt_glob_settings_file = $this->getConfigValue('blt.root') . '/settings/global.settings.default.php';
+      $default_glob_settings_file = $this->getConfigValue('docroot') . "/sites/settings/global.settings.default.php";
```
danepowell: This is part of a loop that runs once for every multisite, and yet it doesn't do anything multisite-specific. So it's just copying the same file to the same place over and over. I'm not sure if that's bad or not, just odd.

danepowell:
Thanks, this will be merged as part of #3342 et al

Issue #2933, opened 2018-07-06
blt vm overwrites existing VM configuration (box/config.yml)


bschilt:
My system information:
* Operating system type: Mac High Sierra
* Operating system version: 10.13.5
* BLT version: BLT 9.1.0-alpha1

When I run this command:
```
blt -vvv vm
```

I get the following output:
```
$ blt -vvv vm
 Drupal VM is not currently installed. Install it now?  (y/n) y
Generating default configuration for Drupal VM...
[info] Adding a drush alias for the new VM...
[info] Creating configuration files for Drupal VM...
[Filesystem\FilesystemStack] mkdir ["/Users/brian.schilt/GIT/progleasing/box"]
[Filesystem\FilesystemStack] _copy ["/Users/brian.schilt/GIT/progleasing/vendor/acquia/blt/scripts/drupal-vm/config.yml","/Users/brian.schilt/GIT/progleasing/box/config.yml",true]
[Filesystem\FilesystemStack] _copy ["/Users/brian.schilt/GIT/progleasing/vendor/acquia/blt/scripts/drupal-vm/Vagrantfile","/Users/brian.schilt/GIT/progleasing/Vagrantfile",true]
 Which base box would you like to use?
  [0] geerlingguy/ubuntu1604
  [1] beet/box
 > 1

BLT has created default configuration for your Drupal VM!
 * The configuration file is /Users/brian.schilt/GIT/progleasing/box/config.yml.
 * To customize the VM, follow the Quick Start Guide in Drupal VM's README:
   https://github.com/geerlingguy/drupal-vm#quick-start-guide
 * To run blt or drush commands against your VM, must SSH into the VM via vagrant ssh.
 * From now on, please use vagrant commands to manage your virtual machine on this computer.

[debug] Drupal VM is not initialized.
> blt:init:settings
[Filesystem\FilesystemStack] _copy ["/Users/brian.schilt/GIT/progleasing/blt/example.local.blt.yml","/Users/brian.schilt/GIT/progleasing/blt/local.blt.yml"]
[debug] Switching site context to docgen.
[Filesystem\FilesystemStack] _chmod ["/Users/brian.schilt/GIT/progleasing/docroot/sites/docgen",511]
[Filesystem\FilesystemStack] _chmod ["/Users/brian.schilt/GIT/progleasing/docroot/sites/docgen/settings.php",511]
[...]

PR #2941, opened 2018-07-11
Do not overwrite existing VM configuration files.

bartfeenstra:
Our project has existing VM configuration files (comitted to VCS) that are overwritten with the defaults whenever BLT recreates the VM. This change makes BLT only write default configuration files if their targets do not yet exist.

See https://github.com/acquia/blt/pull/2942 for the PR targeted at BLT 8.9 (`8.9.x`).

### DIFF
--- a/src/Robo/Commands/Vm/VmCommand.php
+++ b/src/Robo/Commands/Vm/VmCommand.php
@@ -328,15 +328,23 @@ class VmCommand extends BltTasks {
   protected function createConfigFiles() {
     $this->logger->info("Creating configuration files for Drupal VM...");
 
-    $this->taskFilesystemStack()
-      ->mkdir($this->vmDir)
-      ->copy($this->defaultDrupalVmConfigFile, $this->projectDrupalVmConfigFile,
-        TRUE)
-      ->copy($this->defaultDrupalVmVagrantfile,
-        $this->projectDrupalVmVagrantfile, TRUE)
-      ->stopOnFail()
-      ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
-      ->run();
+    $configFiles = [
+      $this->projectDrupalVmConfigFile => $this->defaultDrupalVmConfigFile,
+      $this->projectDrupalVmVagrantfile => $this->defaultDrupalVmVagrantfile,
+    ];
+    foreach ($configFiles as $projectConfigFile => $defaultConfigFile) {
+      // Skip projects' existing configuration files.
+      if (file_exists($projectConfigFile)) {
+          continue;
+      }
+
+      $this->taskFilesystemStack()
+          ->mkdir($this->vmDir)
+          ->copy($defaultConfigFile, $projectConfigFile, true)
+          ->stopOnFail()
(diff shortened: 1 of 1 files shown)


ba66e77:
A very similar issue to this came up the other day. The box/config.yml was being overwritten when running `blt vm`. In that case the root of the problem was that Vagrantfile had not been committed to their repo (thus wasn't present in fresh repo clones) so BLT was treating it as a fresh VM setup. Is that the same issue you're encountering? 

See #2933 for the details.

bartfeenstra:
Thanks! We do not wish to commit the Vagrantfile at this time and are happy with BLT recreating it for now.

ba66e77:
Can I ask why you don't want to commit the Vagrantfile? Trying to understand if there's a use case here that we're not covering.

bartfeenstra:
Fair question! Not at the moment; I'd have to look that up in the archives.

src/Robo/Commands/Vm/VmCommand.php
```diff
@@ -328,15 +328,23 @@ protected function customizeConfigFiles() {
...
+      $this->projectDrupalVmVagrantfile => $this->defaultDrupalVmVagrantfile,
+    ];
+    foreach ($configFiles as $projectConfigFile => $defaultConfigFile) {
+      // Skip projects' existing configuration files.
+      if (file_exists($projectConfigFile)) {
+          continue;
+      }
+
+      $this->taskFilesystemStack()
+          ->mkdir($this->vmDir)
```
ba66e77: If the box/config.yml file doesn't exist, then we need to create the box directory (Probably. There's an odd chance that the box directory could exist but not the config.yml file, but [Symfony's mkdir operation](https://symfony.com/doc/current/components/filesystem.html#mkdir) will just ignore it if we tell it to create a dir that already exists.). But if the Vagrantfile doesn't exist, we don't need to create the box dir.

We'll also get a bit better performance if we just execute the taskFilesystemStack()->run() command once, so could you pull that out of the foreach loop?

ba66e77:
In any case, I'm inclined to agree with you that we shouldn't clobber settings if they already exist, even if we have to create other settings files. So thanks for putting in the PR.

There are a couple of improvements I'd like to see made before we merge this, in addition to the PHPCS tweaks so that it will pass the Travis checks.

danepowell:
As far as BLT and DrupalVM are concerned, everyone should be committing their Vagrantfile. Additionally, you shouldn't be running `blt vm` on a regular basis, you should only run it the very first time to set up the VM. Subsequently you should interact with the VM using `vagrant` commands.

This is all documented in-line when you run `blt vm` as well as in the DrupalVM docs, but I'll add it to the BLT docs as well. Let me know if there's any way to make it more clear. #3036

danepowell:
I'm also not sure that we want to take a piecemeal approach to generating configuration. If we merge this, then if someone already had a partial VM configured and ran `blt vm`, you could end up with a Frankenstein VM that behaves totally unpredictably. BLT should either completely own the VM, or it shouldn't.

I could maybe see an argument for aborting `blt vm` entirely if any config files already exist, but I'm not sure it's worth the lift.

ba66e77:
Given the age of this and concerns about the approach, I'm going to go ahead and close this out.

PR #2937, opened 2018-07-09
Test against Drupal 8.6.x-dev.

grasmash:
Test against Drupal 8.6.x-dev.

PR #2944, opened 2018-07-12
Upgrade PHPunit optionally to 6.5. This is needed to support PHP 7.2+

### LINKED ISSUE #2771: Request: Update PHPUnit library

seanhamlin:
Fixes #2771

Changes proposed:
- Make PHPUnit 6.5+ an option for those that need it (e.g. PHP 7.2+). For those people on <= PHP 5.6, PHPUnit 4.8 is still an option. This brings the version in line with what [Drupal core 8.6 requires](https://github.com/drupal/drupal/blob/8.6.x/core/composer.json#L61).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "grasmash/yaml-cli": "^1.0.0",
         "grasmash/yaml-expander": "^1.2.0",
         "oomphinc/composer-installers-extender": "^1.1",
-        "phpunit/phpunit": "^4.8",
+        "phpunit/phpunit": "^4.8.35 || ^6.5",
         "squizlabs/php_codesniffer": "^2.7",
         "symfony/console": "^3.4.0",
         "symfony/twig-bridge": "^3.3",


ba66e77:
Looks like Travis is showing failures on [PHP7.1 when using PHPUnit 6.5.9](https://travis-ci.org/acquia/blt/jobs/4029136620).
> PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/acquia/blt/tests/phpunit/src/BltTestBase.php on line 10
> 
> Fatal error: Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/acquia/blt/tests/phpunit/src/BltTestBase.php on line 10
> [ExecStack]  Exit code 255  Time 01:32
> [error]   

And if we're going to test support for PHP7.2 let's add it as a setting in the [.travis.yml file](https://github.com/acquia/blt/blob/9.x/.travis.yml)

Last thought of the moment is that we should consider how this PR interacts with #2937 for testing core 8.6

dpagini:
See #2780 for approach to support various versions.

danepowell:
Thanks a lot for the contribution but I'm going to close this in favor of #2780, which seems to be more comprehensive.

PR #2946, opened 2018-07-12
Fixes #2945 by removing cloud hook exception for ACSF.

### LINKED ISSUE #2945: Cloud Hooks Break ACSF Deployments
My system information:
* BLT version: 9.x

When I run this command:
Run a deployment on ACSF

I get the following output:

```
blt artifact:ac-hooks:post-code-deploy $site $target_env $source_branch $deployed_tag $repo_url $repo_type --environment=$target_env -v --yes --no-interaction
[error]  This is not intended to be executed on ACSF environments! 
```
BLT to protect me from doing stupid things.

It seems like cloud hooks cause fatal errors in ACSF. We had protection in BLT 8 that ensured that the hooks didn't try to run in ACSF (https://github.com/acquia/blt/blob/8.9.x/scripts/cloud-hooks/hooks/common/post-code-deploy/post-code-deploy.sh#L23) but that protection was removed and as a result, if you have cloud hooks installed in your code base for ACE, then they will blindly run on ACSF.

it also looks like this survived for the db scrub hook, but it has bad variables in it that need to be fixed:
https://github.com/acquia/blt/blob/9.x/scripts/cloud-hooks/hooks/common/post-db-copy/db-scrub.sh

mikemadison13:
Fixes #2945.

Changes proposed:
- remove exception for ACSF
- wraps the post code deploy hook in an if statement and only runs if NOT ACSF
- fixes variables in the db scrub cloud hook template.
- updates the db scrub to be consistent with other hooks
- updates the post code deploy to be consistent with other thooks


src/Robo/Commands/Artifact/AcHooksCommand.php
```diff
@@ -60,29 +59,17 @@ public function postCodeDeploy($site, $target_env, $source_branch, $deployed_tag
    * @throws \Exception
    */
   public function postCodeUpdate($site, $target_env, $source_branch, $deployed_tag, $repo_url, $repo_type) {
-    $this->dieIfAcsfEnv($target_env);
```
grasmash: This is here to intentionally prevent a post-deploy hook from being executed an ACSF, given that it is a dangerous practice that is discouraged.

@lcatlett right?

mikemadison13: @grasmash the issue is that with the exception it kills the deployment. the intent with this PR is to still "not" do the cloud hook, but not throw an exception in the process (so fail gracefully)

mikemadison13:
this seemed to work on my last deployments to both ACE and ACSF.

src/Robo/Commands/Artifact/AcHooksCommand.php
lcatlett: Failing gracefully on post-code-update looks fine, however I recommend adding the exception to kill the deployment to postCodeDeploy since that is the most problematic of the Cloud Hooks - it may be up for debate whether to throw the exception and terminate the deployment if $env is 01live / production as well as 01update, as using post-code-deploy on an ACSF code update in production is dangerous in all usages since it often will bypass the ACSF update environment during deployments.

mikemadison13:
@lcatlett please have another look, i added the same check to the db scrub and post code deploy hooks as well as post code update. thoughts?

danepowell:
Latest changes look good to me.

src/Robo/Commands/Artifact/AcHooksCommand.php
danepowell: Just to be clear, I doubt we want to throw an _exception_ on post-code-deploy. That would make it impossible to maintain a codebase on ACE and ACSF simultaneously. Rather, we want to bail gracefully on post-code-deploy on ACSF.

mikemadison13:
@ba66e77 how do you feel about getting this merged?
cc @danepowell  / @lcatlett

mikemadison13:
did we need to remove the scrub command commit? i rebased but it looks like that got struck out in the description?

lcatlett:
I think my last comment was deleted because the diff was updated, so posting again below:

This approach is fine for now to unblock deploys, however be aware that post-code-deploy hooks will be disabled on ACSF in the coming weeks/months, even if they are in the deployed codebase. Customers using them will hear from Acquia product and their account managers well in advance, but adding a todo or creating an issue in BLT to come back to this later might make sense.

ba66e77:
Since this is a bug, we'll need to forward-port it to 9.2.x as well.

ba66e77:
@mikemadison13, it's not clear to my why/how that scrub line in the description was struck out. I've un-struck it.

PR #2948, opened 2018-07-12
Simplesaml - Default to db storage based on environment variable.

lcatlett:
Changes proposed:
- Defaults to db session storage for simplesaml 
- Sets environment settings dynamically based on current request

### DIFF
--- a/scripts/simplesamlphp/acquia_config.php
+++ b/scripts/simplesamlphp/acquia_config.php
@@ -65,6 +65,11 @@ else {
   $config = acquia_session_store_config($config, $ah_options);
 }
 
+  // Default to database session storage.
+  if (!isset($ah_options['session_store'][$ah_options['env']])) {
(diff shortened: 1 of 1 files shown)


mikemadison13:
I continue to get an undefined environment of 01live with this code. 

i'm wondering if we should instead do something like...

```
/**
 * Get session storage configuration defined by Acquia.
 *
 * @param array $config
 *   Current configuration.
 * @param array $ah_options
 *   Acquia account specific options.
 *
 * @return array
 *   Updated configuration.
 */
function acquia_session_store_config(array $config, array $ah_options) {
  if ($ah_options['session_store'][$ah_options['env']] == 'memcache') {
    $config = mc_session_store($config, $ah_options);
  }
  else {
    $config = sql_session_store($config, $ah_options['database_name']);
  }
  return $config;
}
```

and just remove the elseif entirely

Issue #2971, opened 2018-07-24
New project setup failing with fatal error, `[error]  Could not initialize new git repository. `


lpeabody:
My system information:
* Operating system type: Ubuntu
* Operating system version: 14.04
* BLT version: 9.1.0

When setting up a new project using the following command:

`composer create-project --no-interaction acquia/blt-project my-project`

The project is not setup and results in a fatal error:

```
...
[ExecStack] git init && git add -A && git commit -m 'Initial commit.'
[ExecStack]  Initialized empty Git repository in /var/www/my-project/.git/ 
[ExecStack]  Exit code 128 
[error]  Could not initialize new git repository. 
BLT installation failed! Please execute /var/www/my-project/vendor/acquia/blt/bin/blt internal:create-project --ansi --verbose to debug the issue.
```

ba66e77:
The output there recommends running `var/www/my-project/vendor/acquia/blt/bin/blt internal:create-project --ansi --verbose` to debug. What do you get when running that?

lpeabody:
That resulted in the following:

```
[les.peabody@~/Sites/newedelman]
-> fin exec /var/www/my-project/vendor/acquia/blt/bin/blt internal:create-project --ansi --verbose
[Filesystem\FilesystemStack] remove ["/var/www/my-project/.travis.yml"]
[Filesystem\FilesystemStack] remove ["/var/www/my-project/LICENSE.txt"]
[Filesystem\FilesystemStack] remove ["/var/www/my-project/README.md"]
Copying files from BLT's template into your project...
[ExecStack] rsync -a --no-g '/var/www/my-project/vendor/acquia/blt/template/' '/var/www/my-project/' --exclude-from='/var/www/my-project/vendor/acquia/blt/scripts/blt/ignore-existing.txt' && rsync -a --no-g '/var/www/my-project/vendor/acquia/blt/template/' '/var/www/my-project/' --include-from='/var/www/my-project/vendor/acquia/blt/scripts/blt/ignore-existing.txt' --ignore-existing
[ExecStack] Running rsync -a --no-g '/var/www/my-project/vendor/acquia/blt/template/' '/var/www/my-project/' --exclude-from='/var/www/my-project/vendor/acquia/blt/scripts/blt/ignore-existing.txt' &&
[...]

PR #2974, opened 2018-07-26
Correct invocation of wizard command.

ba66e77:
Changes proposed:
- In [initializeBlt()](https://github.com/acquia/blt/blob/9058f4a9ea639d0d53595e04f6835931b2d909ca/src/Robo/Commands/Blt/UpdateCommand.php#L81), correct the wizard command invocation.

This addresses an issue identified in #2971 but I don't believe will resolve that issue.


danepowell:
Looks good to me. It appears that this is a regression introduced by https://github.com/acquia/blt/pull/2487 , where the `setup:wizard` command changed to `wizard` but this invocation wasn't updated.

PR #2975, opened 2018-07-27
Add comment to blt.yml explaining where to find a list of available properties

TravisCarden:
`blt.yml` can feel like a black box if you don't know how to find out what properties can go in it.


lcatlett:
Adding that the default values for BLT's config is in https://github.com/acquia/blt/blob/9.x/config/build.yml may offer additional clarity about the "black box" of config

TravisCarden:
Good idea, @lcatlett! Added; thanks.

ba66e77:
Thanks, @TravisCarden!

PR #2978, opened 2018-07-29
Fixes #2967: Explicitly invoke drupal:hash-salt:init in blt:init:settings

### LINKED ISSUE #2967: Hash salt wizard is broken
If you remove `salt.txt` from your repo root and run BLT commands, you'll most likely get a notification from this Robo wizard that the salt is missing:
https://github.com/acquia/blt/blob/c1dbf376419638026c8d461c12d4fa5d7b251e23/src/Robo/Wizards/SetupWizard.php#L24

The wizard then offers to fix this for you, but the fix does nothing. The problem is that the wizard calls `blt:init:settings`, and while the Settings command file does contain a method to generate the hash salt, it's not automatically invoked that I can see. It needs to be explicitly called.

ba66e77:
Fixes #2967 .

Changes proposed:
- Explicitly invoke drupal:hash-salt:init in blt:init:settings

### DIFF
--- a/src/Robo/Commands/Setup/SettingsCommand.php
+++ b/src/Robo/Commands/Setup/SettingsCommand.php
@@ -49,6 +49,9 @@ class SettingsCommand extends BltTasks {
       }
     }
 
+    // Generate hash file in salt.txt.
+    $this->invokeCommand('drupal:hash-salt:init');
+
     $default_multisite_dir = $this->getConfigValue('docroot') . "/sites/default";
     $default_project_default_settings_file = "$default_multisite_dir/default.settings.php";
 
--- a/src/Robo/Commands/Setup/SettingsCommand.php
+++ b/src/Robo/Commands/Setup/SettingsCommand.php
@@ -50,7 +50,7 @@ class SettingsCommand extends BltTasks {
     }
 
     // Generate hash file in salt.txt.
-    $this->invokeCommand('drupal:hash-salt:init');
+    $this->hashSalt();
 
     $default_multisite_dir = $this->getConfigValue('docroot') . "/sites/default";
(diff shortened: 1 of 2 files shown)


src/Robo/Commands/Setup/SettingsCommand.php
```diff
@@ -49,6 +49,9 @@ public function generateSiteConfigFiles() {
       }
     }
 
+    // Generate hash file in salt.txt.
+    $this->invokeCommand('drupal:hash-salt:init');
```
danepowell: This probably works fine, but as a minor stylistic nitpick, it seems like the convention when calling a method in the same class is to invoke the method directly, i.e. `$this->hashSalt()` instead of `$this->invokeCommand('drupal:hash-salt:init');`

danepowell:
Since we are now invoking `drupal:hash-salt:init` as part of `blt:init:settings`, we should remove the direct call to `drupal:hash-salt:init` here since it's redundant:
https://github.com/acquia/blt/blob/67af0325609f0aee9c9ae2eb4dd4d528cd4435ea/src/Robo/Commands/Setup/AllCommand.php#L26

src/Robo/Commands/Setup/SettingsCommand.php
ba66e77: I waffled on that question; couldn't decide if the better style was to call it directly or to go through the extra layer of abstraction of `invokeCommand.`

dpagini: Just food for thought, I think I read that if you call the method directly in the class (ie `$this->hashSalt();`) then the robo command hooks do not run. So for example, if I wanted to run some code everytime the hash salt was generated, I could not hook into the `$this->` version...

PR #2979, opened 2018-07-29
Correcting the vm command invocation.

ba66e77:
Changes proposed:
- Change the invocation of `vm` command to pass array of associative array items.

When building the VM through the `wizard` command, I was seeing the following warning returned:
> PHP Warning:  Illegal string offset 'no-boot' in /Users/barrett.smith/Desktop/blt/src/Robo/Commands/Vm/VmCommand.php on line 81

The invocation of the `vm` command by the wizard is using `invokeCommand` and passing a single array of `['--yes']` rather than a an array containing an array of options like the `vm` command expects.

### DIFF
--- a/src/Robo/Commands/Setup/WizardCommand.php
+++ b/src/Robo/Commands/Setup/WizardCommand.php
@@ -48,7 +48,11 @@ class WizardCommand extends BltTasks {
     }
 
     if ($answers['vm']) {
-      $this->invokeCommand('vm', ['--yes']);
+      $this->invokeCommand('vm', [
+        [
+          'no-boot' => '--yes',
+        ],
+      ]);
     }
   }
 


src/Robo/Commands/Setup/WizardCommand.php
```diff
@@ -48,7 +48,11 @@ public function wizard($options = [
     }
 
     if ($answers['vm']) {
-      $this->invokeCommand('vm', ['--yes']);
+      $this->invokeCommand('vm', [
+        [
+          'no-boot' => '--yes',
```
danepowell: I've never passed arguments to invokeCommand before, but it seems like you would want to do it like this, no?
```
$this->invokeCommand('vm', [
  [
    'no-boot',
    'yes',
  ],
]);
```

ba66e77: That would end up with the [vm method](https://github.com/acquia/blt/blob/36f77b5a573c3d8fe92968791a724c133c4c5416/src/Robo/Commands/Vm/VmCommand.php#L61) getting an array with two numerically indexed values, but it expects an associative array with one value where the key is 'no-boot'.

danepowell: I guess that makes sense as such. But if `no-boot` requires a value, wouldn't it be `TRUE` or something? Why `--yes`?

I get why you'd want `no-boot` and `--yes` as separate options, I guess I just don't understand how there's a key->value relationship between them.

ba66e77: /shrug I made the change that allowed minimal modification to make the thing work as expected. I can't speak to the initial rational for why it was written as it is.

But what we have is a method that expects a $options parameter and checks the value of the 'no-boot' key within that $options array
```
public function vm($options = ['no-boot' => FALSE]) {
    ...blah, blah, blah...
    if (!$options['no-boot'] && !$this->getInspector()->isDrupalVmBooted()) {
      return $this->boot();
    }
  }
```

PR #2982, opened 2018-07-30
Fixes #2771: Support newer PHPUnit versions

### LINKED ISSUE #2771: Request: Update PHPUnit library

danepowell:
Fixes #2771

This is a minor rework of #2780. I think it addresses all open feedback in that PR.

Basically, it follows the same approach as Drupal core--using a shim to simultaneously support PHPUnit 4 and 6.


seanhamlin:
Thanks team!

PR #2983, opened 2018-07-31
Install drush launcher in VM and pin to 0.6.0.

lcatlett:
Changes proposed:
- Installs drush launcher to use vendored project drush
- Install 0.6.0 to resolve multisite --uri and --root bug

### DIFF
--- a/scripts/drupal-vm/config.yml
+++ b/scripts/drupal-vm/config.yml
@@ -104,6 +104,10 @@ php_memory_limit: "512M"
 # Acquia Search defaults to 4.5.1.
 solr_version: "4.5.1"
 
+# Use drush launcher with 0.6.0 default
+drush_launcher_install: 'yes'
+drush_launcher_version: 0.6.0
+
 post_provision_scripts:
   - "../../../acquia/blt/scripts/drupal-vm/configure-solr.sh"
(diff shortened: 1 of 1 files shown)


ba66e77:
@lcatlett, can you give me some more background on what issue this PR addresses and what the testing steps are? Is there a related issue?

lcatlett:
This resolves a bug in drush-launcher that was causing sqlc drush commands without an alias to fail if run outside of the vendored drush directory even if passing `--root=` and `--uri=`. see https://github.com/drush-ops/drush-launcher/pull/44

ba66e77:
With this change in place, I destroyed my drupalvm instance, rm'ed the box directory, and re-ran `blt vm` to build a new VM. When drupalvm boots, I'm getting the following error

ba66e77:
With `"yes"` I got the same results as when it was in single quotes. I ended up changing the line to `drush_launcher_install: true` and that seems to have gotten it installing correctly.

ba66e77:
Following up from side conversations, it was determined that geerlingguy's drush role for DrupalVM already accomplishes this if the box is >=4.9. So we dont need this code but may want to update to require DrupalVM 4.9. 
> ./vendor/geerlingguy/drupal-vm/provisioning/roles/geerlingguy.drush/defaults/main.yml:3:drush_launcher_install: yes

PR #2990, opened 2018-08-03
Fix phpunit version detection

danepowell:


ba66e77:
Fix is passing these builds and the canary builds. Merging

PR #2992, opened 2018-08-04
Fixes #2964: When manually deploying using a tag also push that tag to source repo.

### LINKED ISSUE #2964: tag source repo when a tag is deployed to the artifact repo
As a BLT user deploying manually rather than using CI, I want the source repo to be tagged when I tag the artifact repo so that I can easily tell which commit in the source repo my artifact was generated from.

## Discussion
The [release process documentation](http://blt.readthedocs.io/en/latest/release-process/#tagging) recommends tagging the master branch in your source repo when manually building and tagging an artifact. In practice it seems like this is either ignored or users are not aware of the recommendation. That, combined with [the way in which BLT pushes tags](https://github.com/acquia/blt/blob/9.x/src/Robo/Commands/Deploy/DeployCommand.php#L179), results in tags in the artifact repo which cannot be tied back to matching commits in the source repo.

Rather than making a recommendation and counting on business process to enact it, lets help users do the right thing by automatically creating a tag in the source repo when a user supplies a tag for the artifact repo during the `deploy` process.

ba66e77:
Fixes #2964 .

Changes proposed:
- Add config setting for controlling if source repo should be tagged.
- On manual deploy, if tag is created also tag the source repo.

@todo

- [x]  Update the deploy message to say something more clear than "Typically, you would only create a tag if you currently have a tag checked out on your source repository."
- [x]  Add documentation of this feature.
- [x]  If config is set to FALSE and deploying tag, provide message that source will not be tagged also.


ba66e77:
The documentation just pushed is the last item I think we need for this. I think it's ready for review.

danepowell:
Thanks for the testing steps, I also followed those for an existing BLT install and it seemed to work great.

PR #2994, opened 2018-08-05
Exclude the 'sites/settings' dir for the default list of sites.

### LINKED ISSUE #2989: shared settings files for multisite
According to the `/vendor/acquia/blt/settings/blt.settings.php`, any settings.php files in `/docroot/sites/settings` will be included. I created a base.settings.php and a shared.settings.php file in that location. When the command `blt blt:init:settings` is run, it does indeed included them, but it treats the `/docroot/sites/settings` as if it's a multisite and adds all the other settings files in that directory so it looks like:
```
+-_settings
|  +-- _settings
|   | +-- default.local.settings.php
|   | +-- local.settings.php
| +-- base.settings.php
| +-- default.local.drush.yml
| +-- default.settings.php
| +-- local.drush.yml
| +-- settings.php
| +-- shared.settings.php
```

How do I have a "shared settings" directory without getting it turned into a "site" directory?

ba66e77:
Fixes #2989.

Changes proposed:
- Add 'settings' to the list of directories ignored by `getSiteDirs()` when a `multisites` config array has not been defined.


ba66e77:
*To replicate the problem*
1 - create a /docroot/sites/settings directory
2 - run `blt blt:init:settings`
3 - list /docroot/sites/settings directory and notice that other files from the /docroot/sites directory have been added to /docroot/sites/settings as if it were a multisite directory.

*To confirm the solution*
1 - with the changes in this pr applied, repeat the above process and at step 3 notice the contents of /docroot/sites/settings are unchanged.

ba66e77:
This is listed as a bug, so it should still be applied to the 9.1 series and forward-ported to 9.2.x when we have an approved solution.

PR #3003, opened 2018-08-07
Change default project.local.hostname to ${project.machine_name}.localhost

TravisCarden:
[RFC 2606](https://tools.ietf.org/html/rfc2606) reserves the TLD `.localhost` for precisely the local development use case. It seems better to use it than to "abuse" `.com`.

> The ".localhost" TLD has traditionally been statically defined in host DNS implementations as having an A record pointing to the loop back IP address and is reserved for such use.  Any other use would conflict with widely deployed code which assumes this use.

### DIFF
--- a/template/blt/blt.yml
+++ b/template/blt/blt.yml
@@ -14,7 +14,7 @@ project:
   # This will be used as the local uri for all developers.
   local:
     protocol: http
-    hostname: local.${project.machine_name}.com
+    hostname: ${project.machine_name}.localhost
 
 # Configuration settings for new git repository.
 git:


ba66e77:
@TravisCarden it looks like there's some history in #516 as to why we don't use the *localhost TLD. tl;dr - mostly because it makes Windows sad

TravisCarden:
Ah! Well, I can live with that. Thanks for the explanation, @ba66e77.

greylabel:
Using a local domain structure that does not resemble your production services can sometimes be a bit of an anti pattern -- potentially does not play as well with cookie sharing, CORS, and other features encountered in enterprise builds.

Issue #3005, opened 2018-08-10
Refactor server and other globals during site install and deploy:update


lcatlett:
As discussed in #3000, there are critical contextual variables that Drupal, BLT, various contrib modules, and Acquia Cloud configuration depend on that are set in super globals. This was refactored in Drush 9 and remains in Drush 8, however site-install may be broken before Drupal 8.6 is released as Drush is updated on projects. 

This issue is to capture the changes that will be prioritized in BLT.

dpagini:
Closed by #3172?

Issue #3001, opened 2018-08-07
Remove cache-rebuild prior to updb


seanhamlin:
My system information:
* Operating system type: Acquia Cloud Enterprise
* Operating system version: Ubuntu Xenial
* BLT version: 9.1.0

This is happening on a Drupal 8.5.6 upgrade (from Drupal 8.4.x). The Cloud hooks is triggering BLT to do the following commands.

When I run this command:
```
/mnt/users/[SITE]/uat.shell /var/www/html/[SITE].uat/hooks/common/post-code-deploy/post-code-deploy.sh [SITE] uat upgrade-build upgrade-build [SITE]@svn-[SERVER].prod.hosting.acquia.com:[SITE].git git < /dev/null (as [SITE].uat@staging-[SERVER]})
```

I get the following output:
```
<< [2018-07-31 07:44:36] Starting hook: post-code-deploy
<< [2018-07-31 07:44:39] Executing: /mnt/users/[SITE]/uat.shell /var/www/html/[SITE].uat/hooks/common/post-code-deploy/post-code-deploy.sh [SITE] uat upgrade-build upgrade-build [SITE]@svn-24511.prod.hosting.acquia.com:[SITE].git git < /dev/null (as [SITE].uat@staging-[SERVER]})
<<
<< site="$1"
<< target_env="$2"
<< source_branch="$3"
<< deployed_tag="$4"
<< repo_url="$5"
<< repo_type="$6"
<<
<< # Prep for BLT commands.
<< repo_root="/var/www/html/$site.$target_env"
<< export PATH=$repo_root/vendor/bin:$PATH
<< cd $repo_root
<<
<< blt artifact:ac-hooks:post-code-deploy $site $target_env $source_branch $deployed_tag $repo_url $repo_type --environment=$target_env -v --yes --no-interaction
<< Running updates for environment: uat
<< > artifact:update:drupal:all-sites
<< Deploying updates to default...
<< > drupal:config:import
<< [info] Skipped pre-config-import target hook. No hook is defined.
<< [Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/html/[SITE].uat/vendor/bin/drush @self cc drush --no-interaction -v --ansi in /var/www/html/[SITE].uat/docroot
[...]

Issue #3000, opened 2018-08-06
Prepare for Drush removing superglobals during site install


danepowell:
### Background
Drush is only run from the command line. When you run scripts from the command line, certain global variables aren't available (such as `$_SERVER['HTTP_HOST']`), because Apache and other web services aren't being invoked. Drupal (and BLT) depends on some superglobals being set, so Drush currently ['fakes' them during site installs](https://github.com/drush-ops/drush/blob/22606d1a613ad15914a4e88eb2d2e1c3fe20c613/src/Commands/core/SiteInstallCommands.php#L377).

Note that Drupal used to depend on these superglobals in a lot more places, and Drush thus used to set them for _every_ request (not just site installs), but this changed when [Drupal core stopped using them](https://www.drupal.org/project/drupal/issues/1998638), so [Drush stopped setting them](https://github.com/drush-ops/drush/pull/3109/files).

### Upcoming breaking changes
In 8.6.0, Drupal core is going to [stop using superglobals during site installs](https://www.drupal.org/project/drupal/issues/2958176) as well, and in response [Drush is going to stop setting them](https://github.com/drush-ops/drush/issues/3506).

### Action items
We need to determine what impact that Drush change will have on BLT, which might be using those superglobals for various things. For instance, we currently use `$_SERVER['HTTP_HOST']` to figure out [which multisite is running locally](https://github.com/acquia/blt/blob/887c4fb571898d8df7d9a96ecee8bec15c06de32/settings/blt.settings.php#L119). Presumably, that variable will no longer be available when running Drush commands with the new Drush release.

One way to see what will break is to simply hack out `serverGlobals()` in `drush/drush/src/Commands/core/SiteInstallCommands.php`. Obviously we should also grep for the use of superglobals.

Then we need to test a fix. This won't be possible until the upstream Drush issue has an associated PR: https://github.com/drush-ops/drush/issues/3506

As a side note, I'm curious how pretty much anything at all is working right now outside of site installs. For instance, it seems like `$_SERVER['HTTP_HOST']` wouldn't be set during any Drush commands except `site-install`, so how is BLT's existing settings working during Drush calls? Or is it not?

lcatlett:
I'm breaking this out into blocking tasks that are an immediate priority and then longer term changes for tracking purposes

cosmicdreams:
Currently running into this issue while updating a Site Factory-enabled site to Drupal 8.6.1.  Failures everywhere.

lcatlett:
[...]

PR #3006, opened 2018-08-13
Deploy / Update Hook Enhancements and Bugfixes for Drush 9

### LINKED ISSUE #2957: Extend Drush cache-clear collision protection
In #2922, we added cache-clear collision protection for ACSF sites, which solved this problem in the vast majority of cases.

However, there are still some edge cases where this problem can occur. For instance, if a deploy takes place in two _environments_ at the same time (as opposed to two sites on the same environment), or if this happens on ACE (which we haven't added protection for).

1. Include the environment in the cache directory environment variable: `export DRUSH_PATHS_CACHE_DIRECTORY=/tmp/.drush/${env}/${db_role}`
2. Add this environment variable to Cloud Hooks as well, so ACE users can benefit

This is a follow up from https://github.com/acquia/blt/issues/2951#issuecomment-405700510

lcatlett:
Fixes #2957 #3001 #3000 #3005 

Changes proposed:
- Creates and sets unique drush cache directory per site process for site install and 
- Removed redundant cache clears resetting site context
- Updates drush and pins to latest commit in dev-master to add configurable DRUSH_PATHS_CACHE_DIRECTORY
- Ensures DRUSH_PATHS_CACHE_DIRECTORY config is inherited by child processes spawned in the current session
- Removes deprecated global variables in favor of requests
- Updates `drush.yml` template with cache directory default
- Updates Acquia host forwarding and site path settings.php config to remove globals in favor or requests
- Todo: Update Cloud Hooks that execute drush 9 and BLT commands

To test post-install and db-update locally or manually you would execute the same command outout by ACSF in the logs, substituting your site, environment, database role, and uri as needed: 

`CACHE_PREFIX='/mnt/tmp/sandbox.01live' \drush8 --root='/mnt/www/html/sandbox.01live/docroot' -l 'bash.sandbox.acsitefactory.com' scr --script-path='/mnt/www/html/sandbox.01live/factory-hooks/post-install' 'post-install.php'`

`/mnt/www/html/sandbox.01live/factory-hooks/db-update/db-update.sh' 'sandbox' '01live' '[db_role]' 'bash.sandbox.acsitefactory.com'`

### DIFF
--- a/scripts/blt/cache.php
+++ b/scripts/blt/cache.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * @file
+ * Creates a unique temporary cache directory given a site, env, and uri.
+ * This is used for every blt command invocation on site install and update.
+ */
+
+$site = $argv[1];
+$env = $argv[2];
+$uri = $argv[3];
+
+
+if (empty($argv[3])) {
+  echo "Error: Not enough arguments. Site, environment, and uri are required.\n";
+  exit(1);
+}
+
+// Create a temporary cache directory for this drush process only.
+$cache_directory = sprintf('/mnt/tmp/%s.%s/drush_tmp_cache/%s', $site, $env, md5($uri));
+shell_exec(sprintf('mkdir -p %s', escapeshellarg($cache_directory)));
+
+if (!file_exists($cache_directory)) {
+  syslog(LOG_ERR, sprintf('Drush updates could not be executed, as the required cache directory [%s] is missing.', $cache_directory));
+  die('Missing or corrupted drush cache for this process.');
+} else {
+	echo "$cache_directory";
+}
--- a/scripts/blt/cache.php
+++ b/scripts/blt/cache.php
@@ -23,6 +23,7 @@ shell_exec(sprintf('mkdir -p %s', escapeshellarg($cache_directory)));
 if (!file_exists($cache_directory)) {
   syslog(LOG_ERR, sprintf('Drush updates could not be executed, as the required cache directory [%s] is missing.', $cache_directory));
   die('Missing or corrupted drush cache for this process.');
-} else {
-	echo "$cache_directory";
+}
+else {
+  echo "$cache_directory";
 }
--- a/scripts/factory-hooks/db-update/db-update.sh
+++ b/scripts/factory-hooks/db-update/db-update.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+##!/bin/bash
 #
 # Factory Hook: db-update
 #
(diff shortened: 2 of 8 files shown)


scripts/blt/cache.php
```diff
@@ -0,0 +1,29 @@
...
+$uri = $argv[3];
+
+
+if (empty($argv[3])) {
+  echo "Error: Not enough arguments. Site, environment, and uri are required.\n";
+  exit(1);
+}
+
+// Create a temporary cache directory for this drush process only.
+$cache_directory = sprintf('/mnt/tmp/%s.%s/drush_tmp_cache/%s', $site, $env, md5($uri));
```
danepowell: How is this functionally different from the existing code? It looks like this PR is assigning a cache directory based on site URI, whereas the existing site is assigning it based on db_role. Shouldn't that basically do the same thing?

lcatlett: passing the directory as a cli prefix ensures that the child processes spawned from the current shell inherit the value. Exporting the value in the bash shell could still result in another process changing or overriding what should be a unique value. The db_role alone isnt enough to isolate the calls to cache rebuild in particular because on ACSF the live and update environments share a database, so env differentiates the two. It is more significant that that the cache directly is in /mnt as opposed to /tmp and that this directory is also created and deleted to avoid permissions issues and race conditions

danepowell: My understanding is that using `export` makes the variable available to all child / sub-processes. Ref: https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export

lcatlett: the issue is that the value can then be modified by exec or other subprocesses of the blt and respective drush commands which does not prevent cache collisions on environments with multiple web servers. if you have suggestions on how to better solve this I am happy to refactor.

ba66e77:
Please split this up into more atomic PRs so they're easier to review.

lpeabody:
@lcatlett @danepowell @ba66e77 I recently had to apply the .patch instance of this PR in an ACSF project to get sites installing and updating on PROD (the fix was recommended by Acquia support).

Is there a plan to get this work merged into the main line of development? I'm concerned that the project is now supported by a patch that has no traction.

lcatlett:
thanks for following up from your Acquia support crit @lpeabody - you are spot on that these are critical fixes for updating to 8.6, site installs from config, multisite config split support, and deploy / updates to applications using multi-headed web servers (such as ACSF production). 

@ba66e77 and @danepowell what needs to be done to make this easier for you to review? I can of course put in multiple smaller PRs, but is there any other constructive feedback as to why this was not merged?

scripts/factory-hooks/db-update/db-update.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+##!/bin/bash
```
lpeabody: Extra hash here I had to comment out in order to get bash to run this.

lpeabody:
@lcatlett @ba66e77 @danepowell Merge this PR please. The above is extraordinarily nitpicky and it's preventing critical changes from making their way in. Also this has now happened to me TWICE (two separate projects), only manifesting itself once the code reached 01live. The first time wasn't a big deal because we hadn't launched any sites yet. This 2nd time however, 30+ sites went down for an extended period of time.

This fixes a critical issue with with sites updating to 8.6 on ACSF. Please merge this PR now or actually provide some useful feedback - your customers are pissed and unhappy. PLEASE let me know if there's anything I can do to help. I'd rather not have to deal with this again.

lcatlett:
@lpeabody - I hear you and I've created #3158 to track the progress of getting these fixes in as soon as possible. I understand your frustration and this has been raised to multiple groups at Acquia to get your specific issues resolved and prevent breakdowns like this moving forward.

PR #1435, opened 2017-04-25
BLT-1047: adding ability to customize paths for phpunit tests.

mikemadison13:
Fixes #1047  .

Changes proposed:
- adds configurable paths for PHPUnit tests so that tests can live inline with custom modules
- adds configuration parameter for each PHPUnit test so that drupal may be bootstrapped on demand

PR #3007, opened 2018-08-13
Allow PHPUnit to bootstrap from core.

danepowell:
As part of #2982, we created our own PHPUnit `bootstrap.php` file in order to support multiple versions of PHPUnit.

This created a bit of a problem for folks testing Drupal core and modules. Because BLT forces you to use that bootstrap file, you can't use core's own `bootstrap.php`, and thus if you try to run a PHPUnit test according to the [BLT docs](https://github.com/acquia/blt/blob/9.x/docs/testing.md#configuration-1) you'll end up with errors like this:

> PHP Fatal error:  Class 'Drupal\Tests\UnitTestCase' not found in docroot/modules/contrib/simple_oauth/tests/src/Unit/EntityCollectorTest.php on line 19

I see a few ways to tackle this problem. In general, I question the wisdom in trying to maintain our own copy of `bootstrap.php`. Why not just use core's? And why not simply delegate the choice of which bootstrap file to use to `phpunit.xml`?

This PR takes a middle road by just removing the CLI argument forcing you to use BLT's bootstrap file. But I'd love to hear additional arguments for or against removing our custom bootstrap file altogether, and just use core's instead.


mikemadison13:
@danepowell FYI we CAN bootstrap from core right now using the config key in the phpunit options in blt.yml. See https://github.com/acquia/blt/pull/1435. I'm not sure if this is significantly different than what you are trying to do? but you should be able to do this and bootstrap core without any code changes today:

```
phpunit:
  -
    path: '<your test directory>'
    config: '${docroot}/core/phpunit.xml.dist'
```

ba66e77:
If the config approach solves the issue of being able to bootstrap and run the tests, then the question is more about maintainability and if we _should_ be trying to run a custom bootstrap.php.

My default stance is to remove what we don't actually need, but I don't have strong feeling about it in this case. @danepowell I assume you're advocating for removing it. @mikemadison13 are you advocating for keeping it?

danepowell:
It looks like updating the PR pulled in some extraneous commits :cry:

ba66e77:
Closing in favor of #3071, which just cherry-picks the meaningful bit of this PR into the 9.2.x branch.

PR #3018, opened 2018-08-20
EditorConfig: Use tab indents for *.mk makefiles

hugovk:
Changes proposed:
- We have a `docker.mk` makefile which also requires tab indents, or it will error
- Add `*.mk` alongside `Makefile`
- If we edit `.editorconfig` ourselves, it will be overwritten by the next `acquia/blt` update

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 
 # Tab indentation (no size specified)
-[Makefile]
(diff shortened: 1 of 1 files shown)


ba66e77:
I've verified this doesn't break anything, but since a) I'm not running a Docker instance, and b) we don't officially support Docker, I can't swear it actually solves the described issue. I'm willing to take your word for it though.

Thanks for the contribution!

PR #3025, opened 2018-08-24
Add Drupal Spec Tool to Composer suggestions

TravisCarden:
Let's be friends! 😉


ba66e77:
hello, friend

Since the spec tool is becoming part of standard practice for Acquia PS builds, I think it makes sense to include it in the BLT tooling.

PR #3036, opened 2018-08-28
VM setup tips

danepowell:
This is just a minor doc update in response to some confusion about what files should be committed in #2941

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -27,6 +27,11 @@ To use [Drupal VM](http://www.drupalvm.com/) with a Drupal project that is gener
 1. Create & boot the VM
 
         blt vm
+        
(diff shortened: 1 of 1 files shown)


docs/local-development.md
```diff
@@ -27,6 +27,11 @@ To use [Drupal VM](http://www.drupalvm.com/) with a Drupal project that is gener
 1. Create & boot the VM
 
         blt vm
+        
+1. Commit all resulting files, including `box/config.yml` and `Vagrantfile`.
+        
+        git add -A
+        git commit -m "Added DrupalVM."
```
ba66e77: This commit message wouldn't pass the pre-commit hook's check for project initials and ticket number, I think.

danepowell:
Thanks for the tweaks, that totally makes sense. Sorry I couldn't circle back to this sooner.

PR #3031, opened 2018-08-26
Multisite setup and breakfix

lcatlett:
Fixes #2907, #2996, #2814, #2930, #3030, #3024, #2930, #3029

New Changes proposed: 
- [x]  Manage local multisite $sites[] via local.sites.php includes
- [x]  Exclude local multisite settings from deploy artifacts and vcs

Original Changes proposed:
- [x]  Validate mysql connection with output of drush status rather than drush sqlq for `blt internal:drupal:install` and `blt drupal:install`
- [x]  Create database with drush credentials if it does not exist
- [x]  Grants needed mysql user / db permissions for all multisite database names for all hosts including localhost
- [x]  Remove drupal_* namespace restriction for mysql database name grants
- [x]  Create sites.php file if it does not exist
- [x]  Generate sites.php sites array mapping project multisite local uris to site directories
- [x]  Remove default local.* config/settings files from new multisite directory to regenerate db and drush config from input
- [x]  Fix multisite local drush alias config
- [x]  Populate multsite database credentials from blt and user input
- [x]  Simplify use of $acsf_site_name in multisite setup
- [x]  Exclude default site files from multisite directory
- [x]  Update README

Testing steps

Generate multisite site files, vm config, and settings.php config

 1. run `blt multisite`. Validate that the task: 
 - creates new multisite in docroot/sites/[sitename]
 - generates vm host in box/config.yml
 - generates vm db config in box.config.yml
 - generates local.settings.php and local.drush.yml with new multisite config
 - generates sites.php with new multisite mapping local site hostname to site directory:
 - Excludes files directory from default site
 - Creates drush alias in drush/sites/[sitename].site.yml
 - drush alias returns correct uri, site directory, and db credentials

2. Re-provision vm to add new multisite host and database
- Exit from vm and reprovision with `vagrant reload --provision`
- ssh to vm: `vagrant ssh`

3. Multisite install/setup
- Install drupal / set up blt: `blt setup --site=[sitename]`
- Validate that database is dropped and created automatically
- Validate that site is installed according to config in sites/[sitename].blt.yml / local.settings.php
- Once installed, both `drush @sitename.local st` and `drush --root=path/to/docroot --uri=hostname` should return the correct site uri and site path and the database in `sites/[sitename]/settings/local.settings.php` should bootstrap

Issue #3027, opened 2018-08-24
Cannot install sites via Drush alias


danepowell:
You used to be able to install sites via Drush alias. For instance, it was very useful to define a local alias that would SSH into your VM, so that you could run tasks like `blt setup` outside of the VM.

This is no longer possible in 9.1.2, as it attempts to run raw mysql commands outside of the context of Drush.

This is a regression introduced by https://github.com/acquia/blt/pull/2997

Issue #3026, opened 2018-08-24
ACSF deployments broken on 9.1.2


danepowell:
BLT 9.1.2 overwrites the sites.php file provided by `acsf-init` when running deploys. With this file overwritten, it's impossible to deploy code in ACSF environments. ACSF users affected by this bug will see the following error when attempting to update code ` [error]  The file sites.php is out of date.`

We encountered this error while deploying code using Pipelines, but I _think_ it would occur regardless of CI solution, as long as you are deploying code with `blt deploy`.

The regression was introduced here: https://github.com/acquia/blt/pull/2997

PR #3037, opened 2018-08-29
Fixes 3029: Reintroduce multisite enhancements with modifications

ba66e77:
This PR subsumes #3031, reverting the changes which cased the regression in #3027 (
523ac5e) and ensuring php opening tags are written into `sites.php` to correct the failures of TravisCI checks (20bc0a0).

### DIFF
--- a/docs/multisite.md
+++ b/docs/multisite.md
@@ -6,44 +6,19 @@ There are two parts to setting up a multisite instance on BLT: the local setup a
 
 1. Set up a single site on BLT, following the standard instructions, and ssh to the vm (`vagrant ssh`).
 1. Run `blt recipes:multisite:init`.
-    
+
     Running `blt recipes:multisite:init`...
-    
+
     * Sets up new a directory in your docroot/sites directory with the multisite name given with all the necessary files and subdirectories.
     * Sets up a new drush alias.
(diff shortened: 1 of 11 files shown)


danepowell:
I tested this on ACSF and found that it caused the same regression described in #3026. The problem is that sites.php is getting modified from the stock version provided by ACSF, so ACSF crashes during code updates.

To reproduce:
1. Start with a project that has been initialized on ACSF, or create a new one. Shouldn't really matter either way.
2. Run `blt blt:init:settings`. Note that `docroot/sites/sites.php` in the source directory is now modified to include `local.sites.php`. This is by design according to this PR.
3. Run `blt deploy:build`. Note that `deploy/docroot/sites/sites.php` now also has been modified to include `local.sites.php`.
4. Push the build artifact to ACSF and try to deploy it using the "Update code" page in the management console. The expected behavior is that the code update succeeds. The actual behavior is that it fails with the error "[error]  The file sites.php is out of date.", same as described in #3026.

ba66e77:
This, then, is a battle we cant win on our own, unless we're willing to patch ACSF.

ba66e77:
Since this isn't in a place it can advance as is, let's close it out and we can cherry-pick or reimplement commits as needed to address specific parts of what is attempted here.

PR #3043, opened 2018-08-30
Removing Travis source calls.

dpagini:
Re-roll #2521.

I actually don't really have a solution here at this time. The only way I can figure out how to make this work is to not use `$ source ____` syntax (last commit). But as @grasmash [said](https://github.com/acquia/blt/pull/2521#issuecomment-367784167), that takes away quite a bit of output in the Travis logs.

So the last commit has the intended output via Travis. A script fails, but Travis still runs `after_failure` command. Does anyone have any other ideas for an approach here?


lcatlett:
executing the script with bash +x is one possible alternative to source that will output each line and also show the values assigned to each variable. This is often  how I debug scripts on CI and on Acquia environments. For example, there is a 01live.shell script on my Acquia ACSF environment that sets and exports various variable values. Running `bash -x 01live.shell` on the script runs it and  outputs the following: 		

```
+ export USER=rogue1
+ USER=rogue1
+ export HOME=/home/rogue1
+ HOME=/home/rogue1
+ export PATH=/usr/local/php7.1/bin:/usr/local/drush8:/usr/local/bin:/usr/local/php7.1/bin:/usr/local/drush8:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
+ PATH=/usr/local/php7.1/bin:/usr/local/drush8:/usr/local/bin:/usr/local/php7.1/bin:/usr/local/drush8:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
+ export AH_REALM=enterprise-g1
+ AH_REALM=enterprise-g1
+ export AH_SITE_NAME=rogue101live
+ AH_SITE_NAME=rogue101live
+ export AH_SITE_GROUP=rogue1
+ AH_SITE_GROUP=rogue1
+ export AH_SITE_ENVIRONMENT=01live
+ AH_SITE_ENVIRONMENT=01live
+ export AH_CURRENT_REGION=us-east-1
+ AH_CURRENT_REGION=us-east-1
+ export DRUSH_LAUNCHER_FALLBACK=/usr/local/drush8/drush
+ DRUSH_LAUNCHER_FALLBACK=/usr/local/drush8/drush
+ export ACQUIA_HOSTING_DRUPAL_LOG=/var/log/sites/rogue1.01live/logs/web-1577/drupal-requests.log
+ ACQUIA_HOSTING_DRUPAL_LOG=/var/log/sites/rogue1.01live/logs/web-1577/drupal-requests.log
+ export AH_NON_PRODUCTION=1
+ AH_NON_PRODUCTION=1
+ export TEMP=/mnt/tmp/rogue101live
+ TEMP=/mnt/tmp/rogue101live
+ export TMPDIR=/mnt/tmp/rogue101live
+ TMPDIR=/mnt/tmp/rogue101live
+ '[' 0 -gt 0 ']'
```

dpagini:
Thanks @lcatlett - isn't running `$ bash scriptname` basically the same as `$ . scriptname`?

So I started playing with this more after the last comment.
Going back to my comments from the last PR, I think it just comes down to the `set -e` at the top of all the scripts, and not necessarily the `$ source ____` line. The `set -e` is telling the script to exit as soon as there is a problem instead of continuing to run, so ultimately it speeds up builds. However, it's making it so that the Travis `after_failure` steps are being skipped when there is an error. So to me, there's a pro/con for each method. I'd like to suggest we keep Travis running as it's supposed to.
I guess I knew all of the above from the comments in the original PR, but became focused on the `$ source scriptname` being part of the issue.

danepowell:
It sounds like whichever approach we choose is a compromise, and all other things being equal I'd prefer to preserve the status quo. If you come up with a better solution definitely let me know.

PR #3057, opened 2018-09-05
Remove Twig cache clearing patch.

### LINKED ISSUE #3056: Remove Clear Twig caches on deploys patch to support Drupal 8.6.x
The following patch has landed in core and will no longer apply to `8.6.x`:
https://www.drupal.org/files/issues/2752961-130.patch

bobbygryzynger:
Fixes #3056.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,25 @@
-Fixes # .
+Fixes # 
+--------
 
 Changes proposed:
+---------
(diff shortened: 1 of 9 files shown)


ba66e77:
Thank you, @bobbygryzynger. Because of this and other patch issues with 8.6, we're tracking those changes against the core86 branch rather than the mainline 9.x. I've retargeted your PR against that branch.

ba66e77:
Ah, and it looks like the patch you removed has already been removed on that branch.

PR #3058, opened 2018-09-05
Fixes #3055: Update memcache settings to match changes in Drupal memcache module.

### LINKED ISSUE #3055: DrupalMemcacheConfig does not exist.  Breaks BLT.
My system information:
* BLT version: 9.1.4

Output of `blt doctor`:
```
+--------------------------------------+-----------------------------------------------------------+
| Check                                | Problem                                                   |
+--------------------------------------+-----------------------------------------------------------+
| NodeCheck:checkNodeVersionFileExists | Neither .nvmrc nor .node-version file found in repo root. |
+--------------------------------------+-----------------------------------------------------------+
```

When trying to deploy with BLT 9.1.4 those deployments fail because they expect an object named DrupalMemcacheConfig to exist.  It expects that object to exist because memcache.settings.php on line 41 sets the expectation that it will exist and it should be registered as a service.

But, the memcache module doesn't have the object and it defines a different service : memcache.settings in memcache.services.yml to use the class Drupal\memcache\MemcacheSettings

All of this leads me to believe that BLT is either out of date, or it should pin it's version of memcache to an earlier version that doesn't include this reorganization.

ba66e77:
Fixes #3055  
--------

Changes proposed:
---------
- Mirror changes in this [update to drupal/memcache](https://cgit.drupalcode.org/memcache/commit/README.txt?h=8.x-2.x&id=9b92d6244ad79eb9380e1b728ba882b880274d20)
- Require >= alpha7

Steps to verify the solution:
-----------

1. Deploy to Cloud or other wise enable memcache and verify to errors thrown.
2. ??

### DIFF
--- a/composer.suggested.json
+++ b/composer.suggested.json
@@ -13,7 +13,7 @@
     "drupal/cog": "^1.0.0",
     "drupal/devel": "^1.0.0",
     "drupal/qa_accounts": "^1.0.0-alpha1",
-    "drupal/memcache": "^2.0-alpha5",
+    "drupal/memcache": "^2.0-alpha7",
     "drupal/seckit": "^1.0.0-alpha2",
     "drupal/security_review": "*",
     "drupal/shield": "^1.0.0",
--- a/settings/memcache.settings.php
+++ b/settings/memcache.settings.php
@@ -37,12 +37,12 @@ if ($memcache_module_is_present && ($memcache_exists || $memcached_exists)) {
           'class' => 'Drupal\Core\Site\Settings',
           'factory' => 'Drupal\Core\Site\Settings::getInstance',
         ],
-        'memcache.config' => [
-          'class' => 'Drupal\memcache\DrupalMemcacheConfig',
+        'memcache.settings' => [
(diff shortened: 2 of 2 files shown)


cosmicdreams:
@ba66e77 can you explain why this additional memcache.settings.php is necessary and why only enabling the memcache module is insufficient?

danepowell:
@cosmicdreams if you take a look at the Memcache module readme, you'll see that there's a number of configuration changes that need to be made to move certain bins to memcache and to do advanced steps like enabling compression, completely removing db cache tables, etc... This code basically implements those recommendations.

The other major thing this file does is to automatically enable memcache integration _even if you don't have the memcache module enabled_. This is desirable because turning on the memcache module can lead to a chicken-and-egg problem depending on which environments actually support memcache.

Finally, note that both the Memcache module and Acquia Cloud's integration with it are rapidly evolving so some of these comments may no longer be true in the near future. This file may end up just going away.

danepowell:
I'm Dane Powell and I support this PR 😛

PR #3062, opened 2018-09-06
Lock core in 9.x branch to 8.5.6.

ba66e77:
Since 8.6.0 was released, Travis tests on the 9.x branch are failing because they try to apply patches to 8.6.0 which do not apply. To prevent that, lock the core version in 9.x to 8.5.6 and update the tests to hold tight to that version.

Changes proposed:
---------
- Lock the core version in the 9.x branch to 8.5.6 so that composer doesn't grab 8.6.
- Update the travis definition so it tests against 8.5.6 rather than a floating value.


danepowell:
I think this is probably okay since there shouldn't be _too_ many more 8.5.x releases. But if there is a new release (and it's guaranteed to be a security release), we'll have to fire-drill a new release of BLT to patch.

To get around that, I _think_ we could use a semver range like this:
`drupal/core: >=8.5.6 <8.6.0`

ba66e77:
Good thinking. We could also do `drupal/core:8.5.*` which [should accomplish the same](https://getcomposer.org/doc/articles/versions.md#wildcard-version-range-).

ba66e77:
@danepowell, what do you think of the latest change?

ba66e77:
Good thing you brought up switching to a less pinned version. I hadn't seen that core 8.5.7 came out yesterday.

ba66e77:
Ugh. With 8.5.7 something is tryting to apply the patch referenced below and failing. For the life of me I cant seem to find where it's coming from.
```
    https://www.drupal.org/files/issues/2018-07-04/2921661-86-8.6.x.patch (2921661 (follow-up): Add support to migrate multilingual revisions)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2018-07-04/2921661-86-8.6.x.patch


  [Exception]
  Cannot apply patch 2921661 (follow-up): Add support to migrate multilingual revisions (https://www.drupal.org/files/issues/2018-07-04/2921661-86-8.6.x.patch)!
```

ba66e77:
The Lightning team has said the 2.10 version of lightning-core should be released today, which will remove the patch causing problems.  So once we this merged new builds on the 9.1 branch should work as expected.

ba66e77:
lightning-core 2.10 is out and available.

PR #3068, opened 2018-09-07
Fixes #3065: Use webflo/drupal-core-require-dev instead of tracking core dev dependencies

### LINKED ISSUE #3065: Consider using webflo/drupal-core-require-dev
Whenever core adds a new dev dependency, BLT has to track it, like here:
* https://github.com/acquia/blt/pull/3008
* https://github.com/acquia/blt/pull/1725

This is obviously kind of a pain. I just discovered the webflo/drupal-core-require-dev package, which claims to track all of core's dev dependencies as normal (`require`) dependencies.

I _think_ if we added this to BLT's `require-dev`, we could stop tracking all of the separate core dev dependencies, which would be really nice.

danepowell:
Fixes #3065
--------

Changes proposed:
---------
- Currently we duplicate every single dev dependency defined by Drupal core on an individual basis. This is tedious and error prone. webflo/drupal-core-require-dev already tracks these, so if we just track that one package we can get all new dev dependencies for free.

Steps to verify the solution:
-----------
1. Start with a fresh BLT 9.2.x project.
2. Patch BLT with this PR.
3. Run composer update.
4. Verify that the only package updates are fairly trivial and mostly related to dev dependencies that core defines but BLT previously failed to include.

### DIFF
--- a/composer.required.json
+++ b/composer.required.json
@@ -11,8 +11,6 @@
   },
   "require-dev": {
     "behat/behat": ">=3.1 <3.4",
-    "behat/mink": "~1.7",
-    "behat/mink-selenium2-driver": "^1.3.1",
     "bex/behat-screenshot": "^1.2",
     "drupal/drupal-extension": "~3.2",
     "drupal-composer/drupal-scaffold": "^2.1.0",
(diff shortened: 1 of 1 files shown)


ba66e77:
Looking over the package list, there's one I don't have present after applying this patch: `sensiolabs-de/deprecation-detector`

Looks like that was added in #1973 to support the `tests:php:sniff:deprecated` command.

PR #3075, opened 2018-09-07
Fixes #3074. Further changes to update memcache settings

guitarte:
Fixes #3074
--------

Changes proposed:
---------

- Update default memcache settings to match the following more closely https://cgit.drupalcode.org/memcache/tree/README.txt?h=8.x-2.x#n138

Steps to replicate the issue:
----------
(If this PR is not fixing a defect/bug, you can remove this section.)

1. Install fresh copy of BLT 9.1.5
2. Deploy code to Acquia Cloud, and run `drush si` to try installing and run into error.

Steps to verify the solution:
-----------
(How does someone verify that this does what you think does?)

1. Patch memcache.settings.php with recommendation in PR
2. Deploy to code to Acquia Cloud and run `drush si`, installation should complete.

PR #3076, opened 2018-09-07
Finished updating memcache config for alpha7.

danepowell:
Changes proposed:
---------
We all know that the memcache module has changed a bunch of its class and service names lately (especially as of alpha7).

We attempted to update the memcache settings provided by BLT to match this in #3058 but it looks like we didn't go far enough. Things still fall over on Acquia Cloud:

> You have requested a non-existent service "memcache.config".

Here I take the approach of just copying the recommended settings verbatim from the [memcache readme](https://cgit.drupalcode.org/memcache/tree/README.txt?h=8.x-2.x).

Steps to replicate the issue:
----------
1. Install a vanilla BLT project on Acquia Cloud where memcache is available.
2. Try to do pretty much anything, observe the above error.

Steps to verify the solution:
-----------
1. Install a BLT project with this patch on Acquia Cloud.
2. Observe that the error goes away.
3. Further verify that the settings here match verbatim what's recommended by the memcache module itself.

### DIFF
--- a/composer.suggested.json
+++ b/composer.suggested.json
@@ -13,7 +13,7 @@
     "drupal/cog": "^1.0.0",
     "drupal/devel": "^1.0.0",
     "drupal/qa_accounts": "^1.0.0-alpha1",
-    "drupal/memcache": "^2.0-alpha7",
+    "drupal/memcache": "2.0-alpha7",
     "drupal/seckit": "^1.0.0-alpha2",
     "drupal/security_review": "*",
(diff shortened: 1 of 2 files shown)


ba66e77:
Have you compared this to what @guitarte has in #3075?

danepowell:
I fixed the linting errors (we should probably get the Acquia docs updated to match).

Let's centralize general discussion in #3075.

Issue #3069, opened 2018-09-07
PHPCS linting runs interactively on CI


bobbygryzynger:
My system information:
* Operating system type: Ubuntu
* Operating system version: 14.04
* BLT version: 9.x

When I run `source ${BLT_DIR}/scripts/travis/run_tests` on my CI environment.

I get the following output:
```
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 10 | ERROR | [x] Short array syntax must be used to define arrays
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
Time: 279ms; Memory: 10Mb
[ExecStack]  Exit code 1  Time 0.317s
 Attempt to fix violations automatically via PHPCBF? (y/n) 
```

At which point the tests hang until the build times out.

And I expected this to happen: the build to fail immediately.

dpagini:
Yes, I have seen this before. If I put up a PR with a PHPCS violation, it just sits waiting for user input for the next 10 minutes until it times out. I did not try the patch, however.

Example:
<img width="703" alt="screen shot 2018-09-12 at 2 42 28 pm" src="https://user-images.githubusercontent.com/1724163/45446148-24989b80-b69a-11e8-8059-77c7b44e803e.png">

dpagini:
Ok, since I had the test run up, I pulled in the patch, and it resolves the issue for me. My build failed immediately instead of waiting 10 minutes...

<img width="763" alt="screen shot 2018-09-12 at 2 54 27 pm" src="https://user-images.githubusercontent.com/1724163/45446824-de443c00-b69b-11e8-8713-4bdcea27a3ab.png">

ba66e77:
Great!

The branch has moved forward a bit since this was rolled, so I have it updating the PR. Assuming there's no issue resulting from that, we'll get it merged.

Thank you both!

PR #3090, opened 2018-09-12
DO NOT MERGE

ba66e77:
This is being used only to replicate the problem reported in #3069.


bobbygryzynger:
@ba66e77 I think you'll need to place the failing file in a path typically covered by PHPCS (e.g., `docroot/module/custom`). If I put a file at the root of my project, this isn't picked up by PHPCS

PR #3092, opened 2018-09-12
Use core's PHPUnit bootstrap file.

### LINKED ISSUE #3091: Exemples files that ship with BLT should keep on being managed by BLT
In 9.2.0-alpha2, a PHPUnit breakage happened on Acquia Pipelines due to `./tests/phpunit/ExampleTest.php` which required an updated class name as per the [PHPUnit 6.0.0 changelog](https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-6.0.0).

This file - like other example files - is added to BLT when it's first installed. Since it'll most likely be ignored by end users, it seems like it should be on BLT to keep on updating this file on behalf of customers. Alternatively, we should provide an upgrade path in the form of a warning and how-to for customers to update their files when required.

danepowell:
Fixes #3091 (sort of)
--------
I don't think it makes sense for BLT to attempt to update sample test files that it distributes, because users probably frequently modify those files, and I don't want to get into the business of munging XML files or something.

Rather, I think we should just try to maintain compatibility across PHPUnit versions, the same way that core does, so that users don't _have_ to update those test files when PHPUnit versions change.

Changes proposed:
---------
- By default, use core's phpunit.xml.dist (and therefore core's bootstrap.php as well), which will alias PHPUnit 4 and 6 class names so that they are interoperable
- Remove our own bootstrap.php and phpunit.xml files, which were either empty boilerplate or copies of core's.

Steps to verify the solution:
-----------
1. Use a BLT project running on PHP 5.6 (necessary to test PHPUnit 4). Apply this patch, and remove tests/phpunit/phpunit.xml. 
2. Install PHPUnit 4. Verify that `blt tests:phpunit` runs.
2. Upgrade to PHPUnit 6 and verify that tests still run.


ba66e77:
I've verified this for PHP7/PHPUnit6, but getting PHP5.6 installed is giving me fits.

danepowell:
Tell me about it... getting Composer playing nicely with PHP 5.6 was the most difficult part of this for me as well. Later versions seem to be completely borked, I had to downgrade. https://github.com/composer/composer/issues/7615

PR #3096, opened 2018-09-12
Support memcache on ACSF via site flags.

danepowell:
Changes proposed:
---------
- Turn on memcache for ACSF sites that are "flagged" to run memcache

Steps to replicate the issue:
----------
1. Deploy a BLT site on ACSF
2. Observe that memcache is not enabled for the site, regardless of any configuration for that particular site

Steps to verify the solution:
-----------
1. Deploy a BLT site to ACSF with this patch
2. Verify that memcache is not initially enabled
3. Edit the site node and turn on the memcache flag
4. Verify that memcache is now running for the site

### DIFF
--- a/scripts/factory-hooks/post-settings-php/memcache.php
+++ b/scripts/factory-hooks/post-settings-php/memcache.php
@@ -1,0 +1,18 @@
+<?php
+
+/**
+ * @file
+ * Example implementation of ACSF post-settings-php hook.
+ *
+ * @see https://docs.acquia.com/site-factory/tiers/paas/workflow/hooks
+ */
+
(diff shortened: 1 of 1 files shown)


lcatlett:
This needs to be a pre settings hook because it needs to override config set in settings.php

danepowell:
If it needs to override what's in settings.php, doesn't that need to be a _post_ settings hook? I just copied what was in the previous PR: #2936

lcatlett:
I was referring to the service container settings in memcache.settings.php because it overrides cores default of the database. This has to be set as early as possible so that site installs can use the alternative backend caches

PR #3098, opened 2018-09-13
amotic -> atomic

terrabruder:
Fixes a typo in the code review guide.

Fixes # 
--------
n/a 

Changes proposed:
---------
- fix a typo in code review docs

Steps to verify the solution:
-----------

- Search the docs for "amotic." It should no longer be present.


ba66e77:
That's actually not a typo. We want to see commits that are logically similar to the music of this DJ: https://soundcloud.com/kaup-nomad

Ok, not really. It's a typo. 😄 Thanks for the fix!

As an aside, now that I've stumbled across this DJ, I'm kinda digging it.

Issue #3095, opened 2018-09-12
drupal:config:import on a new project w/ Lightning does not work as expected


justinlevi:
After creating a brand new project, with no configuration exported. I then install drupal and do a config export. The next time I run $ `blt drupal:install` I get the error: 

`Configuration in the database does not match configuration on disk`

This is a brand new install and the prior step was to export the config so this seems like a bug to me. This is also causing pipelines to fail the build artifact as well. 

---

**My system information:**
* Operating system type: Mac
* Operating system version: Latest
* BLT version: 9.1.6

**Output of `blt doctor`:**

<details>

```
+---------------------------+------------------------------------------------+
| Property                  | Value                                          |
+---------------------------+------------------------------------------------+
| %paths.%root              | /var/www/docroot                               |
| %paths.%site              | sites/default                                  |
| %paths.%modules           | sites/all/modules                              |
| %paths.%themes            | sites/all/themes                               |
| %paths.%config-sync       | /var/www/config/default                        |
| %paths.%files             | sites/default/files                            |
| %paths.%temp              | /tmp                                           |
| %paths.%private           | /var/www/files-private                         |
| admin-theme               | seven                                          |
| alias-searchpaths.0       | /var/www/drush/sites                           |
| blt-version               | 9.1.6                                          |
| bootstrap                 | Successful                                     |
| composer-version          | Composer version 1.6.3 2018-01-31 16:28:17     |
| config-sync               | /var/www/config/default                        |
| db-driver                 | mysql                                          |
| db-hostname               | db                                             |
| db-name                   | default                                        |
| db-password               | user                                           |
| db-port                   | 3306                                           |
| db-status                 | Connected                                      |
| db-username               | user                                           |
| drupal-settings-file      | sites/default/settings.php                     |
| drupal-version            | 8.5.7                                          |
| drush-alias-files.0       | /var/www/drush/sites/dev.site.yml              |
| drush-alias-files.1       | /var/www/drush/sites/prod.site.yml             |
| drush-alias-files.2       | /var/www/drush/sites/test.site.yml             |
| drush-cache-directory     | /home/docker/.drush/cache                      |
| drush-conf.0              | /var/www/vendor/drush/drush/drush.yml          |
| drush-conf.1              | /var/www/drush/drush.yml                       |
| drush-conf.2              | /var/www/docroot/sites/default/local.drush.yml |
| drush-script              | /var/www/vendor/bin/drush                      |
| drush-temp                | /tmp                                           |
[...]

PR #3099, opened 2018-09-13
Downgrade config override error to warning.

danepowell:
Changes proposed:
---------
- When BLT detects that configuration is overridden following a configuration import, it currently throws an error and fails the process. Instead, this changes it to a warning and doesn't fail the process.

Steps to replicate the issue:
----------
1. Follow the steps in #3095 (_without_ applying the core patch) to install Drupal, export config, and attempt to install Drupal again to get the error.

Steps to verify the solution:
-----------
1. Follow the same steps, but note that it's now a warning instead of an error.

It's regrettable that we should make this change, since this error is trying to save users from a lot of potential trouble. But it seems like no one really understood what it meant and just ended up blaming BLT, which isn't good for anyone.

### DIFF
--- a/src/Robo/Commands/Setup/ConfigCommand.php
+++ b/src/Robo/Commands/Setup/ConfigCommand.php
@@ -196,7 +196,7 @@ class ConfigCommand extends BltTasks {
    */
   protected function checkConfigOverrides($cm_core_key) {
     if (!$this->getConfigValue('cm.allow-overrides') && !$this->getInspector()->isActiveConfigIdentical()) {
-      throw new BltException("Configuration in the database does not match configuration on disk. BLT has attempted to automatically fix this by re-exporting configuration to disk. Please read https://github.com/acquia/blt/wiki/Configuration-override-test-and-errors");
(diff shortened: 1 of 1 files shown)


danepowell:
I've changed my mind on this. This check and the associated error saved my butt today by catching corrupt configuration updates in several contributed modules. I don't want to lose that coverage. The problem is that if we hadn't noticed these errors right away, we would have had to track them down at a later date and it would have been exponentially more difficult to determine the root cause.

If we want to deflect some of the support requests that this generates, I think we should point people to relevant upstream issues so that they understand that this is simply revealing deeper bugs in Drupal core and contributed modules, and it's not a problem with BLT.

danepowell:
Saved our butts yet again today.

- A new display mode was added for an existing content type
- At the same time on a parallel feature branch, someone added a new field to that content type
- Display modes contain a list of all of the fields for a content type, sorted into display regions. So these PRs essentially conflict, but you wouldn't notice that until one gets merged and you update the other.

The resolution (what running `blt setup` followed by `drush cex` will do) is to simply update the display mode to contain the new field. But this never would have been caught without this test (and without using Github's protected branches and forcing PRs to be up to date before merging)

PR #3107, opened 2018-09-17
Resolves #3106: Add 'skip_permissions_hardening' setting to local settings file template

### LINKED ISSUE #3106: Update default.local.settings.php to tell Drupal to skip permissions hardening
It's very frustrating to have Drupal "harden" the sites/default directory and settings.php file in a local dev environment by removing write access.

[Drupal 8 includes a way to disable this by a setting](https://www.drupal.org/project/drupal/issues/1232572), and even [includes it in their example local settings file](https://github.com/drupal/drupal/blob/8.6.x/sites/example.settings.local.php)

But BLT doesn't use that example file. Instead if provides its own template, which is missing that line.

bkosborne:
Fixes #3106 
--------

Changes proposed:
---------
Add some code to the local settings file template that BLT uses when generating the real local settings file.

### DIFF
--- a/settings/default.local.settings.php
+++ b/settings/default.local.settings.php
@@ -131,6 +131,18 @@ $settings['extension_discovery_scan_tests'] = FALSE;
 $settings['rebuild_access'] = FALSE;
 
 /**
+ * Skip file system permissions hardening.
+ *
+ * The system module will periodically check the permissions of your site's
+ * site directory to ensure that it is not writable by the website user. For
+ * sites that are managed with a version control system, this can cause problems
+ * when files in that directory such as settings.php are updated, because the
+ * user pulling in the changes won't have permissions to modify files in the
+ * directory.
+ */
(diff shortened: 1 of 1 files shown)


danepowell:
This would be awesome if it works... it's super annoying to constantly have to change directory permissions when checking out different Git branches locally.

But is it definitely Drupal doing this? I know that BLT also sets permissions at various points. We should make sure that this is in line with what BLT is also doing.

bkosborne:
@danepowell 644 permissions would actually be OK, because owner still has write perms in that case. But core is setting it to 444 for files. I don't have steps to repro at the moment and am going away for a week but I've added this to my TODO list to come up with repo steps.

danepowell:
Actually 644 would be really bad for directories, because you can't do much of anything without execute permissions on a directory (certainly not listing contents or deleting files.) In this case though, the code is a little unclear: in fact it's setting _files_ to 644, which is fine.

I'll try to repro this as well.

PR #3108, opened 2018-09-17
Fixes 2980: Clear Twig caches on deploys patch.

greylabel:
Fixes # 2980
--------

Changes proposed:
---------

- Remove _Clear Twig caches on deploys_ patch.
- Update Drupal Core to ^8.6.0
- Update Lightning to ^3.2.0

Steps to replicate the issue:
----------

1. Run `composer update` on a BLT 8.9.x project when specifying `"drupal/core": "^8.6.1",` in the project `composer.json`

Result:
```
Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2752961-90.patch


  [Exception]
  Cannot apply patch Clear Twig caches on deploys (https://www.drupal.org/files/issues/2752961-90.patch)!
```

Steps to verify the solution:
-----------

1. Run `composer update` on a BLT 8.9.x project when specifying `"drupal/core": "^8.6.1",` in the project `composer.json`

Result: above error is not present.

### DIFF
--- a/composer.required.json
+++ b/composer.required.json
@@ -6,7 +6,7 @@
     }
   },
   "require": {
-    "drupal/core": "^8.0",
+    "drupal/core": "^8.6.0",
     "drupal/features": "^3.0.0",
     "drupal/config_split": "^1.0.0-beta4",
     "drupal/devel": "^1.0.0-alpha1"
@@ -43,12 +43,7 @@
         "sites/default/default.settings.php": "sites/default/settings.php"
       }
     },
-    "enable-patching": true,
-    "patches": {
-      "drupal/core": {
-        "Clear Twig caches on deploys": "https://www.drupal.org/files/issues/2752961-90.patch"
-      }
-    }
+    "enable-patching": true
   },
   "scripts": {
     "blt-alias": "blt install-alias -y --ansi",
--- a/composer.suggested.json
+++ b/composer.suggested.json
@@ -6,7 +6,7 @@
     }
   },
   "require": {
-    "acquia/lightning": "^2.2.7 | ^3.0.0",
+    "acquia/lightning": "^2.2.7 | ^3.2.0",
     "drupal/acquia_connector": "^1.5.0",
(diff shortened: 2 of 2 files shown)


greylabel:
Travis failure looks unrelated. Maybe restart the build?

greylabel:
@danepowell Good point. I overlooked that. We should probably figure out the best course of action to get BLT 8.9.x compatible with the latest Drupal core to maintain a security update path or deprecate it officially.

danepowell:
I was thinking BLT should follow Drush's support status, but Drush seems a bit waffly on this. It appears that Drush 8 is still a supported release, but officially Drush 8 only supports Drupal 8.3, which is wildly EOL, so why is Drush 8 still supported at all? 😄

greylabel:
@danepowell Drupal 8 is supported for Drupal 7, I think that's probably the reason.

hctom:
I also have the same problem with the patch while trying to update my site to Lightning 3.2.0. But I am using BLT 9.1.7 and therefore this pull request/patch does not apply. Was there any reason to make these changes based on 8.9.x instead of the current 9.x main branch?

danepowell:
Lightning 3.2.0 requires Drupal 8.6, which is only supported by BLT 9.2.x. Please use BLT 9.2.0-alpha3 or later with Drupal 8.6.

hctom:
@danepowell Thanx for clarifying this. After update to 9.2.x everything works like a charm again

danepowell:
I'm afraid we can't bump the core version like this. BLT 8.9.x is EOL and officially doesn't support Drupal 8.6 (nor does any of its dependencies, including Drush 8 and Lightning). This would require a pretty big shift in policy as well as probably be a breaking change for users of 8.9.x who (for better or worse) are using older versions of Drupal core.

If folks need to use Drupal 8.6, they should update to BLT 9.2.x, Drush 9, and Lightning 3.2.0.

PR #3113, opened 2018-09-18
Clarify branch support policies

danepowell:
In responding to a comment in #3108, I realized that we don't have any guidance in the README about what version of BLT to use with Drupal 8.6.

Should we also set the default branch in Github to be 9.2.x?


danepowell:
@ba66e77 maybe let's discuss

Issue #3114, opened 2018-09-18
BLT Drush Aliases recipe support for Acquia Cloud Professional


sebinievas:
My system information:
* Operating system type: OSX
* Operating system version: 10.13.6 (17G65)
* BLT version: 9.2.0-alpha3

Output of `blt doctor`:
```
+---------------------------+----------------------------------------------+
| Property                  | Value                                        |
+---------------------------+----------------------------------------------+
| %paths.%root              | /app/docroot                                 |
| %paths.%site              | sites/default                                |
| %paths.%modules           | sites/all/modules                            |
| %paths.%themes            | sites/all/themes                             |
| %paths.%config-sync       | /app/config/default                          |
| %paths.%files             | sites/default/files                          |
| %paths.%temp              | /tmp                                         |
| %paths.%private           | /app/files-private                           |
| admin-theme               | seven                                        |
| alias-searchpaths.0       | /app/drush/sites                             |
| blt-version               | 9.2.0-alpha3                                 |
| bootstrap                 | Successful                                   |
| composer-version          | Composer version 1.7.2 2018-08-16 16:57:12   |
| config-sync               | /app/config/default                          |
| db-driver                 | mysql                                        |
| db-hostname               | database                                     |
| db-name                   | drupal                                       |
| db-password               | drupal                                       |
| db-port                   | 3306                                         |
| db-status                 | Connected                                    |
| db-username               | drupal                                       |
| drupal-settings-file      | sites/default/settings.php                   |
| drupal-version            | 8.6.1                                        |
| drush-cache-directory     | /var/www/.drush/cache                        |
| drush-conf.0              | /app/vendor/drush/drush/drush.yml            |
| drush-conf.1              | /app/drush/drush.yml                         |
| drush-conf.2              | /app/docroot/sites/default/local.drush.yml   |
| drush-script              | /app/vendor/bin/drush                        |
| drush-temp                | /tmp                                         |
| drush-version             | 9.4.0                                        |
| files                     | sites/default/files                          |
| install-profile           | lightning                                    |
| modules                   | sites/all/modules                            |
| php-bin                   | /usr/local/bin/php                           |
| php-conf.1                | false                                        |
| php-os                    | Linux                                        |
| private                   | /app/files-private                           |
| root                      | /app/docroot                                 |
| site                      | sites/default                                |
| stacks.drupal-vm.inited   | false                                        |
| stacks.dev-desktop.inited | false                                        |
| temp                      | /tmp                                         |
[...]

PR #3115, opened 2018-09-19
Fix drush alias generator for acp servers (#3114)

sebinievas:
Fixes # 
--------
#3114 

Changes proposed:
---------
This more than likely needs work. Currently we simply add an additional condition to check for `acp` hosting vs only `ace` and `acsf`.

Steps to replicate the issue:
----------
1. See #3114 for details

Steps to verify the solution:
-----------
1. After running `blt recipes:aliases:init:acquia` and `drush sa` you should see drush aliases.
1. Verify they were installed correctly in the `drush/sites` directory of your Drupal project.

### DIFF
--- a/src/Robo/Commands/Generate/AliasesCommand.php
+++ b/src/Robo/Commands/Generate/AliasesCommand.php
@@ -247,7 +247,8 @@ class AliasesCommand extends BltTasks {
       $remoteHost = $ssh_split[1];
       $remoteUser = $ssh_split[0];
 
-      if ($hosting == 'ace' || $hosting == 'acp') {
+      // Get aliases for non `acsf` hosting environments
+      if ($hosting != 'acsf') {
 
         $siteID = $site_split[1];
         $uri = $env->domains[0];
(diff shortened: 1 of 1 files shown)


src/Robo/Commands/Generate/AliasesCommand.php
```diff
@@ -247,7 +247,7 @@ protected function getSiteAliases($site) {
       $remoteHost = $ssh_split[1];
       $remoteUser = $ssh_split[0];
 
-      if ($hosting == 'ace') {
+      if ($hosting == 'ace' || $hosting == 'acp') {
```
lcatlett: Instead of limiting this to only ACE or ACP, generate a standard drush alias for every valid hosting type returned except for ACSF - the easiest way to do this is probably `!='acsf'`. Since valid hosting types that could be returned currently are 'free', 'acp', 'ace', 'acsf', 'network', and 'search' and will presumably change as Cloud API v2 gets closer to GA, this will allow for a bit more flexibility

sebinievas: Thank you @lcatlett ! I've made the adjustment. I was tempted to add an `else` to the next condition block, but I think leaving it more verbose is better in this case. Let me know if you disagree and I'll gladly update it.

grasmash: @lcatlett Is this change acceptable?

Elaine-dev:
Does this work in the AC free accunt?

PR #3121, opened 2018-09-22
Improve Composer Performance

lcatlett:
Changes proposed:
---------

- Add zaporylie/composer-drupal-optimizations package to reduce performance overhead of symfony updates due to composer iterating over all tags. The benchmarks below show a 10 second improvement for composer update and significantly lowered memory consumption. 

Benchmarks
----------
Below are benchmarks on a vanilla BLT project

```
~/sites/bltacsf/blted8 master*
❯ composer update nothing --profile
[379.6MB/16.45s] Memory usage: 379.63MB (peak: 1266.55MB), time: 16.45s

~/sites/bltacsf/blted8 master* 18s
❯ composer require zaporylie/composer-drupal-optimizations
Writing lock file
Generating autoload files

~/sites/bltacsf/blted8 master* 20s
❯ composer update nothing --profile
[282.5MB/6.69s] Memory usage: 282.46MB (peak: 386.61MB), time: 6.69s
```


danepowell:
I verified the performance gain over a few runs, just to rule out caches or anything fluky. Without the new package:
> Memory usage: 372.99MB (peak: 1187.91MB), time: 16.78s
Memory usage: 372.99MB (peak: 1187.91MB), time: 15.59s
Memory usage: 372.23MB (peak: 1187.15MB), time: 15.12s

With the new package:
> Memory usage: 276.39MB (peak: 367.09MB), time: 7.22s
Memory usage: 276.23MB (peak: 366.92MB), time: 5.8s
Memory usage: 276.23MB (peak: 366.92MB), time: 6.31s

That's pretty awesome. Of course there ain't no such thing as a free lunch, so the question is... what's the downside? Does this just mean that users can no longer install "legacy" Symfony versions? As long as that's the only major change, I think this would be a win.

aweingarten:
@danepowell, what is the CPU usage.  I wonder if the composer CPU goes down enough to allow for any parallelization of composer + node?

danepowell:
@aweingarten I checked, CPU usage is basically zero regardless of whether this package is used or not. That jives with my previous experience with composer: the performance bottleneck is disk and network i/o latency, CPU has little or nothing to do with it.

lcatlett:
benchmarks for oob blt 10.0.x are below.

~/sites/blt/blted8 master*
❯ composer update nothing --profile
[388.6MB/19.47s] Memory usage: 388.56MB (peak: 1298.94MB), time: 19.47s

~/sites/blt/blted8 master*
❯ composer require zaporylie/composer-drupal-optimizations
Gathering patches for dependencies. This might take a minute.
  - Installing zaporylie/composer-drupal-optimizations (1.0.2): Loading from cache
Writing lock file
Generating autoload files

~/sites/blt/blted8 master* 22s
❯ composer update nothing --profile
[291.8MB/9.33s] Memory usage: 291.84MB (peak: 401.27MB), time: 9.33s

 As far as downsides, this approach will not work for older versions of core or other dependencies that require legacy versions of symfony.

lcatlett:
Coming back to this, it looks like a few projects are using `zaporylie/composer-drupal-optimizations` and seeing dramatic performance improvements. I will open a new PR with updated benchmarks since this one is so out of date. Symfony flex has a similar approach but we arent really leveraging any recipes so it seems like unneeded overhead

geerlingguy:
@lcatlett - Awesome, thanks! I was just testing this on some non-BLT projects and found that it increased performance for `composer require` and reduced memory usage, both by half or more! And with PHP 5.6 (for the brave few still running it), it's even more dramatic.

geerlingguy:
To put a pin in this issue... it looks like @grasmash filed a new PR with just the addition of this library 15 days ago: https://github.com/acquia/blt/pull/3287

geerlingguy:
(However, that's only in 10.x)

Issue #3116, opened 2018-09-19
Establish best practice and support for testing and setup strategies


danepowell:
Based on issues like #3109, it's clear that people are looking for guidance and additional functionality around how to run automated tests specifically in regards to setup strategy.

My opinion is that BLT should default to using the `install` strategy out of the box in CI environments, but that long term all sites should use one of two approaches:
- Single site installs should use the `sync` strategy, which ensures that database updates get tested against live content
- Multisite / ACSF installs should use a duel strategy: `sync` or `import` first in order to test database updates against sites with live content, then `install` to ensure that _new_ sites can be installed in the future.

I think we should try to align on a recommended approach, document it, and then implement any necessary changes to support it. Note that the second approach is currently unintuitive to configure (you basically have to hack the `sync.commands` variable), I think this could be supported more cleanly.

lpeabody:
> (you basically have to hack the sync.commands variable)
Can you elaborate on this statement? In what circumstances would altering sync.commands be needed?

lpeabody:
I think what would be helpful to a lot of folks is, in addition to getting these two items well documented, that the configuration structure in its entirety be documented. For every configuration item, document what commands use it and why they use it.

I find myself digging through the source code and finding configuration in build.yml that is referenced in the source, experimenting with overriding it in various levels of blt.yml files, and see if I'm happy with the resulting execution. I shouldn't be having to dig through the source as much as I am, I should be referencing a set of documentation with cleanly defined use cases and explanations of every BLT command and configuration item.

So my asks are:

1. Document every single command. Explain what they do and which additional commands are invoked by issuing that command.
2. Document all of the configuration items (and possible values) BLT consumes. Explain the purpose of the `setup` configuration and why it exists, and explain what each nested configuration item does (e.g. `setup.strategy` defines which supported setup strategy should be executed in the context of a given site).
3. Keep the configuration overrides documentation front-and-center.

Every complaint or gripe I've heard about adopting BLT has come from people not knowing how to intuitively get started with it or adapt it to suit their current project needs.
They don't know when they should be deferring to BLT to do something or when to simply use Drush, or they don't understand the configuration override system or the notion of site contexts, etc... People just have a lot of questions and they're having a hard time getting answers.

[...]

PR #3126, opened 2018-09-25
#3116: Support dual setup strategies.

danepowell:
Relates to #3116
--------

Currently there are only three strategies to "setup" a site: installing from scratch, syncing a db from a remote environment, or importing a static SQL dump. The problem is that in CI environments you might want to use multiple strategies for testing purposes.

For instance, on ACSF platforms, you want to ensure both that new sites can be installed from scratch, and that existing sites can be updated without error. This requires a combination of the "install" and "sync" strategies.

Changes proposed:
---------
- Add a "dual" setup strategy that essentially just runs a sync followed by an install. In my experience, this is one of the most common testing strategies for multisite installs (although by no means the only one). Long term, we probably want to have a discussion in #3116 about whether to be opinionated about this, or allow users to "mix and match" strategies.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -137,7 +137,12 @@ project:
     uri: ${project.local.protocol}://${project.local.hostname}
 
 setup:
-  # Valid values are install, sync, import.
+  # Valid values are install, sync, import, dual.
+  # - install: installs a site using `drush si`.
+  # - sync: syncs a database from `drush.aliases.remote` and runs updates.
+  # - import: imports a static database dump.
+  # - dual: runs a sync to ensure that database updates are applied correctly
+  #   followed by a fresh install for tests.
   strategy: install
   # If setup.strategy is import, this file will be imported. File path is
   # relative to drupal docroot directory.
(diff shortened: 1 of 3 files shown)


lcatlett:
I'm cautious to recommend this new proposed type of setup strategy in BLT as an Acquia best practice since running import or a sync command followed by a site install is not the same install / update process that occurs on the upstream production environment, especially on ACSF.

While I am very much in favor of running tests against live content in addition to a new install as part of every CI build, I'm not sure I follow the logic behind adding an additional setup strategy to BLT rather than setting up two test scenarios in the CI config which mirror the end-to-end build / deploy/ install / update tasks of the production environment.

It seems like users of BLT would benefit more from improved docs and maybe sample CI scripts about how to use existing BLT config to accomplish common scenarios such as running tests against a demo site, syncing and sanitizing production data, running core or contrib module tests, parallelizing builds, or integrating 3rd additional tools or production-like components like memcache or Varnish within CI builds

danepowell:
Good point. I'll be the first to admit that this "dual" setup strategy is a kind of hack, although I don't think it should hurt anything. Running a sync+update followed by an install isn't supposed to replicate a process, it's just a process unto itself to work around the fact that some CI tools can't run tasks in parallel.

Anyway, I definitely prefer the idea of doing this at the CI level. With Travis, that's easy using the "build matrix". Unfortunately, Pipelines doesn't support that, so you'd just have to use two successive build steps, which would essentially replicate this behavior. Still a better approach I think.

Let me see if I can come up with a CI-centric approach.

PR #3140, opened 2018-10-02
Remove PHP 5.6 support

grasmash:
This reverts commit f43c708df2242b4e38a2432f4b8077119dec7898.

### DIFF
--- a/src/Robo/Hooks/WebServerHook.php
+++ b/src/Robo/Hooks/WebServerHook.php
@@ -3,23 +3,12 @@
 namespace Acquia\Blt\Robo\Hooks;
 
 use Acquia\Blt\Robo\BltTasks;
-use Acquia\Blt\Robo\Common\IO;
-use Acquia\Blt\Robo\Config\ConfigAwareTrait;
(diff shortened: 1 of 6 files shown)


src/Robo/Hooks/WebServerHook.php
```diff
@@ -3,13 +3,24 @@
 namespace Acquia\Blt\Robo\Hooks;
 
 use Acquia\Blt\Robo\BltTasks;
+use Acquia\Blt\Robo\Common\IO;
+use Acquia\Blt\Robo\Config\ConfigAwareTrait;
+use Acquia\Blt\Robo\Inspector\InspectorAwareTrait;
 use Consolidation\AnnotatedCommand\CommandData;
+use League\Container\ContainerAwareTrait;
```
danepowell: I don't understand where all of these includes are coming from. It doesn't seem like they are necessary.

grasmash: If you don't add these, a bunch of PHP warning are thrown.

greylabel:
+100 to this

PR #3145, opened 2018-10-04
Remove PHP 5.6 from the list of tested versions. (#3137)

danepowell:
Backport of #3137


danepowell:
Might need to backport #3129 first.

PR #3146, opened 2018-10-05
Allows Drush ANSI output to be disabled.

bobbygryzynger:
Changes proposed:
---------
- Allow Drush's ANSI output to be disabled.

On environments which don't support ANSI output (looking at you Acquia Cloud), the output generated by Drush calls dispatched through BLT can be pretty ugly. For example, 
```
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/html/<env>/vendor/bin/drush @self cache-rebuild --no-interaction -v --ansi in /var/www/html/<env>/docroot
 [37;46;1m[success][39;49;22m Cache rebuild complete.
```

When the logs are full of this it's pretty distracting. Currently there is no way to disable this, because the `--ansi` flag is passed during every Drush call. This PR allows that to be configured.

Steps to replicate the issue:
----------
1.  Deploy to an environment that doesn't support ANSI output and witness the results.

Steps to verify the solution:
-----------
1. Apply this patch.
2. In `blt.yml` set `drush.ansi` to `false`.
3. Deploy and confirm the output no longer contains ANSI sequences.


mikemadison13:
It might make sense to put this in the same namespace as our other disable code (although, I'm unclear on if that might cause Robo problems).

Any thoughts @grasmash  on if it would be safe to do:

```
disable-tagets
  drush:
    ansi: true
```

in which case, we would also need to flip the check on the config to look for it to be set to true instead of false.

dpagini:
Isn't `disable-targets` meant more for the commands you can call from BLT? This would be turning off the drush option, right? What about a subkey of `drush`?

mikemadison13:
@dpagini yeah i really think that's what it's intended for, i agree, i'm just trying to figure out "where" we could have common disables aren't directly hooks... i don't love the idea of this that and the other thing all having their own separate disabled space.

dlopez-akalam:
Another solution is to allow disabling ansi per command. Here is a Pull Request: #3269

mikemadison13:
@bobbygryzynger can you please update this PR to take into consideration what @dlopez-akalam did in https://github.com/acquia/blt/pull/3269? I think we want to support both of these, but we merged his first since it seemed to have less impact on the codebase.

danepowell:
I reconciled this with #3269. I think the configuration is fine as is. Looking forward to actually getting to use this on Cloud, thanks for the PR!

danepowell:
Follow-up to actually disable ANSI in Cloud: #3426

PR #3151, opened 2018-10-08
Fixed travis output.

danepowell:
I'm trying to figure out why Travis logs look like this:
screenshot-travis-ci org-2018 10 08-13-59-42

My theory is that the problem is that this script is not properly cleaning up after itself by disabling `-v` (print shell input lines)

### DIFF
--- a/scripts/travis/exit_early
+++ b/scripts/travis/exit_early
@@ -10,3 +10,5 @@ then
     exit 0
   fi
 fi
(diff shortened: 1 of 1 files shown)


danepowell:
Looks like it worked, you can actually see what commands are being run now
screenshot-travis-ci org-2018 10 08-14-08-30

PR #3152, opened 2018-10-08
Travis shouldn't duplicate Drupal 8.6 tests

danepowell:
Currently, Travis is testing Drupal 8.6 in 4 variations, with tests not allowed to fail: each combination of PHP 7.1/7.2 and short/long tests. That's fine and dandy.

It then tests Drupal 8.6 in PHP 7.1/7.2 allowing failures. This is obviously redundant... I think in this case it should be testing Drupal 8.7 instead.

To verify this fix: verify that Travis tests Drupal 8.6 in the first four test variations, not allowing failures. It then tests 8.7, which may or may not fail.


danepowell:
Actually it's impossible to test Drupal 8.7 until Lightning also supports it, so we might as well just disable those tests for now and free up some test cycles.

PR #3153, opened 2018-10-08
Clean up PHPUnit directories and namespaces.

danepowell:
If you take a look at the tests in tests/phpunit/Blt and tests/phpunit/BltProject, you'll see there's a lot of inconsistency in how they are namespaced. Some of the BltProject tests are namespaced as BltProject, others as Blt.

According to @grasmash , this is an anachronism and they should all now be in the Blt namespace. This fixes that and organizes the directories accordingly.

Verify that this has no functional impact by checking the PHPUnit ouput before and after (locally or [on Travis](https://travis-ci.org/acquia/blt/jobs/438869421#L9323)), it should be the same:
> OK, but incomplete, skipped, or risky tests!
Tests: 62, Assertions: 197, Risky: 11.


mikemadison13:
I see the same results with this code in Travis:

```
OK, but incomplete, skipped, or risky tests!
Tests: 62, Assertions: 197, Risky: 11.
```

PR #3155, opened 2018-10-09
10.0.x: Upgrade to Drupal 8.6.1

hugovk:
Changes proposed:
---------

- Upgrade to Drupal 8.6.1
- Was released on 2018-09-10
- https://www.drupal.org/project/drupal/releases/8.6.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "drupal/coder": "^8.2.11",
         "drush/drush": "^9.4.0",
         "drupal/config_split": "^1.0.0",
-        "drupal/core": "^8.6.0",
+        "drupal/core": "^8.6.1",
         "drupal/features": "^3.8.0",
         "drupal/memcache": "^2.0-alpha5",
         "drupal/seckit": "^1.0.0-alpha2",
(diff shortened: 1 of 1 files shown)


danepowell:
@hugovk is there some reason that this upgrade is required for BLT users?

Note that the way we have defined the Composer dependency, anyone can easily upgrade to 8.6.1 without this change. The only effect of this PR will be to _mandate_ people use 8.6.1, which I don't think we should do unless there's a compelling reason.

hugovk:
Oops, I misinterpreted the `^` operator, and `^8.6.0` will indeed allow upgrade to 8.6.1 as well. I'll close this, sorry for the noise!

By the way, is there a schedule for the release of 9.2.0 from alpha? We're looking forward to upgrading from 8.5 to 8.6. Thanks!

danepowell:
No worries, thanks for trying to contribute.

I'm not aware of anything blocking a new 9.2 release, let me take a look at that today and see what makes sense.

danepowell:
Good news, 9.2.0 is out now!

hugovk:
Brilliant, thank you!

PR #3164, opened 2018-10-10
Fixes #3116: Suggest testing db updates.

### LINKED ISSUE #3116: Establish best practice and support for testing and setup strategies

danepowell:
Fixes #3116
--------

Changes proposed:
---------
Just adding some suggested CI commands for folks who want to test db updates and config imports on live content, before installing the site from scratch and running the normal test suite.

I prefer this approach since it's a much lighter-touch solution than #3126. It also opens the door to running these tests in parallel.

### DIFF
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -24,6 +24,11 @@ events:
             type: script
             script:
               - source ${BLT_DIR}/scripts/pipelines/validate
+        # Uncomment these lines to test database updates using live content.
+        # - test-updates:
+        #     type: script
+        #     script:
+        #       - blt drupal:sync:default:site
         - setup-app:
             type: script
             script:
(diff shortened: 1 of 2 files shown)


danepowell:
Note that Pipelines doesn't have access to Cloud dbs out of the box, you'd need to add a "machine SSH key" to the Pipelines container that has access to the Cloud db you want to sync.

PR #3166, opened 2018-10-11
Prevent cache collisions in multisite db-update tasks

### LINKED ISSUE #3158: Refactor PR 3006 to Unblock Deploys/Updates
This is a parent issue to track progress of breaking a previously unmerged PR #3006 into smaller PRs to facilitate faster review. Proposed breakdown: 

- [x] Create script that creates and sets unique drush cache directory per site process for use by long running drush tasks
- [x] Refactor DRUSH_PATHS_CACHE_DIRECTORY approach
- [x] Enhance post-install factory hook
- [x] Enhance db-update factory hook
- [x] Remove deprecated global variables in favor of requests in hooks and settings
- [x] Update Acquia host forwarding and site path settings.php config in favor of requests

lcatlett:
Fixes #3158 subtasks
--------

Changes proposed:
---------

- Adds utility script for generating unique drush cache directory per site / environment 
- Creates and sets unique drush cache directory per site process for db update tasks on ACSF
- Removes redundant cache clears resetting site context
- Adds BLT update hook

Steps to replicate the issue:
----------
the original issue can be replicated by running a site update on ACSF. When the db-update Factory hook is executed, updates and config impport  will fail because the site context is being reset by cache race conditions:

```
[ded-000|10677] - The db-update.sh script has completed: Completed in 27 seconds. Exit code: 0. Standard out: Found installation profile profile1.
sandboxacsf.01live: Running BLT deploy tasks on commercialus.sandboxacsf.acsitefactory.com domain in 01live environemnt on the sandboxacsf subscription.
> drupal:config:import
Executing pre-config-import target hook...
No pre-config-import configured.
 [37;46;1m[info][39;49;22m Executing: mysql --defaults-file=/mnt/tmp/sandboxacsf01live/drush_mmFoqe --database=sandboxacsdb216364 --host=dbmaster-1028 --port=3306 --silent  < /mnt/tmp/sandboxacsf01live/drush_8GA1Hh > /dev/null
 [37;46;1m[info][39;49;22m Executing: mysql --defaults-file=/mnt/tmp/sandboxacsf01live/drush_zNbUU7 --database=sandboxacsdb216364 --host=dbmaster-1028 --port=3306 --silent  < /mnt/tmp/sandboxacsf01live/drush_0NVDab > /dev/null
 [37;46;1m[success][39;49;22m 'drush' cache was cleared.

[33mIn EntityTypeManager.php line 133:[39m
[37;41m                                                               [39;49m
[37;41m  [Drupal\Component\Plugin\Exception\PluginNotFoundException]  [39;49m
[37;41m  The "" entity type does not exist.                           [39;49m
[37;41m                        ...nt/www/html/sandboxacsf01live/docroot
[ExecStack] Done in 0.002s
[Acquia\Blt\Robo\Tasks\DrushTask] Running /mnt/www/html/sandboxacsf01live/vendor/bin/drush @self cc drush --uri=commercialus-site.sandboxdigital.net --no-interaction -v --ansi in /mnt/www/html/sandboxacsf01live/docroot
[Acquia\Blt\Robo\Tasks\DrushTask] Done in 0.767s
[Acquia\Blt\Robo\Tasks\DrushTask] Running /mnt/www/html/sandboxacsf01live/vendor/bin/drush @self cache-rebuild --uri=commercialus-site.sandboxdigital.net --no-interaction -v --ansi in /mnt/www/html/sandboxacsf01live/docroot
[Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1  Time 8.993s
[error]  Failed to import configuration! 
[error]  Command `drupal:config:import ` exited with code 1. 
 [error]  Drupal\Core\Entity\EntityStorageException: Exception thrown while performing a schema update. SQLSTATE[22004]: Null value not allowed: 1138 Invalid use of NULL value: ALTER TABLE {taxonomy_term_field_data} CHANGE `status` `status` TINYINT NOT NULL; Array
(
)
 in Drupal\Core\Entity\Sql\SqlContentEntityStorage->wrapSchemaException() (line 1481 of /mnt/www/html/sandboxacsf01live/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php). 
 [success] Cache rebuild complete.
 [success] Finished performing updates.
```
Steps to verify the solution:
-----------
1. Deploy PR to ACSF environment
2. Run site update at admin/gardens/site-update/update, selecting the branch or tag of this PR
2. Validate that db-update hook in site update executes without errors. To test via the command line as if the task was run by the factory, run `/mnt/www/html/sandbox.01live/factory-hooks/db-update/db-update.sh' 'sandbox' '01live' '[db_role]' [uri]'`, substituting your site, environment, database role, and uri as needed.

### DIFF
--- a/scripts/blt/drush/cache.php
+++ b/scripts/blt/drush/cache.php
@@ -1,0 +1,29 @@
+<?php
+
+/**
+ * @file
+ * Creates a unique temporary cache directory given a site, env, and uri.
+ * This is used for every blt command invocation on site install and update.
+ */
+
+$site = $argv[1];
+$env = $argv[2];
+$uri = $argv[3];
+
+
+if (empty($argv[3])) {
+  echo "Error: Not enough arguments. Site, environment, and uri are required.\n";
+  exit(1);
+}
+
+// Create a temporary cache directory for this drush process only.
+$cache_directory = sprintf('/mnt/tmp/%s.%s/drush_tmp_cache/%s', $site, $env, md5($uri));
+shell_exec(sprintf('mkdir -p %s', escapeshellarg($cache_directory)));
+
+if (!file_exists($cache_directory)) {
+  syslog(LOG_ERR, sprintf('Drush updates could not be executed, as the required cache directory [%s] is missing.', $cache_directory));
+  die('Missing or corrupted drush cache for this process.');
+}
+else {
+  echo "$cache_directory";
+}
--- a/scripts/factory-hooks/db-update/db-update.sh
+++ b/scripts/factory-hooks/db-update/db-update.sh
@@ -18,21 +18,27 @@ db_role="$3"
 domain="$4"
 
 # BLT executable:
-blt="/var/www/html/$site.$env/vendor/acquia/blt/bin/blt"
+blt="/mnt/www/html/$site.$env/vendor/acquia/blt/bin/blt"
 
 # You need the URI of the site factory website in order for drush to target that
 # site. Without it, the drush command will fail. Use the uri.php file provided by the acsf module to
 # locate the URI based on the site, environment and db role arguments.
 uri=`/usr/bin/env php /mnt/www/html/$site.$env/hooks/acquia/uri.php $site $env $db_role`
 
-# Print a statement to the cloud log.
-echo "$site.$target_env: Running BLT deploy tasks on $uri domain in $env environment on the $site subscription."
-
+# Create array with site name fragments from ACSF uri. 
 IFS='.' read -a name <<< "${uri}"
 
-# Set Drush cache to local ephemeral storage to avoid race conditions. This is
-# done on a per site basis to completely avoid race conditions.
-# @see https://github.com/acquia/blt/pull/2922
-export DRUSH_PATHS_CACHE_DIRECTORY=/tmp/.drush/${db_role}
+# Create and set Drush cache to unique local temporary storage per site.
+# This approach isolates drush processes to completely avoid race conditions
+# that persist after initial attempts at addressing in BLT: https://github.com/acquia/blt/pull/2922
 
-$blt drupal:update --environment=$env --site=${name[0]} --define drush.uri=$domain --verbose --yes
+cacheDir=`/usr/bin/env php /mnt/www/html/$site.$env/vendor/acquia/blt/scripts/blt/drush/cache.php $site $env $uri`
+
+# Print to cloud task log.
+echo "Generated temporary drush cache directory: $cacheDir."
+
+# Print to cloud task log.
+echo "Running BLT deploy tasks on $uri domain in $env environment on the $site subscription."
+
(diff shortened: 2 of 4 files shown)


lcatlett:
@lpeabody this is the first refactored PR to replace #3006 and you can see the status of the remaining tasks in #3158 and I will have the additional PRs available to test as soon as possible.

seanhamlin:
I have a PR up for our shared ACSF customer with @lpeabody. Will hopefully get a test release out today to test the BLT 9.2.0 upgrade + this PR patch.

ba66e77:
@lcatlett, I filed https://github.com/lcatlett/blt/pull/11 against your branch with a couple minor tweaks. Those aside, I'm not seeing areas of concern with the code here, but I'm still testing things out.

A couple notes:
1. I do not have a setup which will allow me to fully test this out on ACSF. I'm taking on faith that you and @seanhamlin have done that testing.
2. This does not address non-ACSF multisite installations. 
3. I'm not sure about the changes to `src/Robo/Commands/Setup/ConfigCommand.php`. We've beat that question back and forth a few times now and even though they _should not_ be needed we've seen replicable cases where removing them triggered problems. That's what I'm testing at the moment.

lcatlett:
@ba66e77 thanks for the PR - I just merged. The approach in this specific PR was not intended to work on standard multisite since it uses Factory Hooks and ACE uses Cloud Hooks, so I can create an issue to implemement similar functionality on ACE regarding the `DRUSH_PATHS_CACHE_DIRECTORY` config. The alternative would be to configure this value for every BLT drush command, but that seems like overkill. 

Regarding the ConfigCommand.php changes, I realize those changes may seem substantial, however at the least, the `drush cc drush` is redundant and what is causing a lot of challenges since it is already run by the parent db-update Factory Hook process prior to the custom dp-update.sh firing, though I am not sure if this is the case with Cloud Hooks on ACE. Additionally, `drush cc drush` does not pass the site, root, and uri contexts which negates the benefits derived from setting `DRUSH_PATHS_CACHE_DIRECTORY` in the parent process.

Regarding the `drush cr`, this was likely added to ConfigCommand when we didn't know a lot about the internals of deploys in D8. A heavy-handed cache flush or `cc all` was a common step in D7 deploys because we did not yet have the ability to cache the service container or plugin discovery mechanisms, and were forced to use config management such as Features which bypassed core caching. In looking at past issues related to deploying BLT config and code updates, it looks like most would not actually be resolved using current drupal core or current BLT config.

I would argue that we should actually be encouraging developers to enable caching locally since even when caching is turned off, cache flushes are still required to refresh plugins, config, and classes. Using a full cr to accomplish this requires a full bootstrap of Drupal locally which makes development unnecessarily slow without a lot of benefit. Additionally, on the remote upstream environment where caching is presumaly enabled, a cr does not actually refresh those particular cache objects in most cases unless you are using an alternative cache backend.

At any rate, the `drush cc drush` is an absolute must to remove, and if the cr is not removed completely, then BLT should only invalidate the cache objects required for refreshing configuration in ConfigCommand, which in this case I believe is the service container, plugin cache, and config bins, but that should likely be in a dedicated PR and discussed separately due to the urgency of this one.

PR #3169, opened 2018-10-12
EditorConfig: Use four-space indents for Python scripts

hugovk:
Changes proposed:
---------

- Some Python scripts are in our repo
- Python uses four spaces for indents
- https://www.python.org/dev/peps/pep-0008/#indentation

Steps to replicate the issue:
----------

1. Edit a Python script
2. Press tab
-> Adds two spaces

Steps to verify the solution:
-----------

1. Edit a Python script
2. Press tab
-> Adds four spaces

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
 # Drupal editor configuration normalization
-# @see http://editorconfig.org/
+# @see https://editorconfig.org/
 
 # This is the top-most .editorconfig file; do not search in parent directories.
 root = true
@@ -17,5 +17,5 @@ insert_final_newline = true
 [Makefile,*.mk]
 indent_style = tab
 
-[composer.json]
+[{composer.json,*.py}]
 indent_size = 4


danepowell:
Are python scripts commonly used with some particular Drupal or BLT feature? I've never seen python files in a Drupal codebase, most people use PHP (for obvious reasons 😄 )

There are thousands of scripting languages out there and BLT can't accommodate them all. I know this is a pretty tiny change, but I think it still needs to be justified as to why Python is special.

hugovk:
We're using Python scripts as a tool to do extra stuff on the CI, processing Lighthouse output and sending it to some other channels.

I understand this is not really BLT's worry, if we want to use a thousand scripting language, that's our choice and our problem :)

I guess the root problem is that BLT replaces any edits we make to `.editorconfig`.

Is there a way we can override it and avoid BLT replacing it?

Thank you!

mikemadison13:
@hugovk when you say BLT overwrites the .editorconfig file, that only happens during blt updates, right? you should be able to modify the file, commit your version, and then allow blt to overwrite it during updates (and diff what blt wants vs. what you had previously to keep the changes). or is there something else going on?

hugovk:
Yes, something like that.

Currently, we modify and commit it. The next BLT update overwrites it so we either have to lose our changes, or manually fix it again. We don't want to re-fix things.

I'd rather have our version not overwritten and forget about the BLT update to this particular file, so we can maintain code style in our codebase without worry.

danepowell:
Making it easier to manage the editor config file on a per-project basis seems like a better solution here than adding one-off rules for languages (i.e. Python) to BLT.

mikemadison13:
a couple of additional thoughts here:

we shouldn't change the default editor config file for this, so this PR needs to close. HOWEVER the use case presented here is 100% valid, and there are 2 paths forward, 1 immediate and 1 longer term.

the immediate fix is to define your own ignore list https://github.com/acquia/blt/blob/3d70b03917e5c956687394c9ac758d24be521bc2/config/build.yml#L26 in your blt.yml file, and add editor config to that, so blt stops updating that file.

the longer term fix is to add this file to the blt global ignore list by default.

PR #3171, opened 2018-10-12
Prevent cache collisions in multisite install tasks

### LINKED ISSUE #3158: Refactor PR 3006 to Unblock Deploys/Updates

lcatlett:
Fixes #3158 subtasks
--------

Changes proposed:
---------
- Creates and sets unique drush cache directory per site process for post-install site tasks on ACSF
- Adds BLT update hook to refresh hooks

Steps to replicate the issue:
----------
the original issue can be replicated by running a site install / create new site on ACSF. When the existing post-install Factory hook is executed, race conditions result from long running drush tasks such as `cc drush` which then cause failures in importing config and bootstrapping the correct database.

Steps to verify the solution:
-----------
1. Deploy PR to ACSF environment
2. Create site
2. Validate that post-install hook executes without errors. To test via the command line as if the task was run by the factory, run `CACHE_PREFIX='/mnt/tmp/sandbox.01live' \drush8 --root='/mnt/www/html/sandbox.01live/docroot' -l 'site.sandbox.acsitefactory.com' scr --script-path='/mnt/www/html/sandbox.01live/factory-hooks/post-install' 'post-install.php'`, substituting your site, environment, database role, and uri as needed.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -7,7 +7,6 @@
 
 use Acquia\Blt\Robo\Config\ConfigInitializer;
 use Drupal\Component\Utility\Bytes;
-use Symfony\Component\HttpFoundation\Request;
 
 /**
  * Host detection.
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -92,14 +92,10 @@ $is_prod_env = $is_ah_prod_env || $is_pantheon_prod_env;
  * Site directory detection.
  */
 if (!isset($site_path)) {
-  try {
-    $site_path = DrupalKernel::findSitePath(Request::createFromGlobals());
-  }
-  catch (BadRequestHttpException $e) {
-    $site_path = 'sites/default';
-  }
+  $site_path = \Drupal::service('site.path');
 }
 $site_dir = str_replace('sites/', '', $site_path);
+
 
 /*******************************************************************************
  * Acquia Cloud Site Factory settings.
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -5,12 +5,9 @@
    * Setup BLT utility variables, include required files.
    */
 
-  use Acquia\Blt\Robo\Config\ConfigInitializer;
+use Acquia\Blt\Robo\Config\ConfigInitializer;
 use Drupal\Component\Utility\Bytes;
-use Drupal\Core\DrupalKernel;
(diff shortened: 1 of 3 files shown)


settings/blt.settings.php
```diff
@@ -92,15 +92,11 @@
  * Site directory detection.
  */
 if (!isset($site_path)) {
-  try {
-    $site_path = DrupalKernel::findSitePath(Request::createFromGlobals());
-  }
-  catch (BadRequestHttpException $e) {
-    $site_path = 'sites/default';
-  }
+  $site_path = \Drupal::service('site.path');
```
ba66e77: Why is this change necessary?

src/Update/Updates.php
```diff
@@ -627,4 +627,26 @@ public function update_9002000() {
     $this->updater->getOutput()->writeln($formattedBlock);
     $this->updater->getOutput()->writeln("");
   }
+
+    /**
+   * 9.2.1.
```
ba66e77: Should this be 9.2.1 since it's being applied to 9.2.0 and not 9.2.1 (presumably)?

settings/blt.settings.php
lcatlett: I inadvertently put the notes about this in #3172 rather than this one since the two PRs are dependent on one another: Update site path detection to use core site.path service due to findSitePath bug always returning first default directory (see https://www.drupal.org/project/drupal/issues/2280409).

src/Update/Updates.php
lcatlett: I deliberately made the hook 9.2.1 to ensure that the factory hook templates were refreshed for both post-install and db-update since db-update was merged previously in #3166. In looking back through the other hooks again, it doesn't seem like this has been a standard practice so I will remove the 9.2.1 hook

egnikitin:
I'm sorry, but looks like this issue has not been resolved completely.

Issue #3148, opened 2018-10-05
Resolve "Could not scan for classes inside" notices


danepowell:
This is a follow-up to #3120: in BLT 10.0.x. these notices are generated on composer installs. See that issue for more information.

You can see an example of these errors in our tests: https://travis-ci.org/acquia/blt/jobs/438701163#L1411

> source:build:composer
[Exec] Running export COMPOSER_EXIT_ON_PATCH_FAILURE=1; composer install --ansi --no-interaction in /tmp/blt-sandbox-instance
Could not scan for classes inside "/tmp/blt-sandbox-instance/vendor/drupal/core/lib/Drupal.php" which does not appear to be a file nor a folder
(etc)

anavarre:
So yes, it's EXACTLY the same missing classes we're running into with BLT 9.2.0.

mariagwyn:
Just noting that I am seeing this too.

grasmash:
```
composer require drupal/core:~8.5.0 --no-update
rm -rf docroot/core vendor composer.lock && composer update
composer dump-autoload
```
Fixes the issue.

The issue reappears after:
```
composer require drupal/core:^8.6.2 --no-update
rm -rf docroot/core vendor composer.lock && composer update
composer dump-autoload
```

grasmash:
A comparison of the changes to Drupal core's composer.json 8.5.x ... 8.6.x:

https://github.com/drupal/core/compare/8.6.x...8.5.x#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780

The merge plugin changes look a bit suspicious.

grasmash:
I wonder if moving the "classmap" values from Drupal core's composer.json into the "child" composer.json files would fix it. E.g., add 'Timer.php' to 'classmap' in "core/lib/Drupal/Component/Utility/composer.json".

rumyanar:
The issue does not seem to be caused by BLT nor Drupal itself.
I've been using BLT 9.2.0 and Drupal Core 8.6.2 and I start getting the warnings only after I update the following packages:

```bash
composer update
Gathering patches for root package.
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 2 updates, 1 removal
  - Removing symfony/polyfill-php72 (v1.10.0)
Gathering patches for root package.
Gathering patches for dependencies. This might take a minute.
  - Updating symfony/var-dumper (v4.1.6 => v3.4.17): Loading from cache
[...]

PR #3189, opened 2018-10-22
#3148: Remove webflo/drupal-core-require-dev.

grasmash:
#3148

### DIFF
--- a/subtree-splits/blt-require-dev/composer.json
+++ b/subtree-splits/blt-require-dev/composer.json
@@ -12,7 +12,6 @@
         "se/selenium-server-standalone": "^2.53",
         "dmore/behat-chrome-extension": "^1.0.0",
         "sensiolabs-de/deprecation-detector": "dev-master",
-        "webflo/drupal-core-require-dev": "^8.6.1",
         "phpunit/phpunit": "^4.8|^6.5",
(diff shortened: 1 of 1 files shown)


danepowell:
We added webflo/drupal-core-require-dev to composer.required.json in 9.2.x so that folks could run things like core tests.

The dependencies for these tests are in core's require-dev. Ordinarily, if someone downloaded Drupal core directly and ran composer install on it, they'd get these dependencies automatically. But composer only pulls dev dependencies from the root composer.json, so users of BLT _don't_ get these dependencies unless BLT also includes them (via webflo/drupal-core-require-dev) in its own require-dev.

It seems like this PR will break people's ability to run core tests, among other things.

mikemadison13:
@grasmash any thoughts on @danepowell's comments above?

malcolmp:
Maybe move to composer.suggested.json? This dependency added some confusion to us when trying to update from core 8.6.2 to 8.6.3 because of it's dependency on "drupal/core": "8.6.2" without a version range constraint. This blocks a simple composer update drupal/core to get the latest core release. That's fine if you know that this requirement is there and you can update it as well, but otherwise may frustrate users.

hugovk:
Please see https://github.com/acquia/blt/pull/3301 to update docs with the command to update Drupal core in BLT.

PR #3193, opened 2018-10-23
[10.0.x] Add patchutils through extra_packages in VM config (#3191)

### LINKED ISSUE #3191: Add patchutils through extra_packages in VM config.
As a user who uses BLT for Drupal contrib development, I would like to be able to install [patchutils](http://freshmeat.sourceforge.net/projects/patchutils) via `extra_packages` so that I can run the `interdiff` and other patch related commands inside the VM.

greylabel:
Fixes #3191 
--------

Changes proposed:
---------
- Add `patchutils` through `extra_packages` in VM config.

Steps to verify the solution:
-----------

1. Create new project with with BLT 10.0.x
2. Apply changes from this PR to `acquia/blt`
3. Run `blt vm` and proceed with `y`, choose either `beet/box` or `geerlingguy/ubuntu1604`
4. When provision  is complete, `vagrant ssh`
5. `interdiff` should return output like:
```
usage: interdiff [OPTIONS] patch1 patch2
       interdiff --version|--help

{...}

```

### DIFF
--- a/src/Robo/Commands/Vm/VmCommand.php
+++ b/src/Robo/Commands/Vm/VmCommand.php
@@ -277,6 +277,10 @@ class VmCommand extends BltTasks {
     switch ($base_box) {
       case 'beet/box':
         $config->set('workspace', '/beetbox/workspace/{{ php_version }}');
+        $config->set('extra_packages', [
+          'patchutils',
+          'sqlite',
+        ]);
         $config->set('installed_extras', [
           'drush',
           'nodejs',
@@ -287,6 +291,10 @@ class VmCommand extends BltTasks {
 
       case 'geerlingguy/ubuntu1604':
         $config->set('workspace', '/root');
+        $config->set('extra_packages', [
+          'patchutils',
+          'sqlite',
+        ]);
         $config->set('installed_extras', [
           'adminer',
           'selenium',
--- a/scripts/drupal-vm/config.yml
+++ b/scripts/drupal-vm/config.yml
@@ -78,6 +78,7 @@ nodejs_npm_global_packages:
   - name: yarn
 nodejs_install_npm_user: "{{ drupalvm_user }}"
 npm_config_prefix: "/home/{{ drupalvm_user }}/.npm-global"
+extra_packages: ${extra_packages}
 installed_extras: ${installed_extras}
 
 # PHP 7.1.


src/Robo/Commands/Vm/VmCommand.php
```diff
@@ -277,6 +277,10 @@ protected function setBaseBox($config) {
     switch ($base_box) {
       case 'beet/box':
         $config->set('workspace', '/beetbox/workspace/{{ php_version }}');
+        $config->set('extra_packages', [
```
ba66e77: Could this and the other change in this file be refactored to be more DRY? Maybe something like
```
$config->set('extra_packages', $this->getExtraPackages());

protected function getExtraPackages() {
   return [
          'patchutils',
          'sqlite',
        ];
}
```

ba66e77: Actually, better is probably to allow packages to be added in the `box/config.yml` instead of hardcoding them in.

greylabel: @ba66e77 I am following the pattern used for `installed_extras` -- not sure why this should be any different. Thoughts? I think an overall refactor should be done in a separate issue if that is a change that is requested.

(I included `sqlite` since that is currently installed by default by DrupalVM we should maintain that.)

greylabel: @ba66e77 @danepowell ping.

danepowell: I agree that an overall refactor to make this configurable would be good. But that should probably be a follow-up issue. Since this PR follows the existing pattern, I'm okay with it.

greylabel: Created https://github.com/acquia/blt/issues/3252 for the followup.

geerlingguy:
Quick question; is the idea that more than half of BLT's userbase would need patchutils for some reason or another? I personally haven't added it to any of my BLT projects in the past few years (as I use `git` to do any patch-based operations; `git diff [stuff]`, `git apply -v [patch]`, etc.)...

Because if it's just a nice thing to have for a particular project, I recommend just overriding the `installed_extras` variable in your `box/config.yml` or local VM config instead. I'm guessing adding this in wouldn't affect people who have already customized `installed_extras` in their own `box/config.yml`?

I just don't know of the utility of adding that package in for all BLT users unless it's a very commonly-installed tool.

greylabel:
@geerlingguy agreed `git` is the essential tool for patch-based operations, but `patchutils` provides the `interdiff` command ([Creating an interdiff](https://www.drupal.org/documentation/git/interdiff)) which is pretty essential for working with projects on D.O. I feel including `patchutils` in BLT improves its usefulness as a development tool when used for standalone Drupal core and contrib development or as part of a site build. If others disagree, I am happy to just add it manually to my projects, but this addition is a timesaver for anyone using BLT and contributing patches on D.O.

geerlingguy:
@greylabel - Makes some sense, though I usually just use Drupal VM or some other tool when doing drupal core/contrib development (since BLT is great more for the 'full Drupal project lifecycle management' and less for "bring up a local site, do some testing, then delete it").

I use `git diff` to do my interdiffs, never used an `interdiff` command though. I just:

1. Grab patch from Drupal.org
2. Create branch and apply patch
3. Commit those changes
4. Make my new changes
5. `git diff HEAD > interdiff.txt`
6. `git diff 8.x-1.x > actual-patch.patch`

But I know everyone has a unique workflow. I just don't think (my opinion here) that enough people use BLT itself as a Drupal core and contrib workflow tool to justify adding this logic/complexity.

danepowell:
Thanks a lot for the PR, we do appreciate the contribution despite the delay in reviewing it.

I agree with Jeff that doing this kind of intensive development inside the VM may not be ideal, but it's the workflow that we've prescribed for now so we should support it.

PR #3197, opened 2018-10-26
Do not merge: Debugging to resolve composer notices bug

### LINKED ISSUE #3148: Resolve "Could not scan for classes inside" notices

lcatlett:
Fixes #3148 
--------
On BLT 9.2.x, #3148 appears to be caused by a bug in the composer merge plugin which required manually including files in drupal core to scan for autoloader updates (see https://github.com/drupal/core/commit/8f5164829d1b8e5c1d67493630ea9136ed2787b6#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780). This workaround in core can cause autoloader conflicts when running `composer install` and other composer tasks on packages that come alphabetically before `wikimedia/composer-merge-plugin`, in this case `webflo/drupal-core-require-dev`, where drupal core paths are prematurely locked to the vendor directory rather than the path set by BLT using composer/installers.    

Changes proposed:
---------
- Moves webflo/drupal-core-require-dev to composer.suggested.json

Steps to replicate the issue:
----------
1. locally or on travis: run ./vendor/bin/robo release:test
2. Observe notices such as the following: 
```
```

Steps to verify the solution:
-----------
1. on travis: run build
2. locally: run ./vendor/bin/robo release:test
Observe notices are no longer visible

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -92,8 +92,8 @@ class RoboFile extends Tasks implements LoggerAwareInterface {
       ->exec("composer install")
       // I have no idea why this is necessary, but testing on OSX does not pass
       // without it.
-      ->exec("rm -rf $test_project_dir/vendor")
-      ->exec("composer install");
+      ->exec("rm -rf $test_project_dir/vendor $test_project_dir/composer.lock")
+      ->exec("composer dump-autoload --optimize && composer install -vvv");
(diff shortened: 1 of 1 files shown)


grasmash:
Can you fix this on 10.x ? :)

PR #3202, opened 2018-11-01
Adding BLT commands for db/files cloud hooks.

### LINKED ISSUE #2781: BLT 9 Cloud Hooks changes

dpagini:
Closes #2781.
This is an alternate to, and closes #2782. Some good conversation over in that thread.

First, this adds some cloud hook scripts for `post-db-copy` and `post-files-copy` which invokes BLT. I made a choice, for now at least, to make these called methods _do nothing_.

There were two points in the last PR that made me decide on this path for now, and not trying to run update commands...

rjgwiz:
> The only concern I would have is artifact:update:drupal:all-sites is being run on every ACE db copy.

grasmash:
> The only reason I might not want this would be if I wanted to perform the update steps manually for debugging purposes.

I personally want an update command being run on my database when it gets pulled down. I would be OK implementing that manually for now. I also don't think it makes sense to run the update command unless we can limit it to the database that's being copied down, matching to a site. _I'm thinking maybe extracting the database name from the drush command?_ Seems like a good follow up candidate, anyways.

So what this DOES give me, is the ability to use the consolidation/annotated-command hooks to run custom logic before/after/instead of this command. I know that I can do this w/o this contribution, but it really gives a more formalized and standard approach.

So if this goes in, some follow up ideas...
* Come up with a way to turn on/off cloud hook commands on demand (_this is probably not solvable in this library_).
* Provide some helpful cloud API commands... examples:
  * Run update command on only the site for which a database is being pulled.
  * Clear varnish cache after copying down files.
* Allow users to opt-in to these helpful commands via BLT.yml file.


dpagini:
cc: @danepowell - wanted to get your thoughts on this one since you had some on the last version of this...? and thoughts on the suggested long term approach here?

danepowell:
I think as-is, this PR is fine since it shouldn't have any functional effect and just allows people to hook into these commands if they want to.

I haven't had time to circle back after my comment in https://github.com/acquia/blt/pull/2782 and consider a long-term strategy for staging deployments. My concerns from that thread still apply. I think the ideal solution would need to balance two factors as part of the staging process:
* Fast and bit-for-bit identical database copies that happen prior to deploying new code in a lower environment. This generally leans towards _not_ having post-db-copy hooks.
* Configuration changes specific to lower environments should be applied in the case where you aren't necessarily deploying new code right away (the Shield / Env Indicator problem), which _does_ lean toward having post-db-copy-hooks.

PR #3218, opened 2018-11-06
Fixes #3217 to provide acsf-init-verify as part of git hooks and blt …

### LINKED ISSUE #3217: Validate ACSF Settings during BLT Validate
Proposed Changes:

- Add a new tests:acsf:validate command that runs the acsf-init-validate command to be run during blt:validate process. This will ensure that builds fail when the acsf-init command needs to be run, hopefully preventing failed deployments

Note that I propose we add a new config key to prevent this from running on all sites:

- validate
-- acsf: true

mikemadison13:
…validate.

Fixes #3217  
--------

Changes proposed:
---------

- adds validate / acsf to template blt.yml
- adds a acsf-init-validate command to blt validate
- adds a acsf-init-validate command to git hooks

Steps to verify the solution:
-----------

1. update your blt.yml file
2. run either blt validate or tests:acsf:validate

Warning, should not be run on non-acsf projects (which is why it defaults to not running).


mikemadison13:
Outstanding question if it makes sense to do this as part of the pre-commit hook or not. It seems to me that just having the safety check in blt validate (so builds fail) is sufficient.

danepowell:
I like the idea of this, primarily to solve the problem of accidentally updating the ACSF module and failing to re-run acsf-init.

I don't think a pre-commit hook is the logical place for this. pre-commit hooks should be reserved for checks for the most frequent problems like syntax errors and bad commit messages. Failing to run acsf-init is something that happens once or twice per project. Better to do that in travis and avoid slowing down commits any further.

PR #3227, opened 2018-11-09
Fixes #3206 to include sensiolabs/security-checker and new composer s…

### LINKED ISSUE #3206: Composer Security Scans
As a site architect I want to know that my composer.lock does not have vulnerabilities.  

Considering adding a blt command to wrap something like this:
https://github.com/sensiolabs/security-checker

Anyone have better approach or solution before we start building it?

cc @CashWilliams, @grasmash, @lcatlett

mikemadison13:
…ecurity test.

Fixes #3206 
--------

Changes proposed:
---------

- adds sensiolabs/security-checker to composer.json
- adds new tests:security:check:composer command
- bundles tests:security:check:composer into blt tests:all
- updates docs to reference new command

Steps to verify the solution:
----------

1. update blt to include this patch
2. composer install (and ensure that sensiolabs/security-checker is installed)
2. run blt tests:security:check:composer

Outstanding question:
- how to make sure that the composer.json file is updated for people already updated to 10.0.0? Do we need to run a new update script? @grasmash / @danepowell  how has this been handled in the past?


mikemadison13:
@aweingarten does this more or less meet what you wanted #3206 to facilitate? please give it a whirl.

aweingarten:
This is awesome!  I back-ported for myself to 9.2.x.  Tested it on local & CI with both passes and fails.  Worked like a charm!

mikemadison13:
@danepowell yeah, i know in the future we've done a composer.json munge. I wasn't sure how much that would be changing (if at all) with the composer merge changes that are coming along. I also wasn't clear if that needed to happen here specifically or not.

Personally, i think this would be a good addition and it's minimal. Can we merge based on @aweingarten's review?

danepowell:
Looks good to me

PR #3232, opened 2018-11-10
WIP: Scan classes

### LINKED ISSUE #3148: Resolve "Could not scan for classes inside" notices

lcatlett:
Fixes #3148 
--------

Changes proposed:
---------

- Updates composer default config for core package
- Enable autoloading by default for projects

Steps to replicate the issue:
----------

1. 
2. 

Steps to verify the solution:
-----------

1.
2.


grasmash:
I think subsequent PRs are trying to address this problem differently. E.g., https://github.com/acquia/blt/pull/3299

PR #3240, opened 2018-11-14
Simplifying onboarding doc

danepowell:
Just minor tweaks to make this clearer. Mainly, we no longer advise people to run `blt vm`, just go straight to `vagrant up`.

We used to suggest starting with `blt vm` because this would set your local drush alias accordingly, but since you can no longer run commands outside of the VM anyway that's a moot point.


briantully:
Ahh I see it now in context. OK, understood. I guess if we're still going to use the following line

> If your project uses a virtual development environment such as Drupal VM:

It would be extra-friendly if "Drupal VM" was linkable to the Drupal VM quick start guide, e.g.

`If your project uses a virtual development environment such as [Drupal VM] (https://github.com/geerlingguy/drupal-vm#quick-start-guide):
`

briantully:
Good idea! I’m assuming you mean this page https://blt.readthedocs.io/en/latest/local-development/#using-drupal-vm-for-blt-generated-projects ?

mikemadison13:
Yep

PR #3244, opened 2018-11-15
10.0.x local settings file rename

timcosgrove:
Fixes # 
--------

Changes proposed:
---------
Drupal 8.6 changed the recommend name of local settings files: https://cgit.drupalcode.org/drupal/tree/sites/example.settings.local.php#n10

Since we'd want to accommodate legacy filenames in addition to the new recommended filename, I added additional file includes rather than changing the file.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -302,7 +302,7 @@ if ($is_local_env) {
   if (file_exists(DRUPAL_ROOT . "/sites/settings/local.settings.php")) {
     require DRUPAL_ROOT . "/sites/settings/local.settings.php";
   }
-  // Since Drupal 8.6, the file is now recommended to be named settings.local.php.
+  // Since Drupal 8.6, the file is recommended to be named settings.local.php.
   if (file_exists(DRUPAL_ROOT . "/sites/settings/settings.local.php")) {
     require DRUPAL_ROOT . "/sites/settings/settings.local.php";
   }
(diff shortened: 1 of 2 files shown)


grasmash:
Seems like the better solution might be to rename local.settings.php to settings.local.php, right? This creates the possibility of loading both at the same time, which would be confusing.

settings/blt.settings.php
```diff
@@ -302,8 +302,16 @@
...
     require DRUPAL_ROOT . "/sites/settings/local.settings.php";
   }
-  // Load local settings for given single.
+  // Since Drupal 8.6, the file is recommended to be named settings.local.php.
+  if (file_exists(DRUPAL_ROOT . "/sites/settings/settings.local.php")) {
+    require DRUPAL_ROOT . "/sites/settings/settings.local.php";
+  }
+  // Load local settings for given single site.
   if (file_exists(DRUPAL_ROOT . "/sites/$site_dir/settings/local.settings.php")) {
     require DRUPAL_ROOT . "/sites/$site_dir/settings/local.settings.php";
```
dpagini: Could we trigger some kind of warning if using the old name maybe? Something like...
```php
@trigger_error('Drupal has changed the recommended local settings filename convention from local.settings.php to settings.local.php. Please update your file, as it will no longer be included by BLT in 11.x.', E_USER_DEPRECATED);
```

danepowell:
If we're going to make this change, I think it has to include a migration / update path, and we shouldn't try to support both at once.

PR #3245, opened 2018-11-15
Load settings.local.php in addition to local.settings.php.

timcosgrove:
Fixes # 3246
--------

Changes proposed:
---------
Drupal 8.6 changed the recommend name of local settings files: https://cgit.drupalcode.org/drupal/tree/sites/example.settings.local.php#n10

Since we'd want to accommodate legacy filenames in addition to the new recommended filename, I added additional file includes rather than changing the file.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -319,7 +319,7 @@ if (file_exists(DRUPAL_ROOT . "/sites/$site_dir/settings/includes.settings.php")
  * This is intended to provide an opportunity for local environments to override
  * any previous configuration.
  *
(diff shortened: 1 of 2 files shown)


grasmash:
Seems like the better solution might be to rename local.settings.php to settings.local.php, right? This creates the possibility of loading both at the same time, which would be confusing.

danepowell:
I agree, if we're going to make this switch then I think there needs to be a migration process involved and we shouldn't support both at once. I think it's too potentially disruptive for 9.x, but maybe appropriate for 10.x see #3244

PR #3264, opened 2018-11-26
Adds FAQ for drush 8/9 permission denied issue.

mikemadison13:
Fixes N/A 
--------

Changes proposed:
---------

- updates FAQ to include the Drush 8/9 permission denied scenario.


mikemadison13:
@ba66e77 do you think this would have helped with your recent issue? any issues merging?

PR #3269, opened 2018-11-29
Allow drush ansi disable execution time

dlopez-akalam:
Fixes #3146 
--------
drush8 does not support --ansi option, causing issues when working on remote sites.

Changes proposed:
---------
- Allow to disable ansi in runtime per command

Steps to replicate the issue:
----------
- Example: 
$this->taskDrush()
->alias('my-remote-alias')
->drush('updb')
->ansi(FALSE)

Steps to verify the solution:
-----------
Apply the patch and execute the same command
- Example: 
$this->taskDrush()
->alias('my-remote-alias')
->drush('updb')
->ansi(FALSE)
As you can see, the --ansi option gets omited

### DIFF
--- a/src/Robo/Tasks/DrushTask.php
+++ b/src/Robo/Tasks/DrushTask.php
@@ -81,6 +81,13 @@ class DrushTask extends CommandStack {
   protected $include;
 
   /**
+   * Add or not the --ansi option.
+   *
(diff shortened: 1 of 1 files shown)


grasmash:
Looks good. Can you submit to 10.x as well?

PR #3289, opened 2018-12-14
Backport ACSF improvements from 9.2.x

grasmash:


composer.required.json
```diff
@@ -45,11 +46,6 @@
     "enable-patching": true,
     "patchLevel": {
       "drupal/core": "-p2"
-    },
-    "patches": {
-      "drupal/core": {
```
mikemadison13: @grasmash i'm not sure that this patch should come out, i don't think this issue was actually fixed until 8.6.x (see https://www.drupal.org/node/2752961)

composer.suggested.json
```diff
@@ -6,14 +6,13 @@
...
   "require": {
-    "acquia/drupal-spec-tool": "^2.0.0",
     "acquia/lightning": "^3.1.0",
     "drupal/acquia_connector": "^1.5.0",
     "drupal/acquia_purge": "^1.0-beta3",
     "drupal/cog": "^1.0.0",
     "drupal/devel": "^1.0.0",
     "drupal/qa_accounts": "^1.0.0-alpha1",
-    "drupal/memcache": "2.0-alpha7",
+    "drupal/memcache": "^2.0-alpha7",
```
mikemadison13: the 10.x version of our code is still pinned to 2.0-alpha5, https://github.com/acquia/blt/blob/10.0.x/composer.json#L40, is this likely to cause issues?

grasmash: I'm going to change this to match 9.2.x. But, it seems like 10.x is probably missing some memcache updates that were merged into 10.x.

PR #3299, opened 2018-12-17
Use optimized classloader to resolve notices

### LINKED ISSUE #3148: Resolve "Could not scan for classes inside" notices

lcatlett:
Fixes #3148 
--------

Changes proposed:
---------

- Using a PSR-0 class loader requires a file_exists() check on paths to make sure the class exists when it's to be loaded. This is inefficient and also can result in "Could not scan for classes inside" notices each time a class is missing when the autoloader is dumped. BLT should use Composer's optimized ClassLoader to resolve this 

Steps to replicate the issue:
----------

1. 
2. 

Steps to verify the solution:
-----------

1.
2.

### DIFF
--- a/src/Robo/Commands/Setup/BuildCommand.php
+++ b/src/Robo/Commands/Setup/BuildCommand.php
@@ -101,7 +101,7 @@ class BuildCommand extends BltTasks {
    * @aliases sbc setup:composer:install
    */
   public function composerInstall() {
-    $result = $this->taskExec("export COMPOSER_EXIT_ON_PATCH_FAILURE=1; composer install --ansi --no-interaction")
+    $result = $this->taskExec("export COMPOSER_EXIT_ON_PATCH_FAILURE=1; composer install --ansi --no-interaction --optimize-autoloader --apcu-autoloader ")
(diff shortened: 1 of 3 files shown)


mikemadison13:
This DID fix the problem for me briefly... but after running and re-running composer update a few times it came back.

note that initially i nuked vendor and docroot/core, and cleared composer cache to get it to work the "first time."

mikemadison13:
@geerlingguy any thoughts on the above?

danepowell:
I wasn't able to get this patch to work inside the VM even following Mike's tips. The only thing I did differently from Mike was to install Drupal core from outside the VM, since that hangs (presumably due to NFS latency) when I do it inside the VM. Maybe that's one of the root causes of all of this in the first place.

lcatlett:
I havent finished debugging yet to validate, but while debugging some issues that have come up in recent support tickets, I came across https://github.com/symfony/symfony/issues/29877 which may be related to this error in BLT

lcatlett:
Update - I think the proposed bugfix in #3336 might help here as well

PR #3301, opened 2018-12-17
Add command for updating Drupal core

hugovk:
Changes proposed:
---------

- Update the docs with a working command to update Drupal core, as the command recommended by https://www.drupal.org/docs/8/update/update-core-via-composer and the existing page are out of date.
- See https://github.com/acquia/blt/pull/3189 for comments showing the new webflo requirement causes confusion.

Steps to replicate the issue:
----------

1. Run `composer update drupal/core --with-dependencies`
2. Check if Drupal core has been updated ➡️👎 

Steps to verify the solution:
-----------

1. Run `composer update webflo/drupal-core-require-dev drupal/core --with-dependencies`
2. Check if Drupal core has been updated ➡️👍

### DIFF
--- a/docs/configuration-management.md
+++ b/docs/configuration-management.md
@@ -59,7 +59,7 @@ Take caution when updating core and contributed modules when using any sort of c
 The best way to prevent such problems is to always follow these steps when updating contributed and core modules. These steps assume you are using Config Split or a similar CM strategy using `drush cex` and `drush cim`.
 
 1. Start from a clean install or database sync, including config import (`blt setup` or `blt drupal:sync`). Verify that active and exported configuration are in sync (running `drush config-export` should report no changes.)
-2. Use `composer update` or `composer update drupal/[module_name] --with-dependencies` to download the new module version(s).
+2. Use `composer update` or `composer update drupal/[module_name] --with-dependencies` to download the new module version(s). To update Drupal core, use `composer update webflo/drupal-core-require-dev drupal/core --with-dependencies`.
 3. Run `drush updb` to apply any pending updates locally.
 4. Export any configuration that changed as part of the database updates or new module versions by running `drush config-export` and checking for any changes on disk using `git status`.
 5. Commit any changed configuration, along with the updated `composer.json` and `composer.lock`.


lcatlett:
@mikemadison13 - I think this needs to be `composer update drupal/core webflo/drupal-core-require-dev symfony/* --with-dependencies`

hugovk:
Please see https://github.com/acquia/blt/pull/3303 for the same as this for 10.x.

I've not tested it, does the `symfony/*` stuff not get updated by a standard `drupal/core` update, or is there something in BLT preventing it?

malikkotob:
I feel like each release to core often has slightly different needs in terms of what other packages need to get passed along with the `drupal/core` itself. For the jump to 8.6.x from 8.5.x, for example, the above doc listed: `composer update drupal/core webflo/drupal-core-require-dev --with-dependencies`. I have a recollection of previous releases not needing the `webflo` package to be upgraded, but instead various `symfony` packages to be updated.

Since this could theoretically change with each release, it would mean BLT maintainers would have to maintain the page more than what seemingly makes sense for the scope of BLT. I wonder if it might make more sense for our docs to simply link to core's docs, and if those are out-of-date, to contribute the most up-to-date command there, instead.

It also doesn't seem like their docs are outdated:
image

hugovk:
I ran into the problem with BLT from 8.6.2 to 8.6.3: https://github.com/acquia/blt/pull/3189#issuecomment-440814315

No special notes there: https://www.drupal.org/project/drupal/releases/8.6.3.

malikkotob:
Ah, I see. Yeah, I suppose my recommendation still stands though:

> I wonder if it might make more sense for our docs to simply link to core's docs, and if those are out-of-date, to contribute the most up-to-date command there, instead.

It doesn't seem (to me, at least) in scope for BLT to maintain upgrade instructions which _could_ vary with each release of core (as we're seeing here). At the end of the day it's up to the maintainers, though!

lcatlett:
> Please see #3303 for the same as this for 10.x.
> 
> I've not tested it, does the `symfony/*` stuff not get updated by a standard `drupal/core` update, or is there something in BLT preventing it?

I think updating `symfony/*` separately is often recommended when updating `drupal/core` to prevent dependency conflicts when attempting to update core (see https://www.drupal.org/docs/8/update/update-core-via-composer#known-issues). Perhaps we should either patch drupal-project since there [is already a PR for this functionality](https://github.com/drupal-composer/drupal-project/pull/422), or link to d.o docs to avoid maintaining this in BLT and bearing the burden of updating docs every time core is updated...

PR #3231, opened 2018-11-10
Multisite bugfixes

lcatlett:
Fixes #3207 
--------

Changes proposed:
---------
- Update local multisite drush alias because non-self aliases will cause ssh recursion to localhost within VM
- Fix alias bug for site2 alias in testMultisiteGeneratea
- Fix hard-coded Acquia Cloud environment variable overrides
- Resolve phpcs errors in ACSF and muiltisite hooks/settings

Steps to replicate the issues:
----------

To recreate errors phpunit test errors seen on https://travis-ci.org/acquia/blt/jobs/449314384#L5991-L5999: 
1. Run ./vendor/bin/robo release:tests
2. Observe:
```
1) Acquia\Blt\Tests\Blt\MultiSiteTest::testMultisiteGenerate
Exception: BLT command exited with non-zero exit code.
/home/travis/build/acquia/blt/tests/phpunit/src/BltProjectTestBase.php:276
/home/travis/build/acquia/blt/tests/phpunit/BltProject/MultiSiteTest.php:88
```
3. Run ./vendor/bin/robo sniff-code to observe phpcs and other syntax errors in factory hooks reported in #3207

Steps to verify the solution:
-----------

1. PR build passes
2. `./vendor/bin/robo release:tests` / testMultisiteGenerate passes and no phpcs errors are reported (`./vendor/bin/robo sniff-code` runs as part of the release:tests command)

PR #3326, opened 2019-01-04
Create a valid machine name based on the local domain name.

### LINKED ISSUE #3323: Ensure $acsf_site_name is a valid machine name for config_split.
My system information:
* Operating system type: Ubuntu (vagrant, drupal-vm)
* Operating system version: 16.04.5
* BLT version: 9.2.0

in `settings/blt/blt.settings.php` there is the following code calculating the acsf_site_name for local environment
```
if ($is_acsf_inited) {
  if ($is_local_env) {
    // When developing locally, we use the host name to determine which site
    // factory site is active. The hostname must have a corresponding entry
    // under the multisites key.
    $input = new ArgvInput($_SERVER['argv']);
    $config_initializer = new ConfigInitializer($repo_root, $input);
    $blt_config = $config_initializer->initialize();

    // The hostname must match the pattern local.[sitename].com, where
    // [sitename] is a value in the multisites array.
    $domain_fragments = explode('.', $_SERVER['HTTP_HOST']);
    $name = array_slice($domain_fragments, 1);
    $acsf_sites = $blt_config->get('multisites');
    if (in_array($name, $acsf_sites)) {
      $acsf_site_name = $name;
    }
  }
}
```
Where the sitename is token raw from the domain name. After that, the site name is used to activate a config split with this machine name, resulting in a potentially invalid machine name: local.my-site.com for example will take "my-site" as acsf_sitee_name. It should be transformed in a valid machine name like my_name.

dlopez-akalam:
Fixes #3323 
--------

Changes proposed:
---------
- Prepare local sitename detected from domain prior to asigining to acsf_site_name variable

Steps to replicate the issue:
----------
1. Create a local site called local.my-site.com
2. Create a config_split with machine name my_site, disabled by default
3. Blt will not activate it via config override mechanism.

Steps to verify the solution:
-----------
1. Apply the patch.
2. See the config_split successfully enabled.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -151,6 +151,12 @@ if ($is_acsf_inited) {
     $name = array_slice($domain_fragments, 1);
     $acsf_sites = $blt_config->get('multisites');
     if (in_array($name, $acsf_sites)) {
+
+      // Create a valid machine name based on the local domain name.
+      $name = strtolower($name);
+      $name = preg_replace('/[^a-z0-9_]+/', '_', $name);
+      $name = preg_replace('/_+/', '_', $name);
+
       $_acsf_site_name = $name;
     }
   }
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -148,7 +148,7 @@ if ($is_acsf_inited) {
     // The hostname must match the pattern local.[sitename].com, where
     // [sitename] is a value in the multisites array.
     $domain_fragments = explode('.', $http_host);
-    $name = array_slice($domain_fragments, 1);
+    $name = $domain_fragments[1];
     $acsf_sites = $blt_config->get('multisites');
     if (in_array($name, $acsf_sites)) {
 


settings/blt.settings.php
```diff
@@ -148,9 +148,15 @@
...
+    $name = $domain_fragments[1];
     $acsf_sites = $blt_config->get('multisites');
     if (in_array($name, $acsf_sites)) {
+
+      // Create a valid machine name based on the local domain name.
+      $name = strtolower($name);
+      $name = preg_replace('/[^a-z0-9_]+/', '_', $name);
+      $name = preg_replace('/_+/', '_', $name);
+
       $_acsf_site_name = $name;
```
grasmash: This is used in other places so I'm hesitant to mess with it. Might be better to do this when creating the split name in config.settings.php.

lcatlett:
Closing this as outdated alongside #3323, as this has since been addressed in BLT 9.2.2 via #3231 and #3297. I initially missed that BLT 9.2.0 is being used so this PR uses the now outdated config in `blt.settings.php`, `config.settings.php` and the recently updated factory hooks. @dlopez-akalam -  if you are still having problems after updating BLT and the ACSF module and/or using the refactored `blt multisite` command, please open a new issue.

Issue #1793, opened 2017-07-13
Allow custom commands to use autoloader


thom8:
My system information:

- Operating system type: MacOS Sierra
- Operating system version: 10.12.5
- BLT version: 8.9.0-rc2

We are looking to export our custom Commands and Hooks so they can be shared between projects

At the moment custom Commands and Hooks are restricted to `./blt/src/*` as defined [here](https://github.com/acquia/blt/blob/0de0aa71f9340d8b1523ed8d7f57c3f39f574f3e/src/Robo/Blt.php#L105).

So the autoload configuration [here](https://github.com/acquia/blt/blob/8.x/composer.required.json#L35) seems to be ignored.

Is it possible to make use of the autoloader via the `Acquia\Blt\Custom\Commands / Acquia\Blt\Custom\Hooks` namespace for any custom commands or hooks?

grasmash:
Yeah, I think we basically just need to look in `vendor` for files in matching a certain filename pattern or belonging to a certain namespace. We could also try to use reflection to find anything that implements a particular interface.

@greg-1-anderson any suggestions on the best way to discover command files provided by composer packages?

thom8:
@grasmash I think it would be OK to assume another package would need to declare the namespace via the autoloader -- https://github.com/acquia/blt/blob/8.x/composer.required.json#L33

This should make it easier than traversing the `vendor` directory.

The registered paths are available in `vendor/composer/autoload_psr4.php`

thom8:
Something along the lines of >>

```php
    $loader = require 'path/to/vendor/autoload.php';
    $prefixes = $loader->getPrefixesPsr4();
    $paths = is_array($prefixes) ? $prefixes['Acquia\Blt\Custom\\'] : [];
```

Will give you the list of paths which declare the `Acquia\Blt\Custom\` namespace.

greg-1-anderson:
Terminus implemented a plugin mechanism that is not based on Composer; I have not yet implemented a plugin mechanism that works with resources in the vendor directory, although I think it would be a good idea to do so.

@thom8's suggestion seems reasonable. Is [getPrefixesPsr4()](https://getcomposer.org/apidoc/master/Composer/Autoload/ClassLoader.html#method_getPrefixesPsr4) a public Composer API? It's public, but undocumented.

thom8:
This could also be handy for Robo to specify a custom path to a project `RoboFile.php` via a namespace or direct classmap.

nathandentzau:
Hi,

[...]

PR #3347, opened 2019-01-24
Adds Bitbucket pipeline support.

b-sharpe:
- Adds an initial example Bitbucket CI configuration file.
- Adds a Robo task that copies the template to the project root.
- Adds a settings includes and bitbucket ci detection.

### DIFF
--- a/scripts/bitbucket/bitbucket-pipelines.yml
+++ b/scripts/bitbucket/bitbucket-pipelines.yml
@@ -1,5 +1,5 @@
 # Ideally use an image with php, php extensions & composer.
-# i.e. image: sharpe/ci:latest
+# i.e. image: imagexmedia/ci:latest
 # See: https://confluence.atlassian.com/bitbucket/php-with-bitbucket-pipelines-873907835.html
 
 # Defines single re-usable scripts.
--- a/scripts/bitbucket/bitbucket-pipelines.yml
+++ b/scripts/bitbucket/bitbucket-pipelines.yml
@@ -1,0 +1,83 @@
+# Ideally use an image with php, php extensions & composer.
+# i.e. image: sharpe/ci:latest
+# See: https://confluence.atlassian.com/bitbucket/php-with-bitbucket-pipelines-873907835.html
+
+# Defines single re-usable scripts.
+script_definitions:
+  - &env-setup
+    export BIN_DIR=$BITBUCKET_CLONE_DIR/vendor/bin
+  - &composer-tasks
+    composer self-update --no-interaction && composer install --no-interaction --optimize-autoloader
+
+# Defines re-usable steps.
+step_definitions:
+  - validate: &validate
+      name: 'Validation Tasks.'
+      caches:
+        - composer
+      script:
+        - *env-setup
+        - *composer-tasks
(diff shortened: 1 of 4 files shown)


TravisCarden:
@ba66e77 / @danepowell : @b-sharpe's company has decided to adopt BLT internally, and they need this Bitbucket support to get off the ground. Can we rustle someone up to provide some support? Thanks! 😃

aweingarten:
A word of caution here.  While pipelines _technically_ supports bitbucket, in practices there are some nasty landmines here.   When I last tried this about 10 months ago bitbucket's API had a major limitation.  It didn't allow anyone including pipelines to pull in the branch from a fork inside a PR. It's a major limitation of the permission model of bit bucket.  Another manifestation is that you can't see a peer's branches unless you directly add them as collaborators to your fork .  

To work around I think pipelines is taking the patch/diff from bitbucket and applying that to the upstream branch to get a workable reproduction of the fork.  This works fine 80% of the time.  However, if you do a large refactor and move thousands of files this approach will fail. Somewhere it should be documented that bit-bucket is not a first class citizen in this ecosystem.

danepowell:
@aweingarten @TravisCarden @b-sharpe I think there might be some confusion here.

I think the purpose of this issue to support Bitbucket's product called Pipelines, _not_ to add Bitbucket integration to _Acquia's_ product called Pipelines. Furthermore, I think Adam's comment was in reference to Acquia Pipelines. Is that correct?

Assuming so, I'll forward that comment to the Acquia Pipelines team to follow up on.

b-sharpe:
@aweingarten While I agree BB has some limitations, I don't think that should hold up allowing people to use it with BLT. 

Most using BB should already be aware of these limitations, but I could put a doc page explaining the available process due to the limitations if that helps move this along?

Here are currently the different ways to deal with the limitations:

1. Don't fork. Smaller teams this works fine, as PRs from within the same Repo run pipelines with zero issue. BB supports a wide array of branch permissions, so you can even just lock develop/master merging to the Team Lead depending on your git workflow

2. Git Hooks. If forking, we can implement a new pre-commit that forces the same as a PR would (i.e. run behat tests). Then the PR can either be merged with just a code review, or if even further precaution is needed, then merge to a feature branch before develop/master to be tested.

3. Individual pipelines. If forking, pipelines can still run on the fork's repo, in which case while the PR to the upstream cannot run a pipeline, it CAN however see if the source's pipeline has passed or failed. This requires the fork to have pipeline minutes.

4. Your example. We do not use (or advise using) this method due to the issues you mentioned

As @TravisCarden mentioned, We are trying to trying to get our teams on board with using BLT, but unfortunately switching from BB is not an option at this point (and I would assume for other companies as well), so I am willing to contribute to moving this forward in any way I can.

danepowell:
This looks really useful for folks using Bitbucket Pipelines, and we appreciate the contribution. Unfortunately, the BLT maintainers don't have infrastructure in place to access Bitbucket Pipelines or maintain features related to it, so we can't support it. Individual users can still use this PR as a reference and implement it themselves.

danepowell:
I also suggest following #1793 as a means to provide Bitbucket Pipelines integration as part of a distributable package / plugin.

PR #3349, opened 2019-01-27
Added a cache-clear to BLT to ensure memcache/db discrepancies do not…

aweingarten:
When config split is used to enable DBlog in a lower env and you sync a Db from prod down on Acquia.  Memcache will think that DBLog is enabled immediately after a DB sync.  This will lead to CMI import failures as module enables will attempt to log to the nonexistent watchdog table.

A CC before a config import will solve the issue.

### DIFF
--- a/src/Robo/Commands/Setup/ConfigCommand.php
+++ b/src/Robo/Commands/Setup/ConfigCommand.php
@@ -51,6 +51,10 @@ class ConfigCommand extends BltTasks {
         }
       }
 
+      // Caching can interfere with config import depending on the state
+      // of DB relative to memcache.
+      $this->drush("cache-rebuild");
+
       $task = $this->taskDrush()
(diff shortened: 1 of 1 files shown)


dpagini:
Just pointing out, this is something that I see get added and removed from BLT on a cycle...  

see #3166 for an example...

It would probably be beneficial to write some kind of test for these scenarios so it doesn't keep getting removed and re-added, but I'm sure that's not _easy_.

mikemadison13:
i think this is something we do NOT want to introduce. there have been a lot of "to clear or not to clear" caches in various scenarios. i 100% understand this is an issue for you @aweingarten  but i don't want BLT breaking deployments for a ton of other projects because we injected a new cache clear.

i'm open to discussing this if other people start reporting this issue, but thus far, yours is the only case that has cropped up.

Issue #1787, opened 2017-07-11
Better support using BLT for testing core and contrib with PHPUnit.


greylabel:
It would be nice to have better support, documentation, and configurations for using BLT to _Run All PHPUnit Tests The Way The Testbot Does It_™, including kernel tests and browser tests.

- [x] Add setup instructions for PHPStorm + PHPUnit (#3198).
- [ ] Add new task that supports passing addition arguments to `phpunit` (#3249).
- [ ] Add command to set environment variables need for Drupal testing (#3249).
- [ ] Add new task and command to run Drupal's `run-tests.sh` (#3249).
- [ ] Instantiate `chromedriver` for function JavaScript testing (#3249).
- [ ] Decide on final schema for testing commands and refactor (#1524 and #3249).

timcosgrove:
One thing I'm running into playing with this is, we definitely want BLT to be able to interoperate with Drupal 8.4.x in this scenario, so that we can run tests against the same Drupal D.O. does.

The main conflict between the two seems to be in the version of symfony/yaml. drupal/core 8.4.x specifies ^3.2, while both acquia/blt and grasmash/yaml-cli specify ~2.8, probably in line with what drupal/core 8.3.x specified.

Would moving acquia/blt and grasmash/yaml-cli to symfony/yaml ^3.2 be a breaking change?

timcosgrove:
Extremely anecdotally, I can get a composer update/install working and then things going again by:

* updating yaml-cli symfony dependencies to 3.2 (https://github.com/timcosgrove/yaml-cli/tree/feature/symfony-3.2)
* updating blt symfony dependencies to 3.2 and drush to 9.x (https://github.com/timcosgrove/blt/tree/testing-symfony-3.2)
* point to these repos + drupal/core 8.4.x (local BLTED8 project)
* modify drush.wrapper to point to vendor/drush/drush/drush (local BLTED8 project)

yaml-cli and blt run their tests without issue with the new dependencies.  I realize this is not proper usability/functional testing, but it is a start.

I also remember the issues w/ drush 9.x and Acquia, so this might be a non-starter.

grasmash:
We can modify BLT's version constraints to allow `^2.8|^3.2`. That should work. Otherwise, we would need to start a 8.10.x, to be released when 8.4.0 is cut.

grasmash:
@greylabel Adding:

```
phpunit:
  - path: '${docroot}/core'
```

[...]

PR #3351, opened 2019-01-28
Revert "Revert "[10.0.x] Drupal testing enhancements." (#3348)"

### LINKED ISSUE #3346: New PHPUnit features make BLT internal tests unstable.
@greylabel We're getting frequent failures after merging https://github.com/acquia/blt/pull/3249. 

There are two issues I see:

* The log output has become too long to to view in browser. We need to make it less verbose. 
* Test builds fail with `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.` This may be an issue with the new tests failing to kill background processes, or simply taking too long.

This is a beta blocker. We'll need to either resolve it quickly or else revert and refactor.

greylabel:
Fixes #3346
--------

Changes proposed: See #1787


dpagini:
Could I ask is there anything outstanding on this work?

danepowell:
This looks good to me, @greylabel anything you feel still needs to be done here? The "WIP" title just makes me nervous... :smile:

greylabel:
I _think_ this should be fine to merge with one caveat. 

In #3346 @grasmash noted two issues that needed to be fixed.

> The log output has become too long to to view in browser. We need to make it less verbose

I addressed this in this PR by adding a new pipeline just for ensuring that all Drupal test suites succeed with both test runners.

> Test builds fail with No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself. This may be an issue with the new tests failing to kill background processes, or simply taking too long.

This point has not been directly addressed -- and is why there is there is still a WIP label on this PR, which I will remove now -- and may or may not be an issues still. The problem is that all the differential Travis builds are coming back green as you can see in this thread, but when this PR was merged previously, Travis would time out when running the same tests on the integrated code. I didn't see any obvious reasons why (see links in #3346) and since I don't have admin access to this project, couldn't really meaningfully debug in Travis. I could try to recreate the time out in a fork, but did have not have bandwidth to do so.

I would say we should merge this again and I can be available to help debug if we see the timeout issues return.

P.S. I have some future plans to extend and refactor this work, but getting this base piece in would be really helpful.

dpagini:
To address the "output is too long" concern... is there just a _smaller_ test we could choose to run as part of the testing? That Quickedit one seemed to have like 90+ test entries in it.

greylabel:
@dpagini Probably. I just picked a pretty random set of tests. Would be fine followup issue. Splitting all these tests into their own pipeline seems to address the original concern of Travis's out no longer being readable in the UI console.

dpagini:
Yeah, seems fair... however, it probably slows tests down by creating 6 containers instead of 4 for every build now. Honestly, not my call, just thought I'd bring it up.

danepowell:
It looks like it takes 5 minutes to spin up a new BLT project, and then 5 minutes to run the quickedit tests.

@greylabel if you can just find a smaller test suite I'll merge this.

I created https://github.com/acquia/blt/issues/3436 to follow up and reduce test length / verbosity, which we should do regardless of this PR.

greylabel:
Sounds good. #3436 is a good call and there are certainly some improvement that can be made here.

greylabel:
I simplified the tests overall and removed the Simpletest suite altogether from the tests as that will need to happen for Drupal 8.7 anyway.

danepowell:
Thanks a lot for the contribution! Went ahead and merged, let's follow-up with new issues for any additional work.

greylabel:
@danepowell Thanks for merging this. The last build this [PR](https://travis-ci.org/acquia/blt/builds/504868014) was green, but again there are timeouts for jobs when building the integrated [10.x](https://travis-ci.org/acquia/blt/builds/504915961) branch. Curious if you know any differences on how Travis runs for PRs vs branches that might cause this)?

See jobs https://travis-ci.org/acquia/blt/jobs/504868019 and https://travis-ci.org/acquia/blt/jobs/504915966 specifically.

danepowell:
The only major difference I can think of is that PRs usually don't have access to encrypted environment variables.

At least now we can temporarily disable these tests without having to revert everything, which is what I've done for now. But we need to figure out what's going on. #3444

Issue #3220, opened 2018-11-07
Memcache settings doesn't work with memcache 8.x-2.0


haggmand:
System information:
* BLT version: 9.2.1
* Requiring `/../vendor/acquia/blt/settings/blt.settings.php` in `settings.php`

In memcache.settings.php the bootstrap_container_definition is out of date and does not work with the stable release version of the Memcache module (8.x-2.0).

Correct bootstrap_container_denfinition is discussed here:
[https://www.drupal.org/project/memcache/issues/3002858](url)

In the memcache modules README.txt this is also stated.

lcatlett:
@haggmand BLT currently [locks to the 2.0-alpha7 ](https://github.com/acquia/blt/blob/9.2.x/composer.suggested.json#L16)version of the memcache module because Acquia does not yet recommend that customers update to 2.0. We align our settings and package versions with the [Acquia documentation on docs.acquia.com](https://docs.acquia.com/acquia-cloud/performance/memcached/enable/#configuration-for-drupal-8) as much as possible to make sure we don't have contrasting recommendations from different Acquia products to avoid confusion in the Support ticket queue and to ensure that BLT remains an Acquia supported project. 

Internal testing is still ongoing for the memcache module 2.0 stable release, particularly the new support for cache tags, and as you pointed out there are also settings.php changes that will be required which are not backwards compatible. This also means that the settings recommended for releases prior to beta1 (and currently in BLT 9 and 10) will not work with the stable release. BLT will update the settings in lockstep with the Acquia docs team, as we work very closely with the docs team to make sure our messaging is consistent.

lcatlett:
closing due to response above

PR #3358, opened 2019-02-01
Fixes #3357: Don't use relative path for memcache PSR-4 path registration

### LINKED ISSUE #3357: Fatal error during cache rebuild due to registered MemcacheBackendFactory.php path
My system information:
* Operating system type: Drupal VM
* Operating system version: Debian 9
* BLT version: 9.2.3

When I run this command:
```
drush cr
```

I get the following output:
```
Failed to log error: Warning: stat(): stat failed for phar:///usr/local/bin/drush/bin/modules/contrib/memcache/src/MemcacheBackendFactory.php in TYPO3\PharStreamWrapper\PharStreamWrapper->invokeInternalStreamWrapper() (line 453 of /var/www/ghc09ev/vendor/typo3/phar-stream-wrapper/src/PharStreamWrapper.php).
```

The Drush launcher ist installed as PHAR inside the Drupal VM by default and since the upgrade to 8.6.6/8.6.7 it gives me this error when rebuilding caches via Drush.
Of course this is not a direct BLT bug, but it is caused by the way, BLT registers the memcache PSR4 class path in its `memcache.settings.php` file:

```php
$class_loader->addPsr4('Drupal\\memcache\\', 'modules/contrib/memcache/src');
```
I investigated this a little and found that the following would solve the issue:
```php
$class_loader->addPsr4('Drupal\\memcache\\', DRUPAL_ROOT . '/modules/contrib/memcache/src');
```
I'd appreciate to see this changed soon ;)

Thanx in advance!

hctom:
Fixes #3357 
--------

Changes proposed:
---------

- Use absolute path during memcache PSR-4 class path registration

Steps to replicate the issue:
----------

1. Use Drupal VM with Drush Lanucher installed and Memcache set up for Drupal
2. Run `drush cr` and see the PHP fatal error message described in #3357

Steps to verify the solution:
-----------

1. Use Drupal VM with Drush Lanucher installed and Memcache set up for Drupal
2. Run `drush cr` and see if error mentioned in #3357 is still thrown


marcelovani:
The solution works for me

danepowell:
@hctom BLT takes its memcache configuration directly from the memcache module README: https://cgit.drupalcode.org/memcache/tree/README.txt?h=8.x-2.x#n146

If this is a problem, it seems like it needs to be fixed upstream first.

Additionally, BLT does not currently support or recommend running memcache in a local development environment, and I cannot replicate this error in a Cloud environment.

If you can get this changed upstream, or reproduce it in a Cloud environment, then I would definitely reconsider this PR. Thanks!

danepowell:
I filed an upstream issue and would like to wait and see if there's any traction before merging this: https://www.drupal.org/project/memcache/issues/3038440

danepowell:
@lcatlett what prompted you to add the blocked label to this? Is there anything we should be aware of before merging?

danepowell:
Sounds like we should use `$app_root` to be in line with upstream issue and core standards.

lcatlett:
@danepowell I marked this and all the other memcache related issues as blocked because memcache 2.0 is still not supported or recommended by Acquia (see https://github.com/acquia/blt/issues/3220#issuecomment-437619164 and the internal PF tickets on this). When BLT made changes to these settings previously without coordinating with docs and engineering, it was a nightmare for both BLT and support.

hctom:
I also thought about that, but `DRUPAL_ROOT` is used on many places in the BLT settings files - so I think this should be changed all at once in another ticket/PR

hctom:
@lcatlett this issue occures when using memcache 2.0.0-alpha7, so I guess this doesn't have to be blocked by the stable release upgrade stuff.

PR #3360, opened 2019-02-04
Ability to truncate history on git remotes

nedsbeds:
Changes proposed:
---------
A number of projects using BLT and ACE are taking an extremely long time to deploy once the artefact has been built. 

In projects where large numbers of front end and composer dependencies are present, the number of GIT objects is very high in the ACE repository. Our analysis shows that a large portion of the deployment time is spent simply checking out the repository. 

We have experimented with truncating the history on branches in the ACE repo to remove large number of objects in the GIT database. 

This has led to large reductions in deploy time for a branch. In one example ACSF Drupal 8 build the average deploy time has reduced from 40 minutes to 25 when employing this approach.

-

We have performed testing on a number of PS EMEA projects and reviewed these changes here https://github.com/nedsbeds/blt/pull/1 but are now looking for feedback from the BLT team about this approach  . 

This will not suit every project so the functionality must be triggered through an argument or via the configuration deploy: truncate_history

-

Steps to verify the solution:
-----------

1. Trigger by passing `--truncate-history` to the `blt artifact:deploy`
2. Verify the branch you have deployed only has a single commit on the ACE GIT repo
3. The Github repo will not have been altered.


mikemadison13:
@nedsbeds we made an enhancement in the 10x codebase to try and solve a similar issue, have you seen this PR ? https://github.com/acquia/blt/pull/3271

nedsbeds:
> @nedsbeds we made an enhancement in the 10x codebase to try and solve a similar issue, have you seen this PR ? #3271

@mikemadison13 I have not seen it, no. I shall see if we can benchmark against this change

grasmash:
@nedsbeds Thank you.

Hopefully the referenced change addresses your issue. If it does not, I think we will need to find an approach that is different from the one in this PR. BLT should really not interfere with upstream branches if at all possible. Rather, we should narrow the scope of what it explicitly fetches.

sdelbosc:
@grasmash I am not fully convinced about the comment above but before sharing the arguments I would like to complete testing on the project we are experimenting this right now.

PR #3363, opened 2019-02-06
Fixes #3345 to run db updates even when config strategy is none.

### LINKED ISSUE #3345: Bug: Allow drupal:update when cm.strategy=none
I have a similar issue opened with #3194, where I touched on this at the end of the thread, but the more I thought about this specific issue, this feels like a bug to me... if I set config-management strategy to none, running `$ blt drupal:update` essentially does nothing. That seems wrong to me, so I thought I would open this bug/issue specifically to address that point.

#### User Story:
As a BLT user who uses `cm.strategy = none`, when I run the `$ blt drupal:update` commands, I should still have the benefit of an "updatedb" and potentially a cache clear running.

mikemadison13:
Fixes #3345
--------

Changes proposed:
---------

- moves updb from config-import command into drupal:update command
- refactors the deploy command to call drupal:update directly instead of config-import and toggle modules separately 

Steps to verify the solution:
-----------

1. set config strategy to none
2. confirm db updates execute anyway

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -606,8 +606,7 @@ class DeployCommand extends BltTasks {
     $this->say("Deploying updates to <comment>$multisite</comment>...");
     $this->switchSiteContext($multisite);
 
-    $this->invokeCommand('drupal:config:import');
-    $this->invokeCommand('drupal:toggle:modules');
+    $this->invokeCommand('drupal:update');
 
     $this->say("Finished deploying updates to $multisite.");
   }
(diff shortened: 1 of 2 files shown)


mikemadison13:
@dpagini updb runs a cr as part of the process, so i think that covers that.

the uuid thing is, a bit of a hack. we were running into issues last year where UUID wasn't matching between existing config and current site config. so BLT force writes the config UUID to match. I think it won't matter for database updates.

dpagini:
Perfect! Thanks so much then, makes sense to me!

mikemadison13:
there is, of course, an ongoing debate on if a cr should be applied BEFORE the database updates, and i don't want to step into that bear trap in this refactoring. we can fight that out elsewhere.

PR #3374, opened 2019-02-08
Moved pipelines-deploy from build event to post-deploy event

ionutz2k2:
Changes proposed:
---------

- Move the `pipelines-deploy` command from the `build` event into the `post-deploy` event. 
- The `build` event is used to create the pipelines build branch. At the completion of the build event, the build branch gets pushed to Acquia Git. This means that the `pipelines-deploy` command that creates an ODE/CDE and deploys the build branch to it should be called only after the `build` event, when build branch exists on Acquia Git. Because of reasons I will not detail here, calling `pipelines-deploy` inside the `build` event was possible and it was working, but this will change in the very near future. In order for the Pipelines yaml files that are generated using BLT to continue to work, the change in this PR is needed.

Steps to verify the solution:
-----------

1. Pipelines builds using this yaml file should behave exactly the same as they are expected to


grasmash:
We probably need to emit a message in an update hook to let people know that they’ll need to regenerate their build file.

mikemadison13:
We also need to back port this to 9.2.x and 9.x as well, as it will impact everyone using pipelines.

mikemadison13:
https://github.com/acquia/blt/pull/3380 covers regenerating settings files in BLT 10.x, and I added pipelines to that update.

Issue #3209, opened 2018-11-02
blt drupal:install fails after blt update to 9.2.1


shelane:
My system information:
* Operating system type: Linux (through Docker)
* Operating system version: Debian GNU/Linux 9
* BLT version: 9.2.1

Output of `blt doctor`:
```
+---------------------------+------------------------------------------------+
| Property                  | Value                                          |
+---------------------------+------------------------------------------------+
| %paths.%root              | /var/www/docroot                               |
| %paths.%site              | sites/default                                  |
| %paths.%modules           | sites/all/modules                              |
| %paths.%themes            | sites/all/themes                               |
| %paths.%config-sync       | ../config/default                              |
| alias-searchpaths.0       | /var/www/drush/sites                           |
| blt-version               | 9.2.1                                          |
| composer-version          | Composer version 1.6.3 2018-01-31 16:28:17     |
| config-sync               | ../config/default                              |
| db-driver                 | mysql                                          |
| db-hostname               | db                                             |
| db-name                   | default                                        |
| db-password               | user                                           |
| db-port                   | 3306                                           |
| db-username               | user                                           |
| drupal-settings-file      | sites/default/settings.php                     |
| drupal-version            | 8.6.2                                          |
| drush-alias-files.0       | /var/www/drush/sites/default.site.yml          |
| drush-alias-files.1       | /var/www/drush/sites/www.site.yml              |
| drush-cache-directory     | /home/docker/.drush/cache                      |
| drush-conf.0              | /var/www/vendor/drush/drush/drush.yml          |
| drush-conf.1              | /var/www/drush/drush.yml                       |
| drush-conf.2              | /var/www/docroot/sites/default/local.drush.yml |
| drush-script              | /var/www/vendor/bin/drush                      |
| drush-temp                | /tmp                                           |
| drush-version             | 9.5.2                                          |
| modules                   | sites/all/modules                              |
| php-bin                   | /usr/local/bin/php                             |
| php-conf.1                | false                                          |
| php-os                    | Linux                                          |
| root                      | /var/www/docroot                               |
| site                      | sites/default                                  |
| stacks.drupal-vm.inited   | false                                          |
| stacks.dev-desktop.inited | false                                          |
| themes                    | sites/all/themes                               |
| uri                       | drupal8.docksal                                |
+---------------------------+------------------------------------------------+
+------------------------------------------+--------------------------------------------------------------+
| Check                                    | Problem                                                      |
+------------------------------------------+--------------------------------------------------------------+
[...]

Issue #2632, opened 2018-03-13
Invalid drush alias `@${drush.aliases.ci}` after BLT 9 upgrade


justinlevi:
My system information:
* Operating system type: OSX 10.13
* BLT version: 9.0.3

Output of `blt doctor`:
```
soundslides:d8d justinwinter$ blt doctor
+---------------------------+------------------------------------------------------------+
| Property                  | Value                                                      |
+---------------------------+------------------------------------------------------------+
| %paths.%root              | /Users/justinwinter/Sites/d8d/docroot                      |
| %paths.%site              | sites/default                                              |
| %paths.%modules           | sites/all/modules                                          |
| %paths.%themes            | sites/all/themes                                           |
| %paths.%config-sync       | /Users/justinwinter/Sites/d8d/config/default               |
| %paths.%files             | sites/default/files                                        |
| %paths.%temp              | /tmp                                                       |
| %paths.%private           | /Users/justinwinter/Sites/d8d/files-private                |
| admin-theme               | seven                                                      |
| alias-searchpaths.0       | /Users/justinwinter/Sites/d8d/drush/sites                  |
| blt-version               | 9.0.3                                                      |
| bootstrap                 | Successful                                                 |
| composer-version          | Composer version 1.6.3 2018-01-31                          |
|                           | 16:28:17                                                   |
| config-sync               | /Users/justinwinter/Sites/d8d/config/default               |
| db-driver                 | mysql                                                      |
| db-hostname               | localhost                                                  |
| db-name                   | d8d                                                        |
| db-password               | drupal                                                     |
| db-port                   | 3306                                                       |
| db-status                 | Connected                                                  |
| db-username               | drupal                                                     |
| drupal-settings-file      | sites/default/settings.php                                 |
| drupal-version            | 8.5.0                                                      |
| drush-conf.0              | /Users/justinwinter/Sites/d8d/vendor/drush/drush/drush.yml |
| drush-conf.1              | /Users/justinwinter/Sites/d8d/drush/drush.yml              |
| drush-script              | /Users/justinwinter/Sites/d8d/vendor/bin/drush             |
| drush-temp                | /tmp                                                       |
| drush-version             | 9.2.1                                                      |
| files                     | sites/default/files                                        |
| install-profile           | demo_umami                                                 |
| modules                   | sites/all/modules                                          |
| php-bin                   | /usr/local/Cellar/php71/7.1.10_21/bin/php                  |
[...]

Issue #2078, opened 2017-10-04
SQL-Sanitize Removed from Drush 9


mikemadison13:
My system information:
* Operating system type: [replace]
* Operating system version: [replace]
* BLT version: 8.9.x-dev f1901309db4a999db0e7d54c6936452788ffa2dd

When I run this command:
```
blt sync
```

I get the following output:
```
 [error]  Unknown option: --sanitize.  See `drush help sql-sync` for available options. To suppress this error, add the option --strict=0. 
[Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1  Time 6.713s
[error]   
```

And I expected this to happen:

After digging into this, there is something odd going on with sql-sync in Drush 9. I have opened a bug in their issue queue to track this issue here: https://github.com/drush-ops/drush/issues/3013

mikemadison13:
it would appear the sql-sanitize DOES show up as a valid option when there is a functional site / normal database in place (and does not when there isn't).

PR #3384, opened 2019-02-13
Update drush to 9.6.0, and robo to 1.3.4

### LINKED ISSUE #3350: BLT commands throw errors after updating drush 9.6.0 dependencies
In testing the upcoming drush 9.6.0 stable release, I updated robo to 1.3.4 since Drush 9.6.0 requires robo ^1.3.0. This appears to break all BLT commands, which has been reported before in #3209, #2078, and #2632 but was never addressed since pinning BLT to robo ^1.2.4 offered a temporary workaround. We will need to fix this to leverage the bugfixes and new functionality in drush 9.6.0 and its dependencies such as dynamic aliases and fixes for `sql:` commands. 

When I run this command:
- Any blt command

I get the following output:
```
~/sites/blted8/docroot develop* 3m 24s
❯blt list

Fatal error: Uncaught Error: Call to a member function set() on null in /var/www/vendor/acquia/blt/src/Robo/Config/DefaultConfig.php:70
Stack trace:
#0 /var/www/vendor/acquia/blt/src/Robo/Config/ConfigInitializer.php(58): Acquia\Blt\Robo\Config\DefaultConfig->setSite('default')
#1 /var/www/vendor/acquia/blt/src/Robo/Config/ConfigInitializer.php(84): Acquia\Blt\Robo\Config\ConfigInitializer->setSite('default')
#2 /var/www/vendor/acquia/blt/bin/blt-robo-run.php(29): Acquia\Blt\Robo\Config\ConfigInitializer->initialize()
#3 /var/www/vendor/acquia/blt/bin/blt-robo.php(14): require_once('/var/www/vendor...')
#4 /var/www/vendor/acquia/blt/bin/blt(11): require_once('/var/www/vendor...')
#5 {main}
  thrown in /var/www/vendor/acquia/blt/src/Robo/Config/DefaultConfig.php on line 70```

- blt commands to execute successfully

### PR DESCRIPTION
**Pull request by alexxed**
Fixes #3350 
--------

Changes proposed:
---------
Update drush to 9.6.0. Requires Robo updated to 1.3.x.

Caused by https://github.com/consolidation/Robo/commit/3317048f90b78101619d674514101cd99d9ca0a6#diff-20b8d9734807c101cc969094af11a517.

Steps to replicate the issue:
----------

1. Update to drush 9.6.0
2. Run blt list
3. See that it fails
Steps to verify the solution:
-----------

It's covered by tests. A manual test running `blt list` on this branch would be enough.
```
git clone https://github.com/alexxed/blt.git
rm -rf blted8
cd blt
git checkout 3350
composer update
./vendor/bin/robo create:from-symlink
cd ../blted8
blt list


alexxed:
Fore some reason, after the upgrade, the DrupalSettingsTest is missing a `blt:init:settings` run, so it's failing.

mikemadison13:
@alexxed I wondered if that was happening, I think it may be failing farther up in the bootstrap process if that's the case.

alexxed:
Yeah, blt:init:settings does run, but https://github.com/acquia/blt/blob/c2431dcde15b7aa8e3a2845bd815489f2d8dbfd2/src/Robo/Commands/Setup/SettingsCommand.php#L72 returns empty array on this branch, and ['default'] on 10.0.x. Looking into it.

alexxed:
@lcatlett The use of ->config-> in other cases is legitimate, as Commands for example do have access to the ->config property; that's a Robo Config object. 

In the case of DefaultConfig, which is a Robo Config object, the `config` property used to be a Consolidation Config object. This changed in Robo, you don't have access to the config object anymore because a Robo Config object now has 2 contexts, each with its config object: https://github.com/consolidation/Robo/commit/62e4fc36310feaba9ca19cd2c773299079a63c11#diff-4c10610785e0bfc5ca0d0cf62c134679

The default context is used when importing, the process context when using set(). get() will check first in process and then in default.

The problem with BLT using this new Robo version is that there are cases when set() is called before import. If both set a value, the value is not overwritten on import as BLT expects now.

I looked at several ways to fix this, but the simplest one is to go back to one context that would make BLT work as before in terms of configuration. To do this, we need to override the replace method; I'll update the PR.

mikemadison13:
@alexxed please have a look at the additional items @lcatlett mentioned. i want to make sure we have our bases covered there as well.

alexxed:
@mikemadison13 I did, and mentioned that those cases use $this->config as a RoboConfig object, which has a set method, so those calls are fine.

Another thing I wanted to confirm, should we update drush now, or just robo? drush is still in rc, rc3 now.

mikemadison13:
I would prefer to not update into Drush 9.6.3 until after it goes stable. The question is, does the Robo update work ok if we merge it with 9.5.2 still in our codebase?

@lcatlett does the config removal cover your other locations? I know you called out [ConfigInitializer.php](https://github.com/acquia/blt/blob/10.0.x/src/Robo/Config/ConfigInitializer.php), [DeployCommand.php](https://github.com/acquia/blt/blob/10.0.x/src/Robo/Commands/Deploy/DeployCommand.php), and any relevant tests etc. I'm wondering if @alexxed's change is high enough up in the bootstrap that these don't have to be updated explicitly.

alexxed:
I think the same about waiting for drush, so let me try to remove that and see how tests go.

alexxed:
Updated to latest Robo 1.x, that's 1.4.6 now. Removed updating drush, but I tested that it will work with it.
@mikemadison13  but if we'll not be updating Drush in a major version, e.g. 10.0.x which is already beta, it might make sense to hold off merging this one.

lcatlett:
@alexxed - does the update to robo resolve the issues reported in the [original bug](https://github.com/acquia/blt/issues/3350#issue-403503543) report or can you provide any clarity on the root causes and best way forward to resolve the linked #3209, #2078, and #2632 issues?

alexxed:
Sure, I'll have a look at the linked issues.

danepowell:
It sounds like we are just waiting on @alexxed to confirm that this satisfies the referenced issues, let me know if there's anything else required.

I think we definitely want to update Robo (to 1.4.6) and Drush (to 9.6.0) in 10.x

The only remaining question for me is whether we want to force the Robo update in 9.2.x. Will requiring Robo 1.4.6 break Drush 9.5.x or any other related packages? I can't think of any common packages besides Drush and BLT that depend on Robo...

Issue #3366, opened 2019-02-06
PHPCS is invoked inconsistently between blt validate and git pre-commit hooks


malikkotob:
My system information:
* Operating system type: macOS
* Operating system version: Mojave 10.14
* BLT version: 9.2.3

Output of `blt doctor`:
```
vagrant@local:/var/www/project$ blt doctor
+---------------------------+----------------------------------------------------+
| Property                  | Value                                              |
+---------------------------+----------------------------------------------------+
| %paths.%root              | /var/www/project/docroot                               |
| %paths.%site              | sites/default                                      |
| %paths.%modules           | sites/all/modules                                  |
| %paths.%themes            | sites/all/themes                                   |
| %paths.%config-sync       | /var/www/project/config/default                        |
| %paths.%files             | sites/default/files                                |
| %paths.%temp              | /tmp                                               |
| %paths.%private           | /var/www/project/files-private                         |
| admin-theme               | seven                                              |
| alias-searchpaths.0       | /var/www/project/drush/sites                           |
| blt-version               | 9.2.3                                              |
| bootstrap                 | Successful                                         |
| composer-version          | Composer version 1.8.2 2019-01-29 15:00:53         |
| config-sync               | /var/www/project/config/default                        |
| db-driver                 | mysql                                              |
| db-hostname               | localhost                                          |
| db-name                   | drupal                                             |
| db-password               | drupal                                             |
| db-port                   | 3306                                               |
| db-status                 | Connected                                          |
| db-username               | drupal                                             |
| drupal-settings-file      | sites/default/settings.php                         |
| drupal-version            | 8.6.7                                              |
| drush-alias-files.0       | /var/www/project/drush/sites/project.site.yml              |
| drush-cache-directory     | /home/vagrant/.drush/cache                         |
| drush-conf.0              | /var/www/project/vendor/drush/drush/drush.yml          |
| drush-conf.1              | /var/www/project/drush/drush.yml                       |
| drush-conf.2              | /var/www/project/docroot/sites/default/local.drush.yml |
| drush-script              | /var/www/project/vendor/bin/drush                      |
| drush-temp                | /tmp                                               |
| drush-version             | 9.5.2                                              |
| files                     | sites/default/files                                |
| install-profile           | project_lightning                                      |
| modules                   | sites/all/modules                                  |
| php-bin                   | /usr/bin/php7.1                                    |
| php-conf.0                | /etc/php/7.1/cli/php.ini                           |
| php-os                    | Linux                                              |
[...]

PR #3386, opened 2019-02-14
Adding Scans for Deprecated Function in Core

aweingarten:
This PR will scan custom code in the standard locations for deprecated functions and fail the build if they are found.  This will help the progression to D9! 

It should the same patterns as PHPCS or Symfony Security Scans

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,10 +55,12 @@
     "require-dev": {
         "composer/composer": "^1.6.4",
         "knplabs/github-api": "^2.6",
+        "mglaman/phpstan-drupal": "^0.11.1",
         "php-http/guzzle6-adapter": "^1.1",
-        "webflo/drupal-core-require-dev": "^8.6.0",
+        "phpstan/phpstan": "^0.11.1",
+        "phpstan/phpstan-deprecation-rules": "^0.11.0",
         "phpunit/phpunit": "^4.8|^6.5",
-        "squizlabs/php_codesniffer": "^3.0.1"
+        "squizlabs/php_codesniffer": "^3.0.1",
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "phpstan/phpstan": "^0.11.1",
         "phpstan/phpstan-deprecation-rules": "^0.11.0",
         "phpunit/phpunit": "^4.8|^6.5",
-        "squizlabs/php_codesniffer": "^3.0.1",
+        "squizlabs/php_codesniffer": "^3.0.1"
     },
     "autoload": {
         "psr-4": {
--- a/config/build.yml
+++ b/config/build.yml
@@ -239,6 +239,11 @@ validate:
     filters: { }
     # Add any custom Twig functions for linter to ignore.
     functions: { }
+  phpstan:
+    filesets:
+    - files.php.custom.modules
+    - files.php.custom.themes
+    - files.php.tests
   yaml:
     filesets:
       - files.yaml
--- a/src/Robo/Commands/Validate/PHPStanCommand.php
+++ b/src/Robo/Commands/Validate/PHPStanCommand.php
@@ -1,0 +1,141 @@
+<?php
+
+namespace Acquia\Blt\Robo\Commands\Validate;
+
+use Acquia\Blt\Robo\BltTasks;
+use Acquia\Blt\Robo\Exceptions\BltException;
+
+/**
+ * Defines commands in the "tests:phpstan:sniff:all*" namespace.
+ */
+class PhpstanCommand extends BltTasks {
+
+  /**
+   * Executes PHPStan: Static Analysis Tool against all phpstan.filesets files.
+   *
+   * By default, these include custom themes, modules, and tests.
+   *
+   * @command tests:phpstan:sniff:all
+   *
+   * @aliases tptsa phpstan tests:phpstan:sniff validate:phpstan
+   */
+  public function sniffFileSets() {
+    /** @var \Acquia\Blt\Robo\Filesets\FilesetManager $fileset_manager */
+    $fileset_manager = $this->getContainer()->get('filesetManager');
+    $fileset_ids = $this->getConfigValue('validate.phpstan.filesets');
+    $filesets = $fileset_manager->getFilesets($fileset_ids);
+    $bin = $this->getConfigValue('composer.bin');
+    $repo_root = $this->getConfigValue('repo.root');
+    $stringOfFiles = $this->fileSetToList($filesets);
+
+    $command = "$bin/phpstan analyse $stringOfFiles --configuration=$repo_root/phpstan.neon";
+
+    $result = $this->taskExecStack()
+      ->dir($this->getConfigValue('repo.root'))
+      ->exec($command)
+      ->run();
+    $exit_code = $result->getExitCode();
+    if ($exit_code) {
+      throw new BltException("Execution of PHPStan failed.");
+    }
+  }
+
+  /**
+   * Executes PHPStan against a list of files, if in phpstan.filesets.
+   *
+   * This command will execute PHPStan against a list of files if those
+   * files are a subset of the phpstan.filesets filesets.
+   *
+   * @command tests:phpstan:sniff:files
+   * @aliases tptsf
+   *
+   * @param string $file_list
+   *   A list of files to scan, separated by \n.
+   *
+   * @return int
+   */
+  public function sniffFileList($file_list) {
+    $repo_root = $this->getConfigValue('repo.root');
+
+    $this->say("Sniffing directories containing changed files...");
+    $files = explode(" ", $file_list);
+    if ($files) {
+      $command = "analyse $file_list --configuration=$repo_root/phpstan.neon";
+      $exit_code = $this->doSniff($command);
+
+      return $exit_code;
+    }
+
+    return 0;
+  }
+
+  /**
+   * Executes PHPStan against (unstaged) modified or untracked files in repo.
+   *
+   * This command will execute PHPStan against modified/untracked files
+   * if those files are a subset of the phpstan.filesets filesets.
+   *
+   * @command tests:phpstan:sniff:modified
+   * @aliases tpsm
+   *
+   * @return int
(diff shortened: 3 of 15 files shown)


mikemadison13:
@aweingarten might make sense to add this to the validate:all (similar to how we did the security one) so it runs as part of the standard validation process, then we can disable parts of it.

mikemadison13:
also, IMHO, this should be disabled by default.

src/Robo/Commands/Validate/PHPStanCommand.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Converts fileset to a string of all the files.
+     *
+     * @param \Symfony\Component\Finder\Finder[] $filesets
+     *
+     * @return string | NULL
+     *   A space delimeted string of files.
+     */
+    protected function fileSetToList(array $filesets) {
```
aweingarten: Does this make sense as a more global helper function?

aweingarten:
1: Added this to the validate all call.
2: Dumb question how do I disable this off by default?  Is that in the build.yml file? 
like by adding `disable-targets.tests:phpstan`

mikemadison13:
@aweingarten i'm ok with it running, as long as it's not scanning files. so maybe just comment out the file sets by default?

mikemadison13:
FYI looks like phpcs validation failures on the build failure

aweingarten:
PHPCS addressed

aweingarten:
@mikemadison13 okay I think all the feedback has been addressed.  How do you want this to be socialized/documented?

mikemadison13:
honestly, we don't have a section on validation in the docs (that i see). we probably need to refactor the testing docs to better differentiate between automated testing and validation that occurs in blt. let me think on this a bit and get back to you.

aweingarten:
Sounds good.  I am thinking that we should also probably think about adding it to the the release notes so that people start enabling it.  

Also once we get this turned on for deprecation, increasing the sensitivity of PHPstan to Level 2-7 scans is also freaking awesome.  It's really worth considering nudging people as we head towards D9. Checkout the results of the scan for Acquia connector.  

[acquia-connector-scan.txt](https://github.com/acquia/blt/files/2871187/acquia-connector-scan.txt)

aweingarten:
So @mikemadison13 I'm an idiot.  There is an obvious answer here.  What if we make this work like PHPCS where we only scan the files that have _changed_ rather than all the things?  That way we gently make the world a better place without immediate whiplash?

mikemadison13:
@aweingarten i really like that idea, although i would be interested in benchmarking it. i do think that creates some disparity between "why are my changed files getting deprecation scans" and my others ones aren't.

i do think we need this to be on globally or off globally, so maybe the commit scan needs to respect the same rules as the normal scan (just within the changed fileset)

aweingarten:
So anecdotally the time for our custom files is under 30 seconds.  As for the disparity it already exists for PHPCS so I think we are okay.

mikemadison13:
it does, but re: https://github.com/acquia/blt/issues/3366 i'm trying to change that since commit hooks find different things than blt validate and that is driving me batty.

composer.json
```diff
@@ -50,7 +50,10 @@
         "symfony/yaml": "^3.2.8",
         "tivie/php-os-detector": "^1.0",
         "typhonius/acquia-php-sdk-v2": "^1.0.3",
-        "zaporylie/composer-drupal-optimizations": "^1.0"
+        "zaporylie/composer-drupal-optimizations": "^1.0",
```
aweingarten: Reorder this one alphabetically with the rest of the packages

template/phpstan.neon
```diff
@@ -0,0 +1,18 @@
+parameters:
+  customRulesetUsed: true
+  memory_limit: -1
```
aweingarten: Probably set this to a more sane value

template/phpstan.neon
```diff
@@ -0,0 +1,18 @@
...
+  reportUnmatchedIgnoredErrors: false
+  # Ignore phpstan-drupal extension's rules.
+  excludes_analyse:
+  - blazy_test.module
```
aweingarten: Add comment here for why

config/build.yml
```diff
@@ -225,6 +225,11 @@ validate:
     filters: { }
     # Add any custom Twig functions for linter to ignore.
     functions: { }
+  phpstan:
```
aweingarten: Clean up the commit message here.

src/Robo/Commands/Validate/PHPStanCommand.php
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Acquia\Blt\Robo\Commands\Validate;
+
+use Acquia\Blt\Robo\BltTasks;
+use Acquia\Blt\Robo\Exceptions\BltException;
+
+/**
+ * Defines commands in the "validate:*" namespace.
+ */
+class PHPStanCommand extends BltTasks {
```
aweingarten: Capitalization should match the file

alexxed:
What if on the first run, you generate exclusions based on current state of things so that the developer can work on that in his own time, but have phpstan run for any new code he adds. You can use ignoreErrors in phpstan.neon with regex and the filename to be exact.

On the next run, just show a single warning that there are ignoreErrors to be dealt with.

aweingarten:
@alexxed that's a really interesting idea. 
That being said we solved this years ago when we started using phpcs.  Back then most of our codebases bombed out on PHPCS.  It was gloriously bad. The solution back then, which is what we do today, was by default to only scan the files that were changed on commit or in a PR.  

It's a `you touch it you own it and need to make it better` approach.   BLT users have already been trained on this and it will come naturally to them.

tests/phpunit/Blt/PhpStanTest.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+class PhpStanTest extends BltProjectTestBase {
+
+  /**
+   * @group blt
+   *
+   * @dataProvider providerPhpStanFilesBootstrapProvider
+   */
+  public function testPhpStanFilesBootstrap($filename, $needle, $contains) {
+    $process = new Process("./vendor/bin/phpstan analyse $filename --configuration=phpstan.neon");
```
grasmash: We need to run the actual BLT command in testing.

composer.json
```diff
@@ -55,10 +55,13 @@
...
         "knplabs/github-api": "^2.6",
+        "mglaman/phpstan-drupal": "^0.11.1",
         "php-http/guzzle6-adapter": "^1.1",
-        "webflo/drupal-core-require-dev": "^8.6.0",
+        "phpstan/phpstan": "^0.11.1",
+        "phpstan/phpstan-deprecation-rules": "^0.11.0",
         "phpunit/phpunit": "^4.8|^6.5",
-        "squizlabs/php_codesniffer": "^3.0.1"
+        "squizlabs/php_codesniffer": "^3.0.1",
+        "webflo/drupal-core-require-dev": "^8.6.0"
```
grasmash: Why is this necessary?

src/Robo/Commands/Validate/PHPStanCommand.php
```diff
@@ -0,0 +1,141 @@
...
+   * if those files are a subset of the phpstan.filesets filesets.
+   *
+   * @command tests:phpstan:sniff:modified
+   * @aliases tpsm
+   *
+   * @return int
+   */
+  public function sniffModified() {
+    $this->say("Sniffing modified and untracked files in repo...");
+    $arguments = "--filter=gitmodified " . $this->getConfigValue('repo.root');
```
grasmash: `--filter=gitmodified`

I'm a bit concerned this may not catch exactly the files we care about. Can we test that the right files  are scanned?

danepowell:
The disconnect between pre-commit validation and `blt validate` is jarring enough already, I don't want to add this in a way that tests might pass `blt validate` but fail during pre-commit.

Rather, I'd like to see this run alongside all of the other validations during `blt validate`.

In 10.0.x, it can be enabled by default, _if_ we can get this PR in before 10.0.0 drops. But it should have a message clearly indicating how to disable it when/if it fails.

In 9.2.x, it should be disabled by default (if we decide to backport it to 9.2.x at all).

aweingarten:
@danepowell
Thanks for the feedback!  I added some info to the error message for how to turn off the scans! 

> The disconnect between pre-commit validation and blt validate is jarring enough already, I don't want to add this in a way that tests might pass blt validate but fail during pre-commit.  Rather, I'd like to see this run alongside all of the other validations during blt validate.

Can you elaborate on this?  Let me know what you are looking for here.  _I thought_  that the phpstan scans were built into `blt validate`.   So the PR is intended to replicate the PHPCS behavior.  Are you requesting that we also refactor PHPCS functionality in this PR?

composer.json
```diff
@@ -55,10 +55,12 @@
...
         "composer/composer": "^1.6.4",
         "knplabs/github-api": "^2.6",
+        "mglaman/phpstan-drupal": "^0.11.1",
         "php-http/guzzle6-adapter": "^1.1",
-        "webflo/drupal-core-require-dev": "^8.6.0",
+        "phpstan/phpstan": "^0.11.1",
+        "phpstan/phpstan-deprecation-rules": "^0.11.0",
         "phpunit/phpunit": "^4.8|^6.5",
-        "squizlabs/php_codesniffer": "^3.0.1"
+        "squizlabs/php_codesniffer": "^3.0.1",
```
danepowell: Think this trailing comma isn't valid JSON

danepowell:
The problem with PHPCS is that if you pass it a list of files to check (i.e. during git pre-commit hook), it will _always_ scan those files even if they wouldn't otherwise be scanned according to the defined filesets in BLT and/or phpcs.xml.dist.

As long as this filters the file list (i.e. scans an intersection between git changed files and the phpstan fileset) it should be fine.

danepowell:
Looks like tests are still failing, and there are still some comments from Matt to address.

aweingarten:
@danepowell I should have time this weekend to devote to this PR!

aweingarten:
Closing given the updated direction in BLT unsure if this is an appropriate place to commit this.

PR #3387, opened 2019-02-14
Create 10.x branch alias.

grasmash:
Create a branch alias from 10.x-dev to 10.0.x-dev.


dpagini:
Totally just asking out of curiosity... what does this change _do_?

grasmash:
Right now, you need to include BLT 10.x via `composer require acquia/blt:10.0.x-dev`.This allows you to use `composer require acquia/blt:10.x-dev` and still point to the 10.0.x branch.

mikemadison13:
@grasmash, i think we are good to merge this? @TravisCarden is willing to test once it's in.

PR #3394, opened 2019-02-20
Add lando support

raghunat:
# DO NOT MERGE

> I am still validating that the common BLT functionality will still work. I mostly likely still have some templating to fix up.

Changes proposed: 
---------
- Add a recipe for creating lando config that would allow lando and BLT to work together

Steps to verify the solution:
-----------
1. `blt lando`
2. Answer yes a few times
3. `lando blt` to see blt running within the container

### DIFF
--- a/scripts/lando/lando.site.yml
+++ b/scripts/lando/lando.site.yml
@@ -1,5 +1,5 @@
 local:
-  root: /var/www/${project.machine_name}/docroot
+  root: /app/docroot
   uri: http://${project.machine_name}.lndo.com
   host: ${project.machine_name}.lndo.com
   user: root
--- a/scripts/lando/lando.site.yml
+++ b/scripts/lando/lando.site.yml
@@ -1,0 +1,5 @@
+local:
+  root: /var/www/${project.machine_name}/docroot
+  uri: http://${project.machine_name}.lndo.com
+  host: ${project.machine_name}.lndo.com
+  user: root
--- a/scripts/lando/landofile.yml
+++ b/scripts/lando/landofile.yml
@@ -1,0 +1,9 @@
+name: ${project.machine_name}
+recipe: drupal8
+config:
+  webroot: docroot
+
+tooling:
+  blt:
+    service: appserver
+    cmd: /usr/bin/blt
--- a/src/Robo/Commands/Lando/LandoCommand.php
+++ b/src/Robo/Commands/Lando/LandoCommand.php
@@ -1,0 +1,248 @@
+<?php
+
+namespace Acquia\Blt\Robo\Commands\Lando;
+
+use Acquia\Blt\Robo\BltTasks;
+use Acquia\Blt\Robo\Common\ArrayManipulator;
+use Acquia\Blt\Robo\Exceptions\BltException;
+use function file_exists;
+use function file_get_contents;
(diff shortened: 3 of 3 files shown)


mikemadison13:
as this progresses @raghunat it would be good to include doc updates to https://github.com/acquia/blt/blob/10.0.x/docs/alternative-environment-tips/lando.md to ensure we have the most accurate coverage.

aweingarten:
Hey question, what is the business value of this PR?  DrupalVM is super-stable and more performant in most cases than Lando.  

There is a Drupal Core Initiative to come up with an official local development environment.    I'm not sure what the latests is:
https://www.drupal.org/project/ideas/issues/2965681

However it may make sense to punt on this till there is a community decision or to attack this with the community rather than in BLT?

aweingarten:
I did some digging it looks like the community is heading for DDEV
https://www.drupal.org/project/official_docs/issues/2993153

lcatlett:
Is there a compelling reason to manage this in BLT and not contribute a BLT recipe or plugin to lando as has been done with the [docksal addon](https://github.com/docksal/addons/tree/master/blt)?

raghunat:
@mikemadison13 A great point, I will make sure docs are pristine 👍 

@aweingarten I don't think this is an "either or" situation, but a "Open set of choices" approach. Open to feedback here!

@lcatlett This seemed the best medium to start with, given BLT already solved the push, pull, automate, and other tasks. I am open to thoughts though!

I am hoping to get back to this tonight :-)

aweingarten:
@raghunat so having BLT be opinionated here on DrupalVM has been incredibly valuable.  Before we standardized on DrupalVM, when people had choices, we had major problems with small differences in environments derail developers for 3 days and force architects to intervene.  

After we standardized we could expect that entire teams were using the same solution, it also meant that we could move people between projects and expect them to produce code within 24 hours.   Finally when one team hit a problem with DrupalVM we could fix it for everyone.  

Having some BLT teams on Lando and some on DDEV is going to cause support and maintenance issues for Architects/Dev Leads, and the maintainers of BLT. 

Unless there is a compelling reason to support DrupalVM, DDEV and Lando.  I'd recommend deprecating DrupalVM and picking DDEV to preserve the best DX and align with the community.

alexxed:
While I agree on being opinionated, I think we should support at least one Docker based alternative.

lcatlett:
@raghunat - this PR wont be in a reviewable state until the BLT travis build passes and it has automated tests. Take a look at https://github.com/acquia/blt/blob/10.0.x/docs/CONTRIBUTING.md which reviews running the [build/release robo tasks](https://travis-ci.org/acquia/blt/jobs/496014355#L568) locally, which should enable you to automatically test and fix errors such as as the phpcs violations currently [causing build failures](https://travis-ci.org/acquia/blt/jobs/496014355#L583). 

Are you looking for feedback about recommended remaining action items? It isnt clear if this is the first of many PRs to add a docker-based solution to BLT, or if it it is intended to be complete.

raghunat:
Hello All! A few peers of mine have gotten together and believe we may or may not suggest an alternative approach. Until we get a good feeling on where to go with this one, I'll close this, and reopen if we have the right path forward. Thanks!

PR #3397, opened 2019-02-20
Revert "Create 10.x branch alias."

mikemadison13:
Reverts acquia/blt#3387

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -71,8 +71,7 @@
         "composer-exit-on-patch-failure": true,
         "class": "Acquia\\Blt\\Composer\\Plugin",
         "branch-alias": {
-            "dev-master": "10.0.x-dev",
-            "10.x-dev": "10.0.x-dev"
+            "dev-master": "10.0.x-dev"
         },
         "installer-paths": {
             "build/docroot/core": ["type:drupal-core"],


danepowell:
@mikemadison13 what's the backstory here? Why revert it?

mikemadison13:
@danepowell the goal here was to create an alias that would fix the 10.0.x branch name for @TravisCarden. The original PR didn't seem to hurt anything, but also didn't seem to fix it. I put in a revert PR but @grasmash said to hold off on merging it. Not sure where we are. I think it should come out but defer to others.

TravisCarden:
The original issue, @danepowell, was that I reported that it should be possible to `composer require acquia/blt:10.x-dev` (as opposed to `10.0.x-dev`). This is true as a matter of semantic versioning theory, of course, but in practice, ORCA needs to `require` the latest commit with a major version of 10, and it would be preferable not to have to manually update the code every time BLT bumps a minor version number.

danepowell:
It sounds like there's no harm in reverting this since it didn't work in the first place. @grasmash is there any reason we shouldn't merge this?

grasmash:
There's no harm it reverting it, but it would be even better to make it work.

grasmash:
Looks like the `dev-master` alias also is not working.

grasmash:
Maybe this will fix it: https://github.com/acquia/blt/pull/3432/files

PR #3408, opened 2019-02-27
Replace includes.php $data by (stable) global variable equivalent

### LINKED ISSUE #3362: ACSF Multi-Site Deploy Failure
My system information:
* BLT version: 9.2.3

During AFSC Deploy

I get the following output:
```
Notice: Undefined offset: 3 in /mnt/www/html/<project>/vendor/acquia/blt/scripts/blt/drush/cache.php on line 11
Error: Not enough arguments. Site, environment, and uri are required..
```

Deploy to succeed.

This appears related to updates made here https://github.com/acquia/blt/pull/3166 by @lcatlett.

After discussing with her, this may be an issue. If so, we will likely need to port forward to 10.x.

I don't understand why the URI isn't defined here: https://github.com/acquia/blt/blob/9.2.x/scripts/blt/drush/cache.php#L15 which is being called from https://github.com/acquia/blt/blob/9.2.x/scripts/factory-hooks/db-update/db-update.sh#L35

@grasmash  this is related to work you were involved in, so we should discuss.

rmuit:
Fixes #3362
--------

Changes proposed:
---------
In factory-hooks/post-sites-php/includes.php, change the usage of $data['gardens_site_settings'] into $GLOBALS['gardens_site_settings'].

At the same time, remove the PHPCS indicator that was only added because $data was used in this file.

Reasons: 
* The ACSF Connector module's sites.php, in their version *.x-*.52 has stopped using $data, along with several other variables.
* There was no indication that (non-Drupal specific) variables in the local namespace of sites.php could just be used. On the other hand, it is hopefully self evident that we will never just change the contents of the the global $gardens_site_settings variable without notification. A global variable is (explicitly meant to be) used in other places in the Drupal codebase.

_Please note:_ since $GLOBALS['gardens_site_settings'] has 'always' existed, this should be backportable without any issue to every 9.x version. (This includes.php was introduced in an early alpha version of 9.x)

Steps to replicate the issue:
----------
Can be found in a.o. [ internal tickets, known by lcatlett]. I hope you will forgive me for not having exact steps and not even reproducing the error.

But note that the issue means that includes.php does not define global $_acsf_site_name - I don't know BLT but from my viewpoint that can have any kind of effects.

Steps to verify the solution:
-----------
* Reproduce whatever issues those tickets (or any others that you know of) had.
* Notice that the PHP Notices about undefined $data and undefined indexes disappear.
* Run PHPCS (your own standards) on includes.php and check that there are no issues.

### DIFF
--- a/scripts/factory-hooks/post-sites-php/includes.php
+++ b/scripts/factory-hooks/post-sites-php/includes.php
@@ -5,8 +5,6 @@
  * ACSF post-sites-php hook.
  *
  * @see https://docs.acquia.com/site-factory/tiers/paas/workflow/hooks
- *
- * phpcs:disable DrupalPractice.CodeAnalysis.VariableAnalysis
  */
 
 // The function_exists check is required as the file is included several times.
@@ -54,7 +52,7 @@ if (!function_exists('gardens_data_get_sites_from_file')) {
 
 }
 // Get all the domains that are defined for the current site.
-$domains = gardens_data_get_sites_from_file($data['gardens_site_settings']['conf']['acsf_db_name']);
+$domains = gardens_data_get_sites_from_file($GLOBALS['gardens_site_settings']['conf']['acsf_db_name']);
 // Get the site's name from the first domain.
 global $_acsf_site_name;
 $_acsf_site_name = explode('.', array_keys($domains)[0])[0];


lcatlett:
@rmuit - thank you for submitting this PR! could you please remove the internal ticket number DG-20990 and PF-1795 when you get a chance?

rmuit:
Removed.

mikemadison13:
@rmuit thank you so much for this!

am i reading you correctly that this will only work with ACSF version 2.52? perhaps we should pin to that version as part of this PR as well if so?

rmuit:
Nope. I should have said this  more clearly, but "since $GLOBALS['gardens_site_settings'] has 'always' existed," 
* it is not only backportable to BLT 9.x
* but also it works with 'every' past ACSF versions (at least years back)

mikemadison13:
oh awesome. glad i asked then! thanks

mikemadison13:
this definitely fixes the missing "$data" warnings I was seeing recently in factory hook output. so that's a good sign.

PR #3412, opened 2019-02-28
Improve logging ex and set explicit exit conditions.

### LINKED ISSUE #3325: db-update.sh factory hook doesn't output an error code when drupal:update has failed
The db-update.sh factory hook is currently failing on my ACSF instance.

The log output is:

```
[staging-2364|26825] - The db-update.sh script has completed: Completed in 1 seconds. Exit code: 0. Standard out: Running BLT deploy tasks on championstennis.dev-endeavorsf1.acsitefactory.com domain in 01dev environment on the endeavorsf1 subscription. > drupal:config:import > drupal:toggle:modules 0.898s total time elapsed.; Standard error: [info] Skipped pre-config-import target hook. No hook is defined. [warning] BLT will NOT import configuration, /mnt/www/html/endeavorsf101dev/docroot/../config/default/core.extension.yml was not found. [info] modules.01dev.enable is not set. [info] modules.01dev.uninstall is not set.
```

This message clearly shows there is a problem. However, the log level in the ACSF instance is 'INFO'. That's obviously wrong -- something has failed.

I filed a ticket with Acquia support (https://insight.acquia.com/support/tickets/760236), and the response was that this was correct behaviour for the ACSF platform:

> The work process runs an external script (db-update.sh), which emits a "warning" message while it runs. However, as this script runs and exits without returning an error message, the Site Factory task log displays the script's output as an "info" level message.

So therefore the problem is that db-update.sh should be exiting with an error code when its command fails.

lcatlett:
Fixes #3325  
--------

Changes proposed:
---------
- Properly define exit conditions and enable verbose log / per-line output for db-update factory hook tasks

Steps to replicate the issue:
----------
1. View log in ACSF of platform update/release with known failure
2. Observe mangled log output

Steps to verify the solution:
-----------
1. Run platform code update/release
2. Observe improved log output ux

### DIFF
--- a/scripts/factory-hooks/db-update/db-update.sh
+++ b/scripts/factory-hooks/db-update/db-update.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Factory Hook: db-update
 #
@@ -9,6 +9,10 @@
 #
 # Usage: post-code-deploy site env db-role domain custom-arg
 # Map the script inputs to convenient names.
+
+# Exit immediately on error and enable verbose log output.
+set -ev
+
 # Acquia hosting site / environment names
 site="$1"
 env="$2"
@@ -25,7 +29,7 @@ blt="/mnt/www/html/$site.$env/vendor/acquia/blt/bin/blt"
 # locate the URI based on the site, environment and db role arguments.
 uri=`/usr/bin/env php /mnt/www/html/$site.$env/hooks/acquia/uri.php $site $env $db_role`
 
-# Create array with site name fragments from ACSF uri. 
+# Create array with site name fragments from ACSF uri.
 IFS='.' read -a name <<< "${uri}"
 
 # Create and set Drush cache to unique local temporary storage per site.
@@ -41,4 +45,6 @@ echo "Generated temporary drush cache directory: $cacheDir."
 echo "Running BLT deploy tasks on $uri domain in $env environment on the $site subscription."
 
 DRUSH_PATHS_CACHE_DIRECTORY=$cacheDir $blt drupal:update --environment=$env --site=${name[0]} --define drush.uri=$domain --verbose --yes --no-interaction
+
+set +v
 


mikemadison13:
@lcatlett do you have an example of what the updated log output might look like? i'd like to test this.

scripts/factory-hooks/db-update/db-update.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
danepowell: Is there a reason to change this? Unless ACSF is planning on moving the bash binary I can't see why we'd want to change it.

scripts/factory-hooks/db-update/db-update.sh
```diff
@@ -42,3 +46,5 @@ echo "Running BLT deploy tasks on $uri domain in $env environment on the $site s
 
 DRUSH_PATHS_CACHE_DIRECTORY=$cacheDir $blt drupal:update --environment=$env --site=${name[0]} --define drush.uri=$domain --verbose --yes --no-interaction
 
+set +v
+
```
danepowell: Don't need a newline here.

scripts/factory-hooks/db-update/db-update.sh
lcatlett: the use of env here is not to launch the interpreter, though it does allow for a more portable script by using the first bash executable in the user's path as opposed to making assumptions about the location of bash. env is instead used to allow the drush script / user to set environment variables such as `DRUSH_PATHS_CACHE_DIRECTORY` without resetting the current environment

`#!/bin/bash` was flagged by my local shellcheck as an error and it just occurred to me that BLT isnt running shellcheck validation on scripts..that would probably catch a lot of issues and enforce best practices like this one (see https://github.com/koalaman/shellcheck/wiki/SC2239#rationale)

danepowell: I don't think there's anything wrong with using `#!/bin/bash` on a known system, which is exactly what ACSF is. And in fact, I think it's preferable from a security and stability standpoint, since a bad or incompetent actor could change an environment variable to run any arbitrary script if we used `#!/usr/bin/env bash`

Additionally, I think there are limitations to `#!/usr/bin/env` such as the [inability to accept multiple arguments](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash).

> env is instead used to allow the drush script / user to set environment variables such as DRUSH_PATHS_CACHE_DIRECTORY without resetting the current environment

You should have no problem setting these variables if the script is invoked via `#!/bin/bash`. I can't find any reference indicating that this limits how you can set environment variables.

danepowell:
Re-rolled as #3443

PR #3420, opened 2019-03-06
Move ACSF validate command to default config.

dpagini:
Just thought this was a tad bit confusing to see this setting on my latest project. For most ACSF stuff, we use the BLT command to "init" that within a repo. Maybe that `recipes:acsf:init:all` command could add this to project's config file?

Anyways, I thought this would make more sense to give this setting a default value within the library config, and that can be discovered and overridden in project specific config if a user wishes.

This isn't a big deal to me if it was done on purpose, I just thought this was a suggestion to be more inline with the treatment of other default config.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -211,6 +211,9 @@ tests:
     url: http://127.0.0.1:${tests.server.port}
 
 validate:
+  # You can change this to true to have blt automatically validate acsf-init.
+  acsf: false
+
   deprecated:
     filesets:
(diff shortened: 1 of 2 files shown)


mikemadison13:
I think this PR makes sense. it should be backported to 9.2.x as well.

dpagini:
Thanks for the feedback @mikemadison13 - do you need me to post a similar 9.2 PR or do you just cherry pick the commits all at one time?

Issue #3010, opened 2018-08-14
drush.yml config paths are wrong


TravisCarden:
The config paths in [template/drush/drush.yml](https://github.com/acquia/blt/blob/9.1.1/template/drush/drush.yml) are relative to the file, but Drush interprets them relative to the current working directory; so the referenced files are only found and interpreted if you happen to be invoking Drush from the `drush/` directory. Since I almost always work from either the repository root or within the VM, the below change mostly covers the bases for me; but it's just a bandaid. Anyone have any thoughts on a more scalable solution?

```diff
--- a/drush/drush.yml
+++ b/drush/drush.yml
@@ -5,9 +5,11 @@ drush:
   paths:
     config:
       # Load a drush.yml configuration file from the current working directory.
-      - ../docroot/sites/default/local.drush.yml
+      - docroot/sites/default/local.drush.yml
+      - /var/www/myproject/docroot/sites/default/local.drush.yml
       # Allow local global config overrides.
-      - local.drush.yml
+      - drush/local.drush.yml
+      - /var/www/myproject/drush/local.drush.yml
     include:
       - '${env.home}/.drush'
       - /usr/share/drush/commands
```

lcatlett:
I have never had this issue with the BLT default drush config so I wonder if it is a drush bug. 

It is my understanding that your default project `drush.yml` should go in blt's repo.root/drush directory, then you can include individual paths like you said to the `local.drush.yml` config for each multisite in the config section either hard coded or through an environment variable via env subtitution such as ${env.name}. If I set a `SITE` environment variable in the shell, the drush.yml in the project's drush directory could use the value of SITE through something like the below:

````
# This file defines drush configuration that applies to drush commands
# for the entire application. For site-specific settings, like URI, use
# ../docroot/sites/[site]/drush.yml
drush:
  paths:
    config:
      # Load a drush.yml configuration file from the current working directory.
     - ../docroot/sites/default/local.drush.yml
      # Allow local global config overrides.
      - local.drush.yml
      - '../docroot/sites/${env.site}/local.drush.yml'
    include:
      - '${env.home}/.drush'
      - /usr/share/drush/command
```

TravisCarden:
Everything you articulated is essentially correct, @lcatlett, but beside my point. The config paths are interpreted relative to the current working directory of your terminal session, not the location of the `drush.yml` file. Therefore relative path specifications depend on you being in a certain directory and cease to work if you `cd` out of it. **And that's the problem I'm highlighting.**

[...]

Issue #2996, opened 2018-08-05
Automatically create sites.php


lcatlett:
As a BLT user, when I create a multisite with the `blt multisite` task, sites.php should automatically be created so that sites in the docroot/sites directory are recognized by Drupal core and drush as valid multisites.

lcatlett:
closed with #2997

lcatlett:
Reopening due to #2997 being reverted.

ba66e77:
As discussed in #3031, we can't do this with ACSF sites as on deploy ACSF checks to ensure the sites.php in the codebase matches that provided by the module. So whatever solution is implemented will need to take that into consideration.

lcatlett:
@ba66e77 settings.php can be used for this, ideally in a pre-settings-php hook and the `skip-default-settings` flag can be used as well if modifying sites.php locally is an issue.

danepowell:
Let's consolidate discussion around the management of sites.php in #3477

PR #3422, opened 2019-03-06
Remove unreliable drush setting.

dpagini:
I commented on this in #3010 when I found some issues with the drush config BLT provides.

As @TravisCarden points out in that issue, the include config.paths are relative to the drush command working directory. So, in the lines coming out here, these config files are loaded:
* a "relative" `local.drush.yml` will only be loaded if drush is run from the `$PROJECT/drush` directory.
* the site specific `local.drush.yml` file will be loaded if the drush command is being run from the `$PROJECT/drush` directory (for example), or even the `$PROJECT/docroot` directory, but not from the project root.

This is incredibly confusing to try and figure out. I think we should just remove these options unless there is a better way to indicate exactly where these files live...?

### DIFF
--- a/template/drush/drush.yml
+++ b/template/drush/drush.yml
@@ -3,11 +3,6 @@
 # ../docroot/sites/[site]/drush.yml
 drush:
   paths:
-    config:
-      # Load a drush.yml configuration file from the current working directory.
-      - ../docroot/sites/default/local.drush.yml
-      # Allow local global config overrides.
-      - local.drush.yml
     include:
       - '${env.home}/.drush'
       - /usr/share/drush/commands


dpagini:
Hmm, so tricky error here by the tests... before I try to go further, wondering if anyone has any thoughts on this? The test is erroring b/c it was calling drush from the docroot, but you should be able to see that if I call drush from `docroot/sites` directory, it will ignore this file again, leaving to _inconsistent_ and _confusing_ results.

danepowell:
There's some related upstream discussion here: https://github.com/drush-ops/drush/issues/3968

Drush maintainers make a couple salient points:
* If you want to set a site URI, you should use sites.php instead of local.drush.yml
* There are no use cases for local.drush.yml other than setting a URI

If that's true, then we should get rid of `local.drush.yml` altogether, and instead use `sites.php` or something else to fix tests? What do folks think? Seems like a big change and I'm not sure I fully appreciate the impact...

cc @TravisCarden

TravisCarden:
@danepowell I concur in theory that we should altogether eliminate `local.drush.yml` and use `sites.php` instead. I can't comment on the effect that may have on any other aspect of the architecture, though.

dpagini:
This is probably a drush issue and not BLT, but I do think there's value in the ability to have local only drush settings. For example, I develop on a Ubuntu VM, but like to use my browser from the host. When I type a `drush uli` right now, it opens Chrome on Ubuntu. I realize I could use user settings for this, but I would like to be able to set a project specific, local-only, config option. There are probably many other better use-cases, but this is the one I was investigating when I realized this issue was present.

danepowell:
Sounds like `sites.php` may not be able to be usable on ACSF projects: #2996

danepowell:
This PR is broken and jumps straight to a potential fix without identifying exactly what issue needs to be solved, so I'm closing it for now. To follow up, I created this issue that captures the problem as I understand it: #3467

PR #3426, opened 2019-03-07
Disable ANSI when BLT runs in Cloud Hooks.

danepowell:
Cloud Hook task logs don't support ANSI output, causing behavior described in #3146. This builds on that PR by disabling ANSI output in Cloud Hooks.

To recreate the issue, simply view any log output in a Cloud hook task. Without this PR, it'll look like `�[37;46;1m[success]�[39;49;22m Cache rebuild complete.`. With the PR, should be much cleaner: `[success] Cache rebuild complete.`

### DIFF
--- a/scripts/cloud-hooks/hooks/common/post-db-copy/db-scrub.sh
+++ b/scripts/cloud-hooks/hooks/common/post-db-copy/db-scrub.sh
@@ -18,6 +18,6 @@ repo_root="/var/www/html/$site.$target_env"
 export PATH=$repo_root/vendor/bin:$PATH
 cd $repo_root
 
-blt artifact:ac-hooks:db-scrub $site $target_env $db_name $source_env --ansi=false
+blt artifact:ac-hooks:db-scrub $site $target_env $db_name $source_env --drush.ansi=false
 
 set +v
--- a/scripts/cloud-hooks/hooks/common/post-db-copy/db-scrub.sh
+++ b/scripts/cloud-hooks/hooks/common/post-db-copy/db-scrub.sh
@@ -18,6 +18,6 @@ repo_root="/var/www/html/$site.$target_env"
 export PATH=$repo_root/vendor/bin:$PATH
 cd $repo_root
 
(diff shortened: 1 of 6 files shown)


danepowell:
If there were some way to set this option automatically for all commands in the `artifact:ac-hooks` namespace, I guess that might be a bit cleaner. I'm not sure that it matters enough to warrant the extra effort. Open to opinions.

alexxed:
How come this is not automated, why do we have to force it? Whatever automated process runs a script should be detected as non-interactive and without color support.

Robo supports this:
screenshot from 2019-03-07 05-54-31

scripts/cloud-hooks/hooks/common/post-db-copy/db-scrub.sh
```diff
@@ -18,6 +18,6 @@ repo_root="/var/www/html/$site.$target_env"
 export PATH=$repo_root/vendor/bin:$PATH
 cd $repo_root
 
-blt artifact:ac-hooks:db-scrub $site $target_env $db_name $source_env
+blt artifact:ac-hooks:db-scrub $site $target_env $db_name $source_env --ansi=false
```
mikemadison13: is there a reason this one is not drush.asni?

danepowell: Yeah, the reason is this was my last PR of the day and I must have been getting tired :) I'll fix it.

danepowell:
@alexxed great question. I did some digging and here's what I think is going on.

These hooks all just call Drush, which is built on Symfony Console, which _does_ have [auto-color support detection](https://github.com/symfony/symfony/blob/790854989e4b11462dc37cfd272d4cf3a542f73d/src/Symfony/Component/Console/Output/StreamOutput.php#L98).

I think the problem is in the Cloud UI or task runner. It probably claims to support colorized stream output, but when the logs are escaped for web rendering the ANSI characters aren't properly converted.

I'll refer this to the Cloud team for further investigation, but I think fixing it in BLT is easiest for now.

PR #3428, opened 2019-03-07
Add isset check on $_SERVER['PWD'] in blt-robo.php to fix PHP notice

brooke-heaton:
Some local environments, such as Lando, will not have a PWD key in $_SERVER, thus triggering a PHP Notice, which prevents Behat scripts from executing. Add an `isset` check on  `$_SERVER['PWD']` to allow other environments to function.

Fixes # 
--------

Changes proposed:
---------
 Add an `isset` check on `$_SERVER['PWD']` within `blt-robo.php` to prevent a PHP notice from breaking Behat scripts. 

Steps to replicate the issue:
----------

1. Using a local Lando environment with correctly configured BLT Behat 
2. Run `lando blt tests:behat:run` command

Steps to verify the solution:
-----------

1. There is no PHP Notice reading `Undefined index: PWD in /app/vendor/acquia/blt/bin/blt-robo.php on line 23` in your Lando environment
2. Behat tests run successfully in Lando
3. Yay

### DIFF
--- a/bin/blt-robo.php
+++ b/bin/blt-robo.php
@@ -19,8 +19,10 @@ require_once __DIR__ . '/blt-robo-run.php';
  * @return bool|string
  */
 function find_repo_root() {
+  // Check for PWD - some local environments will not have this key.
+  $pwd = isset($_SERVER['PWD']) ? $_SERVER['PWD'] : NULL;
   $possible_repo_roots = [
-    $_SERVER['PWD'],
+    $pwd,
     getcwd(),
     realpath(__DIR__ . '/../'),
     realpath(__DIR__ . '/../../../'),
--- a/bin/blt-robo.php
+++ b/bin/blt-robo.php
@@ -19,14 +19,15 @@ require_once __DIR__ . '/blt-robo-run.php';
  * @return bool|string
  */
 function find_repo_root() {
-  // Check for PWD - some local environments will not have this key.
-  $pwd = isset($_SERVER['PWD']) ? $_SERVER['PWD'] : NULL;
   $possible_repo_roots = [
-    $pwd,
     getcwd(),
     realpath(__DIR__ . '/../'),
     realpath(__DIR__ . '/../../../'),
   ];
+    // Check for PWD - some local environments will not have this key.
+  if(isset($_SERVER['PWD'])) {
+   array_unset($possible_repo_roots, $_SERVER['PWD']);
+  }
   foreach ($possible_repo_roots as $possible_repo_root) {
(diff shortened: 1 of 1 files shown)


mikemadison13:
@brooke-heaton  i'm a little concerned at just setting this to NULL. I understand that it fixes the warnings (GOOD) but... I assume that even on a Lando system we need to current path. So I'm curious what a NULL actually does to the code here?

brooke-heaton:
@mikemadison13 I may be confused. I don't see how this works:

```
  // Check for PWD - some local environments will not have this key.
  if(isset($_SERVER['PWD'])) {
   $pwd = $_SERVER['PWD'];
  }
  $possible_repo_roots = [
    $pwd,
    getcwd(),
    realpath(__DIR__ . '/../'),
    realpath(__DIR__ . '/../../../'),
  ];
```

Setting it to `NULL` allows `find_directory_containing_files` to run and if it is set, it will trigger a file check. How does this work without the `NULL` setting?

brooke-heaton:
@mikemadison13 

Another approach would not set it but only add it to `$possible_repo_roots IF it isset:

```
/**
 * Finds the root directory for the repository.
 *
 * @return bool|string
 */
function find_repo_root() {
  $possible_repo_roots = [
    getcwd(),
    realpath(__DIR__ . '/../'),
    realpath(__DIR__ . '/../../../'),
  ];
    // Check for PWD - some local environments will not have this key.
  if(isset($_SERVER['PWD'])) {
   array_unset($possible_repo_roots, $_SERVER['PWD']);
  }
  foreach ($possible_repo_roots as $possible_repo_root) {
    if ($repo_root = find_directory_containing_files($possible_repo_root, ['vendor/bin/blt', 'vendor/autoload.php'])) {
      return $repo_root;
    }
  }
}
```

mikemadison13:
@brooke-heaton i'm honestly not sure what the best solution is here, i may need to defer to you (or others). i just always get concerned setting a variable like this to null. i don't want to "fix" a warning but leave a hole for some other functionality, if that makes sense. we may be ok!

brooke-heaton:
@mikemadison13 I tweaked this to not set it but if it does exist to add the value to `$possible_repo_roots`. I've pinged the Lando team to get feedback on what the Lando equivalent of `$_SERVER['PWD']` would be. Sadly, I'm not really as well versed on System issues. I don't want to add anything specific to Lando. FWIW the patch does work on my local Lando environment, probably because one of the other three options is valid: 

```
getcwd(),
    realpath(__DIR__ . '/../'),
    realpath(__DIR__ . '/../../../'),
```

PR #3435, opened 2019-03-08
Update README.md

danepowell:
The section on releases and versioning is pretty verbose, and could be consolidated / simplified.

Additionally, I propose May 2019 as EOL for BLT 9.x (to correspond with EOL for Drupal 8.5.)


danepowell:
Mike approved out of band

PR #3439, opened 2019-03-11
Fixes #3050: Memcache stampede protection doesn't exist.

### LINKED ISSUE #3050: Memcache locking and stampede protection
I think we should discuss the memcache locking and stampede protection that BLT provides by default.

For at least some customers with certain traffic patterns (lots of concurrent auth users), we have evidence that memcache locking and stampede protection can have a significant negative performance impact.

At the very least, we should make it easier for these customers to disable locking and stampede protection, and document how to do this. Stampede protection is easy to disable by just setting ` $settings['memcache']['stampede_protection'] = FALSE;`, but locking I'm not 100% sure how to remove, since you not only have to remove the configuration in `acquia/blt/settings/memcache.yml`, but probably have to modify `$settings['bootstrap_container_definition']` as well.

Going one step further: should we disable locking and stampede protection for _all_ customers by default? Do we have any data on whether most users will benefit from those settings or not?

danepowell:
Fixes #3050
--------
Thanks to @bkosborne for pointing out that this cruft.

Changes proposed:
---------
- Remove the settings for memcache stampede protection, which are simply snakeoil and do nothing since [memcache removed stampede protection in 2.0-alpha7](https://cgit.drupalcode.org/memcache/commit/?h=8.x-2.x&id=7bba7982c129a0c0632b1483ab4dc0ff312e1859). This also brings us in line with the official [Acquia-recommended settings file](https://docs.acquia.com/acquia-cloud/performance/memcached/enable/#configuration-for-drupal-8).

Steps to replicate the issue:
----------
1. Observe that stampede protection wasn't actually working prior to this change.

Steps to verify the solution:
-----------
1. Observe that it still doesn't work after :smile:

### DIFF
--- a/settings/memcache.settings.php
+++ b/settings/memcache.settings.php
@@ -10,7 +10,8 @@ use Composer\Autoload\ClassLoader;
 // Check for PHP Memcached libraries.
 $memcache_exists = class_exists('Memcache', FALSE);
 $memcached_exists = class_exists('Memcached', FALSE);
-$memcache_module_is_present = file_exists(DRUPAL_ROOT . '/modules/contrib/memcache/memcache.services.yml');
+$memcache_services_yml = DRUPAL_ROOT . '/modules/contrib/memcache/memcache.services.yml';
+$memcache_module_is_present = file_exists($memcache_services_yml);
 if ($memcache_module_is_present && ($memcache_exists || $memcached_exists)) {
   // Use Memcached extension if available.
   if ($memcached_exists) {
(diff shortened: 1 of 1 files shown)


bkosborne:
@danepowell I noticed that Acquia's memcache docs still reference stampede protection and provide a code snippet to enable for D8, which doesn't make sense since it's no longer a feature: https://docs.acquia.com/acquia-cloud/performance/memcached/locks/. Maybe you could pass along to docs team internally. If more appropriate, I can submit a ticket, but it doesn't seem like support portal is correct usage for this type of request either. Would be good to have a feedback button on all doc pages.

danepowell:
@bkosborne thanks for the tip, I'll pass it along, but please do submit a support ticket of your own as well, direct feedback is the best way to ensure that this gets noticed. Make sure to request a better feedback mechanism for docs as well, I agree that would be useful.

PR #3440, opened 2019-03-11
Issue cloud hooks

danepowell:
Fixes a regression introduced by https://github.com/acquia/blt/pull/3426

Once confirmed, forward-port this to 10.x and make sure to update other cloud hooks as well.


mariagwyn:
Confirmed that this fix allows cloud deploys.

PR #3447, opened 2019-03-12
Fixed #1793: Allow custom commands via autoloader.

### LINKED ISSUE #1793: Allow custom commands to use autoloader

danepowell:
Fixes #1793 
--------

Changes proposed:
---------
This uses Robo's new class autoloading to allow third-party Composer packages to provide custom BLT commands.

Users that have already defined plugins will need to modify them slightly to use the new namespace / structure.

Additionally, hooks must now be defined within command files rather than as separate hook files.

Steps to verify the solution:
-----------
1. Verify that the built-in example command (`blt custom:hello`) works for new projects out of the box (i.e. check for regressions).
2. Verify that Composer packages can provide their own plugins. Do this by creating a new project and then adding this example repo as a composer dependency: https://github.com/danepowell/blt-example-plugin. Then you should be able to run `blt example:command-one` or `blt example:command-two`.

A few thoughts / to-dos:
- We should figure out how to handle users with existing plugins when they upgrade to 10.x. My feeling is that the number of people implementing these custom commands is small enough, and complexity involved with trying to update them large enough, that we can just advise folks of the change and let them manually update their custom commands. Open to other opinions though.
- Do we want to keep providing the "custom:hello" example command within BLT, or should we just point people to an external example Composer package? Should we clean up the namespace somehow?


greg-1-anderson:
Hooks should work from any commandfile that can define commands. If this isn't happening, it sounds like a Robo framework bug. I'll try to reproduce.

danepowell:
@greg-1-anderson thanks, that solved it! I just had to combine the two files. I'll update my PR.

greg-1-anderson:
The annotated commands project allows commands and hooks to be separated if you want, so we could support that. It would probably take some small enhancement to the Robo plugin mechanism. A PR would be welcome.

danepowell:
I'm not too bothered by the prospect of commands and hooks existing in the same files. If anyone else feels differently let me know.

danepowell:
I'm stumped as to how to fix the tests. The immediate problem causing them to fail is that they can't find the example custom commands.

I fixed this in normal usage by passing the autoloader/classloader to `src/Robo/Blt.php`, which in turn registers it with the Robo runner.

But this fails in tests because BltProjectTestBase creates its own copy of `src/Robo/Blt.php` and doesn't pass it a classLoader. And I have no idea where to get that classLoader from if it's already been loaded at some point earlier in the process (I can't just require it again).

@grasmash any ideas?

danepowell:
Nevermind, I figured it out and tests are passing now. General reviews appreciated.

greg-1-anderson:
In the current version of the annotated-command library, the behavior of adding two commands with the same command name is undefined. Under the current implementation, the last commandhandler registered will "win"; its commands simply overwrite whatever was there before.

A PR in [consolidation/annotated-command](https://github.com/consolidation/annotated-command) to alter the default behavior would be welcome. Note that there is already an annotation to explicitly override a command with a new implementation.

danepowell:
It doesn't seem ideal that it's undefined, but I'm not sure what a good explicit behavior would be. Maybe throw an exception if a command is defined twice?

Anyway, we should probably handle that in an upstream issue. I'm going to merge this as-is so we can cut a new beta and get some more eyes on it.

@greg-1-anderson thanks for your help and input on this! And for pointing out the [replace command hook](https://github.com/consolidation/annotated-command#replace-command-hook), which I wasn't aware of.

greg-1-anderson:
Throwing an exception wouldn't be great; under that model, if you added two plugins that each defined the same command, you'd kill the whole app. First-wins might be better than last-wins, though.

PR #3450, opened 2019-03-13
Exclude settings.php from codesniffing.

danepowell:
When upgrading to 9.2.x-dev, couldn't commit changes to settings.php because it actually isn't standards compliant.

Need to forward-port this

Also might need an update hook?

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -34,5 +34,8 @@
   <exclude-pattern>*/node_modules</exclude-pattern>
   <exclude-pattern>*/vendor</exclude-pattern>
   <exclude-pattern>*/default.local.settings.php</exclude-pattern>
+  <!-- Drupal itself doesn't meet code standards. -->
+  <!-- @see https://www.drupal.org/project/drupal/issues/2571965 -->
+  <exclude-pattern>*/settings.php</exclude-pattern>
 
 </ruleset>


phpcs.xml.dist
```diff
@@ -34,5 +34,8 @@
   <exclude-pattern>*/node_modules</exclude-pattern>
   <exclude-pattern>*/vendor</exclude-pattern>
   <exclude-pattern>*/default.local.settings.php</exclude-pattern>
+  <!-- Drupal itself doesn't meet code standards. -->
+  <!-- @see https://www.drupal.org/project/drupal/issues/2571965 -->
+  <exclude-pattern>*/settings.php</exclude-pattern>
```
greylabel: Shouldn't this already take care of it? https://cgit.drupalcode.org/drupal/tree/sites/default/default.settings.php#n3

danepowell: Good catch, thanks. The problem is that that line was only added a little over a year ago. Projects created before then won't have it. I'm not sure we have a mechanism to update it but maybe we should.

greylabel:
I feel that it is up to project owners to monitor and incorporate change from Drupal's `default.settings.php` into their project's `settings.php`, similar to how changes to robots.txt, etc often need special considerations.

danepowell:
@greylabel I think we should just advise people to update these files periodically. BLT can help, but it would be hard to automated because of the complexities around multisite: #3451

PR #3455, opened 2019-03-14
Backport - Replace  by (stable) global variable equivalent (#3408)

### LINKED ISSUE #3362: ACSF Multi-Site Deploy Failure

lcatlett:
Fixes #3362 
--------

Changes proposed:
---------
- Backports #3408 to BLT 9.2.x
- This has been tested and validated in production by multiple customers


danepowell:
This was already backported in #3414, hence the null diff.

PR #3458, opened 2019-03-14
Update Robo and add composer.lock

danepowell:
Backport of #3384

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# Ignore composer.lock to allow highest/lowest testing.
-composer.lock
-
 # Ignore blted8 files
 blted8/composer.lock
 blted8/vendor
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "composer/installers": "^1.2.0",
         "composer/semver": "^1.4",
         "consolidation/config": "^1.0.0",
-        "consolidation/robo": "~1.2.4",
+        "consolidation/robo": "~1.4.6",
         "cweagans/composer-patches": "^1.6.5",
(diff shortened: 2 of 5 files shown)


danepowell:
@alexxed I merged your PR to update Robo in 10.x, but for some reason when we try to update Robo in 9.2.x it fails tests. This is the main error I see: `Error: Class 'Symfony\Component\Config\Definition\Processor'`

Any ideas what's going on there?

If we can't get this updated in 9.2.x, we'll need to update the documentation to indicate that 9.2.x won't support Drush 9.6.

alexxed:
I can't reproduce the error locally. I might be missing something compared to Travis.

Drush 9.6 was released 20 hours ago, this failing build was 15 hours ago and logs say it pulled Drush 9.6 (which does not depend directly on symfony/config anymore). So I suspect it's related to that, and not this PR.

If we use symfony/config in BLT we should require it as a direct dependency.

I restarted the build for the 10.0.x PR: https://travis-ci.org/acquia/blt/builds/498280959 ; if it fails then we have a confirmation that Drush 9.6 is causing this.

alexxed:
Oh, can't run a build on a closed PR. Can you require symfony/config ^3.4 (this was what Drush 9.5 pulled) and see if the build passes?

danepowell:
Adding a composer.lock file fixed this, as expected. I'm going to merge this as-is (without updating Drush) and then we'll start new PRs to update Drush, and I'll respond to your feedback there.

alexxed:
This kind of confirms that drush 9.6 would break the tests as it did here as it doesn't pull symfony/config.

PR #2369, opened 2017-12-12
Fixes #2061: Validation doesn't occur in multisite paths.

grasmash:
Fixes various bugs discovered in the process of researching #2061.

PR #3462, opened 2019-03-14
Add composer.lock.

danepowell:
Not having a lock file committed to BLT causes problems. Tests take longer to run because the initial composer install has to do the all of the nasty dependency calculations. [Tests inexplicably fail from one run to the next](https://github.com/acquia/blt/compare/ec4d9d26a1ca...8e5ae2db0a6c) when dependencies update. Local development on BLT is harder because every time you switch branches, it preserves the composer.lock file from the wrong branch and truly bizarre things happen.

BLT's sister packages such as Robo, Drush, and Drupal all include a composer.lock, so it seems that we're the odd one out.

I'm not aware of any concrete reason not to have a composer.lock file committed, other than the fact that we'll have to update it occasionally.

It was originally removed in #2369 but no justification is provided as to why. @grasmash let me know if there's a reason you took it out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-# Ignore composer.lock to allow highest/lowest testing.
-composer.lock
-
 # Ignore blted8 files
 blted8/composer.lock
 blted8/vendor
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,9905 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,6 +1,3 @@
-# Ignore composer.lock to allow highest/lowest testing.
```
alexxed: I think this comment may make sense, but we could just remove the file during tests and run tests after a composer update as well.

danepowell:
Looks like this shaves at least two minutes off of every build (if you look at the initial `composer install` time), and solves the Drush 9.6 error we were running into in #3458. Merging.

.gitignore
danepowell: Agreed, we'll do that later if it's necessary.

PR #3479, opened 2019-03-19
Fixes #3238: Removed gitignore files from git on deploy.

### LINKED ISSUE #3238: Preventing dirty git index/cache on build artifacts
**Summary**: changes to `.gitignore` that exclude previously committed files leads to cached 'ghost' files on re-used artifact branches. This is bad because it can lead to non-removal of files, or, files that no longer get updated.

**Example**:
Bear with me on this one...
- `themes/custom/mytheme/.gitignore` says to ignore the theme's `/css` folder
- `deploy-exclude-additions.txt` additions says ignore `/docroot/themes/custom/**.gitignore`
  - Result: `themes/custom/mytheme/.gitignore` is not copied to the build artifact
- CI runs `blt deploy` on `master` and successfully creates `master-build`

Then in a new commit:
- The `/docroot/themes/custom/**.gitignore` entry is mistakenly removed from `deploy-exclude-additions.txt`
- Now `themes/custom/mytheme/.gitignore` is copied to the build artifact, and the css folder is ignored
- CI runs `blt deploy` on the `master` branch and successfully updates `master-build`. While `themes/custom/mytheme/css/` is now git ignored, the old css folder still lives in `master-build`
- A release is tagged and CI builds it. The release tag artifact has no theme css folder

This may be a rare case, but it's easy to prevent. In theory this could impact any folder/file type and occurs due to changes involving `.gitignore` file(s), or changes to any `deploy-exclude` rules that involve `.gitignore` file(s). Such as: adding a `.gitignore` to an existing theme or module, or performing a BLT upgrade from a much older version.

**Potential solutions**:

In `DeployCommand::commit()`, 
**A)** Remove all files from git cache during the deploy commit
- run `git rm -r --cached .` **before** `git add -A`
- Forces consistency across all artifact branches (tags by nature don't have this issue)

**B)** Add an option to check for dirty git cache, similar to `deploy()`s `ignore-dirty` option and `checkDirty()` function
- Option would trigger a check that executes `git ls-files --cached --ignored --exclude-standard` to identify cached git files that fall under any custom git ignore rules
- Option could cause deploy build failure. Additionally it could fail PRs if a CI test script called `blt artifact:deploy --dry-run` (for example when `if [ ${TRAVIS_PULL_REQUEST} != 'false' ];` to target PRs only)

Thoughts on adding either of these into BLT? I could open a PR for either. I feel **A** is preferred because it's more direct, and I can't think of any use cases for a build artifact to have gitignored items in its git index.

Neither would be easily accomplishable with any type of command hook due to when/where `DeployCommand::commit()` is called. And since this is related to the build artifact, I can't think of a reason to touch `Acquia\Blt\Robo\Commands\Git\GitCommand`.

danepowell:
Fixes #3238
--------

Changes proposed
---------
- Remove files from Git that would otherwise be ignored. Typically these are cruft that was committed prior to a change in the .gitignore file, which then never got deleted.

Steps to replicate the issue
----------
Follow steps in #3238 (requires a deployable repo, aka ACE)

Expected behavior
-----------
Ignored files get deleted on the next run of `blt deploy`.

Additional details
-----------
I tried this for myself on some personal sites and they turned up a bunch of cruft that shouldn't be there.

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -539,6 +539,7 @@ class DeployCommand extends BltTasks {
     $this->say("Committing artifact to <comment>{$this->branchName}</comment>...");
     $result = $this->taskExecStack()
       ->dir($this->deployDir)
+      ->exec("git rm -r --cached .")
       ->exec("git add -A")
       ->exec("git commit --quiet -m '{$this->commitMessage}'")
       ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)


christopher-hopper:
Brilliant improvement. Love this change!

PR #3489, opened 2019-03-21
Note release cadence in README.

danepowell:
@anavarre @mikemadison13 I've revamped a little more than just the release cadence here, please review.

Since we get a lot of questions about these things, I want to make it explicitly clear how long each release of BLT will be supported, what's the recommended/stable release, and how long folks will have to upgrade.

I chose this cycle to keep BLT and Drupal core upgrades decoupled. This ensures that when projects have to upgrade to the next version of BLT / Drupal core, they always have the option of upgrading _either_ BLT first or Drupal core first. They don't have to upgrade both at once, which can be more challenging.

For instance, imagine you are using BLT 9.2.x and Drupal 8.6.x, as most users are right now. When BLT 10.0 and Drupal 8.7.0 drop in April/May, you have the option of either upgrading to BLT 10.0 first and then to Drupal 8.7.0, or vice versa. And you have 6 months total to make both upgrades.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,9 +22,9 @@ See [INSTALL.md](INSTALL.md) for a list of prequisites and links to instructions
 
 ## Releases and versioning
 
-BLT generally has an LTS release, supported stable release, and unstable (HEAD) release at any given time, each corresponding to a major [semantic version](https://semver.org/). The newest supported major version will receive bug fixes and new features, while the penultimate ("LTS") major version will receive security fixes for six months. This is intended to coincide with the [Drupal core release cycle](https://www.drupal.org/core/release-cycle-overview), so that users can continue to use a single BLT major release through the lifecycle of a Drupal core minor release.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -22,16 +22,20 @@ See [INSTALL.md](INSTALL.md) for a list of prequisites and links to instructions
...
-| 9.x           | LTS, EOL May 2019           | 8.5    | >=9.1.0        | \*bug fixes  |
-| <=8.9.x       | Unsupported, EOL            | <=8.5  | ~8             |              |
+| BLT version | Support status        | End of life  |  Drupal versions* | Drush versions |
+|-------------|-----------------------|--------------|-------------------|----------------|
+| 10.x        | Unsupported (beta)    | May 2020     | 8.6, 8.7          | >=9.5.0        |
+| **9.2.x**   | **Supported, stable** | **Dec 2019** | **8.6, 8.7**      | **>=9.4.0**    |
+| 9.x         | Security fixes only   | May 2019     | 8.5               | >=9.1.0        |
+| <=8.9.x     | Unsupported           | Dec 2018     | <=8.5             | ~8             |
+
+*Note that when Drupal ends support for a minor core release, BLT will cease supporting that core release as well. For instance, as of December 2019, BLT 10.x will no longer support Drupal 8.6, and will instead support Drupal 8.7 and 8.8.
```
anavarre: We could also clarify that what we're doing here is to stick with what the Drupal Security Team supports: https://www.drupal.org/drupal-security-team/general-information

PR #3491, opened 2019-03-22
Clarify the BLT release cadence

anavarre:
Per the PR, we'd now commit to the following:

* First Wednesday of every month: Bugfix release window
* Third Wednesday of every month: Security release window


danepowell:
Thanks, although I've already opened a PR for this so let's consolidate discussion there: #3489

PR #3492, opened 2019-03-22
Fixes #3255: On Acquia Cloud, acquia_config.php should better handle multisite

### LINKED ISSUE #3255: On Acquia Cloud, acquia_config.php should better handle multisite
Currently, BLT ships with `acquia_config.php` and recommends:

```
elseif (getenv('AH_SITE_ENVIRONMENT')) {
(snip)
 $config['baseurlpath'] = 'simplesaml/';
```

However, in a multisite environment with `simplesamlphp/simplesamlphp:1.16.2` and `simplesamlphp/saml2:3.2.2`, it seems we don't have a choice but to change the code to do this instead:

```
elseif (getenv('AH_SITE_ENVIRONMENT')) {
(snip)
 $config['baseurlpath'] = $protocol . $_SERVER['HTTP_HOST'] . $port . '/simplesaml/';
```

Otherwise, port **443** will be missing from the URL and give a permission denied. The [SAML Chrome Panel](https://chrome.google.com/webstore/detail/saml-chrome-panel/paijfdbeoenhembfhkhllainmocckace) extension can help with confirming this issue.

Considering we are in a multisite, was this expected? Why wouldn't the BLT config file ship with this OOTB setup, then?

cc/ @thomashowell

anavarre:
Fixes #3255 
--------

Changes proposed
---------
- Place the multisite-friendly code right where it belongs, under `elseif (getenv('AH_SITE_ENVIRONMENT'))`

Steps to replicate the issue
----------
1. It's not easy to replicate the issue. In multisite envs, there are instances where port 443 will be missing from the constructed URL and give a permission denied without any easy way to get to the bottom of the issue. When this happens you need to know about and leverage the [SAML Chrome Panel extension](https://chrome.google.com/webstore/detail/saml-chrome-panel/paijfdbeoenhembfhkhllainmocckace) to find out port 443 isn't being passed, thus causing the issue.

Previous behavior (before applying PR)
----------
Port 443 could be missing from the constructed URL.

Expected behavior (after applying PR)
-----------
Port 443 is always being passed in the constructed URL.

Additional details
-----------
Please review carefully before merging. There might be a need to simply duplicate the `baseurlpath` config override in and out of the `AH_SITE_ENVIRONMENT` if statement to account for non-Acquia hosted users.

### DIFF
--- a/scripts/simplesamlphp/acquia_config.php
+++ b/scripts/simplesamlphp/acquia_config.php
@@ -44,14 +44,6 @@ $protocol = 'https://';
 $port = ':' . $_SERVER['SERVER_PORT'];*/
 
 /**
- * Support multi-site and single site installations at different base URLs.
- *
- * Overide $config['baseurlpath'] = "https://{yourdomain}/simplesaml/"
- * to customize the default Acquia configuration.
- */
-$config['baseurlpath'] = $protocol . $_SERVER['HTTP_HOST'] . $port . '/simplesaml/';
-
-/**
  * Cookies No Cache.
  *
  * Allow users to be automatically logged in if they signed in via the same
@@ -84,7 +76,14 @@ if (!getenv('AH_SITE_ENVIRONMENT')) {
 
 }
 elseif (getenv('AH_SITE_ENVIRONMENT')) {
-  // Set  ACE ad ACSF sites based on hosting database and site name.
+  /**
+   * Support multi-site and single site installations at different base URLs.
+   *
+   * Overide $config['baseurlpath'] = "https://{yourdomain}/simplesaml/"
+   * to customize the default Acquia configuration.
+   */
+  $config['baseurlpath'] = $protocol . $_SERVER['HTTP_HOST'] . $port . '/simplesaml/';
+  // Set ACE and ACSF sites based on hosting database and site name.
   $config['certdir'] = "/mnt/www/html/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/simplesamlphp/cert/";
   $config['metadatadir'] = "/mnt/www/html/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/simplesamlphp/metadata";
   $config['baseurlpath'] = 'simplesaml/';


scripts/simplesamlphp/acquia_config.php
```diff
@@ -84,7 +76,14 @@
...
+   * Support multi-site and single site installations at different base URLs.
+   *
+   * Overide $config['baseurlpath'] = "https://{yourdomain}/simplesaml/"
+   * to customize the default Acquia configuration.
+   */
+  $config['baseurlpath'] = $protocol . $_SERVER['HTTP_HOST'] . $port . '/simplesaml/';
+  // Set ACE and ACSF sites based on hosting database and site name.
   $config['certdir'] = "/mnt/www/html/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/simplesamlphp/cert/";
   $config['metadatadir'] = "/mnt/www/html/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/simplesamlphp/metadata";
   $config['baseurlpath'] = 'simplesaml/';
```
danepowell: This needs to be removed or it will clobber what you've defined above, right?

anavarre: I'm not sure I understand what you mean. I've just moved around the `$config['baseurlpath'] = $protocol . $_SERVER['HTTP_HOST'] . $port . '/simplesaml/';` line which fixed the missing port 443 in the URL for a multisite we have.

I'm not saying this is best practice or the right fix for BLT, but it's what worked for us. It was really not obvious to understand it has to be within the `elseif (getenv('AH_SITE_ENVIRONMENT'))` block thus why putting it here - assuming it's where it should belong by default - looked best to prevent others from running into the issue.

danepowell: I'm talking about this line, three lines below the code that you inserted: `$config['baseurlpath'] = 'simplesaml/';`

That will clobber the value of baseurlpath that you just assigned.

lcatlett:
Port 443 should only be explicitly required in this config when deliberately manipulating server variables for HTTPS and SERVER_PORT when SSL is terminated at the ELB/balancer - this should already be done in BLT https://github.com/acquia/blt/blob/10.x/settings/simplesamlphp.settings.php#L15-L17 so it seems that something else is off with this config. 


I think it is more likely that host forwarding is misconfigured in the BLT config which is causing issues on both multisite and single site implementations, it may just be more obvious on multisite since the uri of an individual multisite is usually different than the absolute url of the saml SP. This should likely be tested/validated in the BLT Travis build, as it currently only tests that the saml config files were copied to the correct place rather than if the saml settings being used in BLT actually work... @danepowell do you agree?

scripts/simplesamlphp/acquia_config.php
```diff
@@ -84,7 +76,14 @@
...
 }
 elseif (getenv('AH_SITE_ENVIRONMENT')) {
-  // Set  ACE ad ACSF sites based on hosting database and site name.
+  /**
+   * Support multi-site and single site installations at different base URLs.
+   *
+   * Overide $config['baseurlpath'] = "https://{yourdomain}/simplesaml/"
+   * to customize the default Acquia configuration.
+   */
+  $config['baseurlpath'] = $protocol . $_SERVER['HTTP_HOST'] . $port . '/simplesaml/';
```
timcosgrove: Noting here that `$protocol` and `$port` are not defined unless the code block above is uncommented. This was also true of the code prior to this PR.

danepowell:
As far as I can tell this will have no functional change because `$config['baseurlpath']` gets overwritten to be `/simplesaml` regardless. So if people find that this is somehow more helpful I guess I'm fine with that.

PR #3498, opened 2019-03-25
Update simplesaml acquia_config.php for use on ODEs.

### LINKED ISSUE #3437: acquia_config.php for SimpleSAMLphp does not work on Acquia Cloud ODE.
In the `acquia_config.php` included in BLT, the database credentials are extracted in lines 95-97:
```
$creds_json = file_get_contents('/var/www/site-php/' . $_ENV['AH_SITE_GROUP'] . '.' . $_ENV['AH_SITE_ENVIRONMENT'] . '/creds.json');
$databases = json_decode($creds_json, TRUE);
$creds = $databases['databases'][$_ENV['AH_SITE_GROUP']];
```

The hostname of the database is then looked up on line 106-107:
```
$response = $resolver->query("cluster-{$creds['db_cluster_id']}.mysql", 'CNAME');
$creds['host'] = $response->answer[0]->cname;
```

On the ODE I want to run SSO tests on, $creds['db_cluster_id'] is `null`, so the host is empty, which causes an exception. I was able to fix this by hardcoding the database host for the specific ODE environment it was on.

But since https://support.acquia.com/hc/en-us/articles/360008830834 implies that ODEs should work, this code may need to be updated to get the DB host for ODEs.

wu-edward:
Fixes #3437 
--------

Changes proposed
---------

-
-

Steps to replicate the issue
----------
1. 
2. 

Previous behavior (before applying PR)
----------
(Remove if this is not a bug fix)

Expected behavior (after applying PR)
-----------

Additional details
-----------

### DIFF
--- a/scripts/simplesamlphp/acquia_config.php
+++ b/scripts/simplesamlphp/acquia_config.php
@@ -95,19 +95,24 @@ elseif (getenv('AH_SITE_ENVIRONMENT')) {
   $creds_json = file_get_contents('/var/www/site-php/' . $_ENV['AH_SITE_GROUP'] . '.' . $_ENV['AH_SITE_ENVIRONMENT'] . '/creds.json');
   $databases = json_decode($creds_json, TRUE);
   $creds = $databases['databases'][$_ENV['AH_SITE_GROUP']];
-  require_once "/usr/share/php/Net/DNS2_wrapper.php";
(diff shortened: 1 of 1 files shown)


wu-edward:
@danepowell Tested on dev and ode for client site successfully.

PR #3503, opened 2019-03-28
Cleaned up multisite docs and best practices.

danepowell:
The docs were a little out of date, BLT now does a few more things than before such as setting up db users and passwords.

I think we should also make a strong recommendation to use production FQDNs (i.e. `example.com` instead of just `example`) as site names in order to ensure compatibility with other BLT commands, be consistent with the rest of our docs (such as the multisite array example below), and the core multisite system, which generally expects the sites directory to be structured like `sites/example.com` instead of `sites/example`.

### DIFF
--- a/docs/multisite.md
+++ b/docs/multisite.md
@@ -5,25 +5,19 @@ There are two parts to setting up a multisite instance on BLT: the local setup a
 ## Local setup
 
 1. Set up a single site on BLT, following the standard instructions, and ssh to the vm (`vagrant ssh`).
-1. Run `blt recipes:multisite:init`. It is suggested to use a fully-qualified production domain name as the name of your site (e.g. ‘example.com’ instead of just ‘example’) in order to avoid having to manually configure sites.php.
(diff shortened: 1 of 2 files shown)


lcatlett:
@danepowell - using a FQDN for $site_name breaks config splits and drush aliases in BLT's current implementation of both. I get that the intention was to leverage the auto-detection in drush rather than having blt update sites.php, but the site directory is more associated with $site_path rather than the prod FQDN to support site install and autodetection in any environment. This is why the default drush uri fallback is `default` as opposed to a standard domain, otherwise drupal couldnt be installed in many environments and test fixtures. Mapping of the multiple domains to the site path is then usually managed in sites.php, as there is usually at least two production domains anyway since Acquia and other hosting providers provide an internal default domain and that is rarely equivalent to the real live customer domain.

danepowell:
When I tried to set up site names like "example" instead of "example.com", commands like `blt setup --site=example` would result in calls to Drush like `drush si --uri=example`. That seems a little screwy, is that intentional for the uri?

PR #3510, opened 2019-04-02
Move Drupal module dependencies to project template

danepowell:
This moves Drupal modules out of BLT's root composer.json and into the blt-project composer.json. It also removes the build directory where these modules (and core) used to be installed. I think this is an improvement in a couple ways.

As a BLT developer, the existence of the build directory seems completely unnecessary except as a place for composer-installers to dump Drupal modules. This is a little annoying because it clutters up the repo and PHPStorm is always trying to index these files, and occasionally Composer does weird things to it.

As a BLT end user, this should provide some additional flexibility by allowing users who don't use e.g. Features to remove that module from their codebase entirely. There's no direct link between BLT and these modules that I'm aware of (in terms of inheriting classes or anything), they are only invoked via Drush if configured as such. So it seems safe to make them configurable per-project.


danepowell:
Actually we might want to move drupal/core as well, since that would _drastically_ improve Composer memory usage by working around this issue: https://github.com/zaporylie/composer-drupal-optimizations/issues/12

alexxed:
Well I think we can merge this and open another one for Drupal core to keep things moving.

danepowell:
I forgot we might need an update hook to stick these modules into 9.x users' composer.json. I'll take a look.

PR #3512, opened 2019-04-03
Use PHP 7.2 by default for Pipelines

danepowell:


danepowell:
I think we want to use the same PHP version across the board (DrupalVM, Pipelines, etc...) and @mikemadison13 correctly points out that [PHP 7.2 isn't supported on ACSF](https://docs.acquia.com/site-factory/known-issues/#acsf-php72), meaning that we'd have to downgrade folks when they init ACSF in BLT. That doesn't seem worth the effort.

lcatlett:
PHP 7.2 is supported for Drupal 8 on ACSF - the docs are being updated today

PR #3516, opened 2019-04-04
Make Behat play nicely with multisite setups

### LINKED ISSUE #2860: Why does BLT require a local behat config file?
We're trying to simplify our behat config files and just use the main `behat.yml` config file instead of also using the `local.yml` that BLT generates.  The reasoning: In my project, we use a docker container for local development and use the same exact container setup for our CI/testing infrastructure, so no matter where the tests are run, the config will always be the same. 

My initial understanding of this is that I could just put stuff in the `local.yml` file if I wanted to _override_ configuration from the main config file, but that's not the case. BLT actually requires that this file exist and it even checks for specific configuration from that file.

lpeabody:
Fixes #2860 
--------

Changes proposed
---------
Add a `behat.validate` configuration item to disable Behat validation and wizards and simply just run the `behat` command. Trust that, if we so choose, we know how to setup our own Behat configuration as needed.

-----------
When setting `behat.validate` to `false`, you are able to run Behat tests even if the base_url value in local.yml is not the same value as base_url included in whichever configuration file/profile you have chosen to run.

Additional details
-----------
This is just a starting point for having something tangible to help guide discussion around issue #2860. Currently built against the 9.2.x branch since this is an early prototype for this feature that some developers (me) might be leveraging in their current (non-10.x) projects.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -1,5 +1,6 @@
 # @todo Move to subkey of tests.
 behat:
+  validate: true
   config: ${repo.root}/tests/behat/local.yml
   profile: local
   # The URL of selenium server. Must correspond with setting in behat's yaml config.
--- a/src/Robo/Commands/Tests/BehatCommand.php
+++ b/src/Robo/Commands/Tests/BehatCommand.php
@@ -75,7 +75,7 @@ class BehatCommand extends TestsCommandBase {
   }
 
   /**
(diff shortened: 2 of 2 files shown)


danepowell:
This also would need to go into 10.x. I'm going to consolidate it into #3517

PR #3517, opened 2019-04-04
Allow easier non-interactive Behat tests on multiple sites.

### LINKED ISSUE #2860: Why does BLT require a local behat config file?

danepowell:
Fixes #2860 

@lpeabody @brooke-heaton does this help if you follow the multisite testing pattern I described [here](https://github.com/acquia/blt/issues/2860#issuecomment-480052280)?

Maybe you still want something like #3516 to completely disable validation, but it seems like it would be better to natively support the multisite use case. Following the pattern I laid out in #2860, this seems to work great. You define each site's URL (`project.local.hostname`, and any other site-specific behat config) in `blt.yml`, then run Behat tests like `blt tests:behat --site=baz --no-interaction` and BLT will automatically regenerate local.yml to contain the right base url (as well as other Behat params).


lpeabody:
It would be nice to handle natively but I really really really don't want to be forced to run every site through the same `local.yml` generated from `example.local.yml` since `example.local.yml` would be a constant for all sites. If I want to actually utilize profiles and suites, that file gets real big real fast and contains configuration that isn't necessarily pertinent to other sites/profiles/whatever. I want to be able to setup and use everything defined in my own Behat configuration files that are themselves setup to handle the testing of the site with which they are associated.

I would personally like to see hardcoded strings referencing `local.yml` and the `base_url` key path to be brought into configuration, and those values used in the pertinent BLT commands. Then you could start doing things like, on a per-site level:

```
behat:
  example-config: ${repo.root}/tests/behat/example.local.yml
  config: ${repo.root}/tests/behat/local.yml
  base-url-path: ${behat.profile}.extensions.Behat\MinkExtension.base_url
  profile: local
  suite: default
  always-recreate: false
```

1) The BLT commands know where the given site's example config lives and where it's computed config file should go. This is inline with how `local.settings.php` is generated as well.
2) The validation knows the path to `base_url`, so it can get it's value from the computed config file and verify that it matches the site's `project.local.hostname`.
3) Having per-site profiles that aren't named local (you can kind of already do this as the correct profile would get passed to the command, but can't really because of the validation on base_url).
4) Utilize multiple suites.
5) You can specify whether or not the local.yml file should always be recreated prior to running tests.
6) Everything is overridable at the time of the command by passing `-D`.

That feels much more BLT-y to me, personally. I think that also preserves backwards compatibility. This would also let us run tests in CI sandboxes where the hostname is automatically generated based on the name of the branch.

In any case, I'm definitely willing to help out with this if we can get agreement on what we want our most beloved BLT to be able to do for us :)

danepowell:
What you're describing doesn't really sound that different from what's supported today other than validation needing to work a little more gracefully, if I understand right.

Just to make sure I completely understand what you are trying to accomplish, could you create a dummy project with a few example sites and Behat tests and profiles all configured the way you _think_ it should work, so it's more clear what the feature gap is? And maybe create a new issue where we can focus discussion on this, along with steps to reproduce the problem.

I still think there's value in this PR, because it allows you to run `blt tests:behat --no-interaction` and have BLT automatically regenerate local.yml if necessary.

My only concern is that it dumps the interaction and validation hooks in favor of just validating / regenerating local.yml all inline with the behat method. This is a little ugly, but shouldn't break anything (nothing else uses those hooks), and I don't know how to preserve the old architecture because interaction hooks will always be skipped when BLT is run with `--no-interaction`. So there'd be no way to force BLT to regenerate local.yml.

PR #3518, opened 2019-04-05
Turn off ssh strict host checking when drush is connecting to drupal vm

alexxed:
Fixes # 
--------

Changes proposed
---------

When drush connects to the vm, turn strict host checking to off. This avoids prompting the user like this:

The authenticity of host 'local.aszaszbltproject.com (192.168.114.250)' can't be established.
ECDSA key fingerprint is SHA256:ujyiJ45p0pxPhacRWHlXGDA797msE7X4G6lWBDWQhGI.
Are you sure you want to continue connecting (yes/no)

Steps to replicate the issue
----------
1. temporarily remove your ssh known hosts file
2. run drush browse 
3. you should get prompted like above

While using this branch, you should not get prompted

### DIFF
--- a/scripts/drupal-vm/drupal-vm.site.yml
+++ b/scripts/drupal-vm/drupal-vm.site.yml
@@ -4,4 +4,4 @@ local:
   host: ${project.local.hostname}
   user: vagrant
   ssh:
-    options: -o PasswordAuthentication=no -i $HOME/.vagrant.d/insecure_private_key
+    options: -o StrictHostKeyChecking=no -o PasswordAuthentication=no -i $HOME/.vagrant.d/insecure_private_key


danepowell:
Looks good, thanks.

I only had one concern, but I confirmed it's a non-issue. This will still write host keys to `known_hosts` and throw a nasty warning if the key changes (although it won't throw a fatal like before). I confirmed that if you destroy and recreate a VM, the host key doesn't change, which is good, so it shouldn't be a problem. I'm honestly not sure how the host key gets set or what would cause it to change :man_shrugging: 

For posterity, this is how I used to work around this, but obviously your approach is less work. I would add the following to `/etc/ssh/ssh_config` and then delete `known_hosts`:
```
Host local.*
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  LogLevel=quiet
```

alexxed:
I opened a PR to get vagrant transport in Drush, if that gets in, it will simplify this: https://github.com/consolidation/site-process/pull/40/files . If merged, it will need another PR in drush.

PR #3519, opened 2019-04-05
Removed Phing cruft.

danepowell:
I'm pretty sure this is just cruft left over from when we used Phing. It worries me a little that it wasn't causing a failure or something, I'm wondering what else could be cleaned up here. I don't really understand it well enough to say.


tests/phpunit/Robo/Commands/CommandTestCase.php
```diff
@@ -174,8 +174,6 @@ public function setMockExecutor() {
    */
   protected function createDefaultConfig() {
     $this->config = new DefaultConfig();
-    $this->config->extend(new YamlConfig($this->config->get('blt.root') . '/phing/build.yml',
-      $this->config->toArray()));
     $this->config->extend(new YamlConfig($this->config->get('blt.root') . '/template/blt/blt.yml',
       $this->config->toArray()));
     $this->config->extend(new YamlConfig($this->config->get('blt.root') . '/template/blt/local.yml',
```
danepowell: This is also not a thing any more...

danepowell:
Actually I think this entire class has never been used since it was first added in https://github.com/acquia/blt/pull/1234

PR #3522, opened 2019-04-05
Fixed profile splits.

danepowell:
Drupal dropped `$settings['install_profile']`, which has [caused consternation for folks](https://www.drupal.org/project/drupal/issues/2975328#comment-12711910) who need this information to dynamically set things in settings.php, such as the active config split.

This should get around it. It requires that multisite users set their profile in the multisites array or a site-specific blt.yml.

If the profile configured in blt.yml doesn't match the actually-installed profile in your active config, tough cookies, this will still set the profile split based on blt.yml. I don't think there's any way for us to determine the actual profile for an installed site. This seems like a reasonable compromise to me, but let me know if I'm missing something.

### DIFF
--- a/settings/config.settings.php
+++ b/settings/config.settings.php
@@ -1,5 +1,7 @@
 <?php
 
+use Acquia\Blt\Robo\Config\ConfigInitializer;
+
 /**
  * @file
  * Controls configuration management settings.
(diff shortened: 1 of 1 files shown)


danepowell:
I don't think this should be considered BC-breaking because the upstream issue to remove `$settings['install_profile']` from Drupal core went into Drupal 8.6, which is the lowest supported core version for BLT 9.2. But it should probably be included as a release note.

PR #3530, opened 2019-04-08
Issue 3529: Adding --ignore-unmatch parameter so deploy stops failing.

### LINKED ISSUE #3529: Artifact deploy failing with message: fatal: pathspec '.' did not match any files
- [ X ] I have read and understand the [contribution guidelines](../../docs/CONTRIBUTING.md) for new issues.

**Describe the bug**
Running:
`vendor/bin/blt artifact:deploy --commit-msg "Test commit" --ignore-dirty --no-interaction --verbose -Dgit.remotes.1='REMOTE' --branch=BRANCH`

Throws:

`fatal: pathspec '.' did not match any files
[ExecStack]  Exit code 128  Time 0.014s
[error]  Failed to commit deployment artifact! `

**Expected behavior**
Artifact to be deployed.

**Additional context**
This seems to happen because the repo doesn't have any files yet, so the command fails to remove all files as none exists. Adding [`--ignore-unmatch`](https://www.atlassian.com/git/tutorials/undoing-changes/git-rm) fixes it. 

**System information**
* Operating system type: MacOS
* Operating system version: Mojave
* BLT version: 10.0.0-rc-2

jose-oliveira:
Fixes #3529

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -537,7 +537,7 @@ class DeployCommand extends BltTasks {
     $this->say("Committing artifact to <comment>{$this->branchName}</comment>...");
     $result = $this->taskExecStack()
       ->dir($this->deployDir)
-      ->exec("git rm -r --cached .")
+      ->exec("git rm -r --cached --ignore-unmatch .")
       ->exec("git add -A")
       ->exec(["git commit --quiet -m", escapeshellarg($this->commitMessage)])
       ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)


TravisCarden:
Your understanding of ORCA's purpose is correct, @JakeWilund. The reason a BLT change with a negative impact wouldn't be caught at present is that BLT is not yet using ORCA to test itself. As you probably know, BLT is currently being transitioned from PS to Product to get official, staffed development and support. I don't know what the status or priority is for getting it testing itself with ORCA. Perhaps we can check into that together.

JakeWilund:
Thanks @TravisCarden, I am actually encouraged to hear that BLT is going to get the product level commitment it deserves and that ORCA's purpose is in fact to catch this sort of thing in the future.

Looks like @jose-oliveira has a minor fix that would resolve our testing failure issues. Hoping we can get that approved and merged to unblock us!

danepowell:
Thanks for the PR and quick response to this everyone.

The reason this wasn't caught beforehand is because we do all of our manual testing with a "live" (pre-existing) representative customer application (including deploying to a Cloud env), and automated testing by creating new projects in a CI environment (up to, but _not_ including, deploying projects to a Cloud env).

To catch this particular error would have required testing the initialization of a brand new Cloud environment on every test run, which is obviously somewhat more complex. I'd love your thoughts on how to accomplish that (it sounds like ORCA must be doing that if this caused it to fail).

As to why BLT isn't integrated with ORCA yet, it's a bit more complex than you might think because BLT is part of the orchestration layer of ORCA itself. It's not a typical stand-alone product or service like the other components of ORCA. Obviously we'd still like to integrate with it if possible, it's just not trivial.

TravisCarden:
Testing BLT with ORCA may actually prove somewhat easier than you expect, @danepowell. Just ping me when you're ready to investigate it.

mpodorov:
Is it possible to add --quiet to remove thousands of meaningless lines in the command output?

danepowell:
Under what conditions are you seeing "thousands of meaningless lines"? It sounds like you are referring to Git removing previously cached / now ignored files. I don't think it's at all meaningless to warn (or at least log) that these files are being removed. Plus, it should only happen once, right?

timcosgrove:
I expect we will see it on the creation of every deploy artifact, no?

I'm not sure it warrants `--quiet` though the option might be nice.

danepowell:
Sorry, I think I see what you are saying now. I was confused because you brought it up in this thread instead of where we actually added `rm -rf --cached` in https://github.com/acquia/blt/pull/3479, so I thought you were specifically complaining about something related to the --ignore-unmatch parameter.

See https://github.com/acquia/blt/pull/3552

Issue #3526, opened 2019-04-08
False positive suspected with BLT 9.2.6 and Acquia Pipelines


anavarre:
Upon trying to update to BLT 9.2.6, Acquia Pipelines fails with the following:

```
blt deploy:check-dirty --ansi --verbose

[error]  There are uncommitted changes, commit or stash these changes before deploying.
```

Running the same command locally (against the _same_ Git branch) works as expected since the working tree is clean:

```
$ git status
On branch blt-9.2.6
Your branch is ahead of 'origin/master' by 1 commit.
(use "git push" to publish your local commits)
nothing to commit, working tree clean
$ blt deploy:check-dirty --ansi --verbose
0.088s total time elapsed.
```

-----
I understand there's a way to temporarily disable the command with BLT's `blt.yml`'s `disable-targets` but haven't yet found a way to do so reliably.

danepowell:
Running the command locally won't have the same effect since there are lots of things in the Pipelines scripts that only run on Pipelines. For instance, if you've applied this PR, it would cause this error: https://github.com/acquia/blt/pull/3525

You'll need to add a `git status` to your Pipelines scripts to figure out what exactly has uncommitted changes. Maybe that's something BLT could do for you in the long-run.

danepowell:
This should help you figure out what changed: https://github.com/acquia/blt/pull/3531

anavarre:
@danepowell thanks for #3531 - Applied the patch and that's really helpful to understand what's going on, though I can't explain _why_ those files are there.

danepowell:
If you still need more support, please post the output of the checkDirty command so we can see what files changed.

anavarre:
Sure, here's the output:

```
blt deploy:check-dirty --ansi --verbose

 M docroot/sites/default/settings.php
 M docroot/sites/glossary.acquia.com/settings.php
 M docroot/sites/ls.acquia.com/settings.php
?? docroot/sites/default/settings/default.includes.settings.php
?? docroot/sites/glossary.acquia.com/settings/default.includes.settings.php
?? docroot/sites/ls.acquia.com/settings/default.includes.settings.php
?? docroot/sites/settings/

[error]  There are uncommitted changes, commit or stash these changes before deploying.
```

danepowell:
I pinged @anavarre via Slack, it turns out this is most likely a dupe of https://github.com/acquia/blt/issues/3496

BLT's settings includes were not committed to Git and thus being re-written as part of the deploy process.

[...]

PR #3531, opened 2019-04-09
Print output from checkDirty command.

danepowell:
Should help folks to fix problems like #3526


anavarre:
It does help, thanks.

PR #3535, opened 2019-04-09
Clarify ACSF config setup docs.

danepowell:
Looking for feedback on this, I think this is considered best practice on ACSF.


danepowell:
@davidtrainer let me know if you have any feedback on this based on your ongoing work

PR #3537, opened 2019-04-10
Don't glob global settings by default.

### LINKED ISSUE #3528: Suggestion: Remove global.settings.default.php
This is not really a problem, just a suggestion.

The `global.settings.default.php` is more of documentation than a usable PHP file. Can this just be a `README.md` file in the `docroot/sites/settings/` folder?

Or why not make it an actual `*.settings.php` file that gets included? Someone has to modify it before it _does_ anything.

danepowell:
Fixes #3528
--------

Changes proposed
---------
- Rename global.settings.default.php to default.global.settings.php
- Instead of globbing all settings in `docroot/sites/settings/*.settings.php`, only include `docroot/sites/settings/global.settings.php` if it exists.
- Add a snippet to default.global.settings.php to glob any other files in `docroot/sites/settings/*.settings.php` in order to mimic the old behavior.

This is a breaking change for users that have files in docroot/sites/settings named anything other than global.settings.php, so it will require a change record. But I don't think it needs an update hook.

### DIFF
--- a/docs/next-steps.md
+++ b/docs/next-steps.md
@@ -58,4 +58,41 @@ Other commonly used commands:
 
 ## Drush aliases
 
-See [BLT Drush Documentation](drush.md) for more information on Drush aliases. 
+See [BLT Drush Documentation](drush.md) for more information on Drush aliases.
+
+## Adding settings to settings.php
+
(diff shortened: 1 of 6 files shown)


danepowell:
@ba66e77 this time you really did write this code originally in https://github.com/acquia/blt/pull/2939, so let me know if you have any concerns :)

danepowell:
@dpagini any feedback on this?

dpagini:
Hey @danepowell - even though I just went through all the work to update to the current way, ha... this method seems to make a lot of sense to me! It's much more inline with the Drupal core way of providing that default file and renaming if you want to use it, and that makes a lot of sense to me. Thanks!

PR #3538, opened 2019-04-10
Make the composer create-project acquia/blt work on Windows

alexxed:
Changes proposed
---------
Make composer create-project acquia/blt work in native Windows, under cmd.exe or equivalent.

Steps to replicate the issue
----------
1. Run composer create-project acquia/blt on Windows under cmd.exe

----------
An error message is thrown, installation is aborted.

-----------
Installation completes successfully.

Additional details
-----------

This needs more manual testing by making composer create-project use this PR code with extra params like repository-url.


src/Robo/Commands/Blt/UpdateCommand.php
```diff
@@ -314,8 +314,8 @@ protected function convertLegacySchemaVersion($version) {
    * @throws BltException
    */
   protected function rsyncTemplate() {
-    $source = $this->getConfigValue('blt.root') . '/template';
-    $destination = $this->getConfigValue('repo.root');
+    $source = 'vendor/acquia/blt/template';
```
alexxed: @danepowell what is the impact of this? Isn't BLT always run from the root project directory? Do you need full paths?

On Windows, the problem is rsync does not work with full paths as it's running under cygwin, so whatever php returns as root path, it's different on cygwin: c:/ is /cygdrive/c/ for example.

danepowell: _Usually_ folks run it from the project root, but we support running it from anywhere below the project root. I don't think we want to break that support.

This is the function that discovers the BLT root, you can try to make it relative. Although it makes me pretty nervous to change this fundamental.

https://github.com/acquia/blt/blob/10.x/src/Robo/Config/DefaultConfig.php#L28-L45

alexxed: Got it, I'll revisit and try to just patch for Windows.

alexxed:
Well, build failures seem unrelated, maybe master is broken.

danepowell:
Yep there was an upstream bug in drupal/coder, I rebased on master which should restart the test, get the newest version of coder, and fix this.

alexxed:
Cool, the approach is not safer, with if statements and it only fixes the create-project command. Once you get past that, all other commands can be run in the vm/container.

PR #3541, opened 2019-04-10
Fixes #3511: Check Git user before committing, set dummy user if necessary

### LINKED ISSUE #3511: New project creation fails if global Git user is unconfigured
`blt internal:create-project:init-repo` fails with an error in any environment without a global Git user email configured, such as Acquia Pipelines.

To reproduce, ensure the setting is unconfigured in your test environment.

    git config --global --unset user.email

Then create a BLT project:

    composer create-project --no-interaction acquia/blt-project:10.x-dev

At the end of the process, the following results:

    > blt internal:create-project:init-repo
    [warning] The xDebug extension is loaded. This will significantly decrease performance.
    [ExecStack] git init && git add -A && git commit -m 'Initial commit.'
    [ExecStack]  Initialized empty Git repository in /Users/travis.carden/Projects/Work/test/blt-project/.git/
    [ExecStack]  Exit code 128
    [error]  Could not initialize new git repository.
    Script blt internal:create-project:init-repo handling the post-create-project-cmd event returned with error code 1

I suggest that whenever BLT wants to commit it does the following:

1. Check for the setting, i.e., `git config user.name; git config user.email`
1. If absent...
    1. configure it locally, e.g., `git config user.name BLT; git config user.email no-reply@example.com`
    1. unset it after committing, i.e., `git config --unset user.name; git config --unset user.email`

Related issues:
- https://github.com/acquia/blt/issues/2971
- https://github.com/acquia/blt/issues/2940

danepowell:
Fixes #3511
--------

Changes proposed
---------
- Better document git.user.name and git.user.email config properties in build.yml (these already existed before as an undocumented feature for deploys)
- Add a validation function (validateGitConfig) that ensures that the Git user is set correctly _either_ via the system or via the above config properties
- Use this validation function prior to deploys and new project creation (the only two places we do Git commits)
- If git.user.name and git.user.email are set, add an `--author` argument to Git commits during deploys and new project creation.
- Add an empty "precommit" during project creation to make @TravisCarden happy :smile: 

I haven't fully tested this yet, it very well might not work just yet. There's probably some opportunity to refactor the construction of the commit command. But I'd still appreciate any feedback.

@ba66e77 this is modifying a feature that you originally requested in https://github.com/acquia/blt/pull/1323 , on the off chance that that is something precious to you. I _think_ it should operate basically the same, except that we'll set the Git user on each commit rather than once for the whole repo.


config/build.yml
```diff
@@ -98,6 +98,11 @@ git:
                       fifteen characters or more describing the commit, and end with a period."
     # Provide an example of a valid commit message.
     example: "${project.prefix}-123: Update module configuration."
+  user:
+    # Name and email to use for the purposes of Git commits if you don't want to
+    # use global Git configuration.
+    name: ''
```
danepowell: Actually I don't think we ever deal with the name and email separately, so maybe just replace this with a single `git.user.author` property that should be set in the form `John Doe <no-reply@example.com>`

TravisCarden: That could be, but if you ever decide to use the values separately in the future, you'll have to pull them apart. And if you put them together in the same string, you should probably validate the value before using it to ensure the user didn't mess up the syntax. I don't care one way or another, but if it were me I'd probably leave the values separate. ¯\_(ツ)_/¯

src/Robo/Commands/Deploy/DeployCommand.php
```diff
@@ -535,11 +521,17 @@ protected function setMultisiteFilePermissions($perms) {
...
+    $git_email = $this->getConfigValue('git.user.email');
+    if ($git_name && $git_email) {
+      $author_string = sprintf('--author "%s <%s>"', $git_name, $git_email);
+    }
     $result = $this->taskExecStack()
       ->dir($this->deployDir)
       ->exec("git rm -r --cached --ignore-unmatch .")
       ->exec("git add -A")
-      ->exec(["git commit --quiet -m", escapeshellarg($this->commitMessage)])
+      ->exec(["git commit --quiet -m $author_string", escapeshellarg($this->commitMessage)])
```
alexxed: Why are you passing the author as message ? Did you want to use --author ?

src/Robo/Hooks/ValidateHook.php
```diff
@@ -131,4 +131,22 @@ public function validateVmConfig() {
...
+   * Validates that Git user is configured.
+   *
+   * @hook validate @validateGitConfig
+   */
+  public function validateGitConfig() {
+    if (!$this->getInspector()->isGitMinimumVersionSatisfied('2.0')) {
+      throw new BltException("Your system does not meet BLT's requirements. Please update git to 2.0 or newer.");
+    }
+    if (!$this->getInspector()->isGitUserSet()) {
+      if (!$this->getConfigValue('git.user.name') || !$this->getConfigValue('git.user.email')) {
```
alexxed: Can't you prompt the user for these values and set defaults in case -n ?

src/Robo/Commands/Deploy/DeployCommand.php
danepowell: Whoops, I got the order of the arguments backwards.

src/Robo/Hooks/ValidateHook.php
danepowell: Yeah, I thought about that but wasn't sure the best way. Normally we use interact hooks to get user input, but they don't run during `--no-interaction`, so we'd have to duplicate a bit of code between the interact and validate hooks, which seems needlessly complex. It doesn't seem kosher to solicit for user input during a validate hook, but I don't see a better option.

danepowell:
I refactored the Git commands into their own class, so we can set the author consistently for commits.

PR #3544, opened 2019-04-11
Allow both supported minor versions of Drupal Core 8

TravisCarden:
[Per Drupal.org](https://www.drupal.org/drupal-security-team/general-information#supported), "Starting with the release of Drupal 8.6.0, two minor releases receive security support at a time. For example, Drupal 8.6 receives security support until the release of Drupal 8.8.0." BLT should therefore remain compatible with two minor versions. This is important for ORCA, because we want to test that our product modules are compatible with all supported versions of Drupal, but we can't do so with a BLT-based test fixture if the version constraint only allows the latest version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "drupal/coder": "^8.3.1",
         "drush/drush": "^9.5.0",
         "drupal/config_split": "^1.0.0",
-        "drupal/core": "^8.6.0",
+        "drupal/core": "^8.5.0",
         "drupal/features": "^3.8.0",
         "drupal/memcache": "2.0-alpha7",
         "drupal/seckit": "^1.0.0-alpha2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "dbb71ca2256b20d47044ea0db135d2f5",
+    "content-hash": "d13acf0a616233c2c481fa29553dc1ae",
     "packages": [
         {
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -35,7 +35,7 @@
         "drupal/coder": "^8.3.1",
         "drush/drush": "^9.5.0",
         "drupal/config_split": "^1.0.0",
-        "drupal/core": "^8.6.0",
+        "drupal/core": "^8.5.0",
```
anavarre: Do we really want this? Drupal 8.7.0 is almost out of the door, which will de facto make 8.5.x deprecated entirely. What we can refrain from doing, however, is to update core to `^8.7.0` and leave it to `^8.6.0`.

TravisCarden: Well, @anavarre, I'm ready to roll out ORCA functionality that automatically selects the two supported releases via Composer and tests against them. I could roll it out today if it weren't for this line. I would prefer to make this change today and revert it a short time in the future. It's a trivial effort, and we get significant instant value from it.

TravisCarden: ...oh, and it should have no impact whatsoever on current users. Even new users will automatically get the latest version of Drupal Core by virtue of Composer's logic. To oversimplify things, this helps ORCA and has no effect on anyone else. 😃

danepowell:
I'm merging this PR solely for the purpose of temporarily unblocking ORCA. This shouldn't be taken as an indication that BLT 10.x officially supports Drupal core 8.5.x (it [officially does not](https://github.com/acquia/blt#release-support-status)). We'll revert it in coordination with ORCA once 8.7 drops and 8.5 becomes officially unsupported by core.

TravisCarden:
Thank you, @danepowell!!

PR #3545, opened 2019-04-11
Clarify composer error message

danepowell:
I don't _think_ this should have any negative effects unless someone is running this in some sort of mission critical environment where they don't want Composer to stomp on other processes. Which is maybe a scenario worth considering. I'm open to ideas.


danepowell:
Alternatively, this PR would _drastically_ reduce composer memory usage on BLT projects: https://github.com/acquia/blt/pull/3510

danepowell:
I'm taking the conservative approach here since #3510 should help a lot with memory usage.

PR #3554, opened 2019-04-14
Prevent cache collisions in multisite install tasks #3171.

egnikitin:
Fixes #3171 
--------

Changes proposed
---------
There is an issue https://github.com/acquia/blt/pull/3171 "Prevent cache collisions in multisite install tasks". It should create new temporary cache folder but actually doesn't do it.

Steps to replicate the issue
----------
When I run command similar
```
CACHE_PREFIX='/mnt/tmp/sandbox.01live' \drush8 --root='/mnt/www/html/***/docroot' -l '***.acsitefactory.com' scr --script-path='/mnt/www/html/***/factory-hooks/post-install' 'post-install.php'
```
it shows in console 

```
Running updates on: site: ***; env: 01dev; db_role: dvc***; name: ***;
Site domain:***.acsitefactory.com;
sh: 1: /mnt/www/html/***/vendor/acquia/blt/scripts/blt/drush/cache.php: Permission denied
mkdir: cannot create directory ‘’: No such file or directory
Executing: DRUSH_PATHS_CACHE_DIRECTORY='' '/var/www/html/***/vendor/bin/blt' drupal:update --environment='01dev' --site='***' --define drush.uri='***.acsitefactory.com' --verbose --yes --no-interaction with cache dir ;
``` 

----------
Temporary cache directory isn't created.

-----------
Temporary cache directory is created.

Additional details
-----------
After applying a fix post-install.php prints:
```
Post install has been started.Running updates on: site: ***; env: 01dev; db_role: dvc***; name: ***;
Site domain: ***.acsitefactory.com;
Executing: DRUSH_PATHS_CACHE_DIRECTORY='/mnt/tmp/***.01dev/drush_tmp_cache/e6ba068946f71ed14ef1dd66d91173e2' '/var/www/html/***.01dev/vendor/bin/blt' drupal:update --environment='01dev' --site='***' --define drush.uri='***.acsitefactory.com' --verbose --yes --no-interaction with cache dir /mnt/tmp/***.01dev/drush_tmp_cache/e6ba068946f71ed14ef1dd66d91173e2;
```

### DIFF
--- a/scripts/factory-hooks/post-install/post-install.php
+++ b/scripts/factory-hooks/post-install/post-install.php
@@ -64,7 +64,7 @@ if (!$uri) {
 
 $docroot = sprintf('/var/www/html/%s.%s/docroot', $site, $env);
 
-$cache_directory = exec("/mnt/www/html/$site.$env/vendor/acquia/blt/scripts/blt/drush/cache.php $site $env $uri");
+$cache_directory = exec("/usr/bin/env php /mnt/www/html/$site.$env/vendor/acquia/blt/scripts/blt/drush/cache.php $site $env $uri");
 
 shell_exec(sprintf('mkdir -p %s', escapeshellarg($cache_directory)));
 


danepowell:
Thanks for the PR.

The fix makes sense as such. I might prefer to just include the PHP file rather than shelling out to run it, but since we have precedent to invoke it via the shell (in db-update.sh), I guess this is fine.

The only thing I don't understand is why this issue hasn't been reported before now, since we've had this code in place for over 6 months now, and my understanding is that the error would occur with _any_ project using BLT on ACSF that's installed a new site. My worry is that there might be something different between ACSF environments or codebases to trigger this error, and I'd like to understand what that is.

Can anyone comment on why this is only appearing now? Is the failure not immediately obvious or something?

lcatlett:
@egnikitin which version of BLT and the ACSF module are you using when you get this error? I'm not sure the issue is creating the actual directory, or if one of the values for the required site group, environment, and uri arguments used in `cache.php` is null or not correct since in the output you posted of `Executing: DRUSH_PATHS_CACHE_DIRECTORY=''`, the cache directory variable appears to be null.

egnikitin:
Hi.

@danepowell Actually this issue doesn't trigger any errors or warnings. We found this information in log after we had tried to install 3 sites at once and installation of one site had been broken.

@lcatlett BLT version is 9.2.4, ACSF module was 2.53.0. Cache directory wasn't created before so script printed `Executing: DRUSH_PATHS_CACHE_DIRECTORY=''` after applying the patch it prints `Executing: DRUSH_PATHS_CACHE_DIRECTORY='/mnt/tmp/***.01dev/drush_tmp_cache/e6ba068946f71ed14ef1dd66d91173e2'`

danepowell:
Oh I think I understand. So it's not like this fails the task in a way that would be immediately obvious, it just fails to prevent the cache collision. So this is probably in everyone's log files and they just didn't realize to attribute the cache clear failures to it.

@lcatlett you wrote this code, did you ever test it on ACSF? Can you verify the fix?

egnikitin:
@danepowell thank you for merging this pull request. Just wanted to note that would be nice to merge it to BLT 9.2 as well.

lcatlett:
@danepowell - I validated the original functionality in #3171 on all ACSF/BLT customer codebases and the functionality worked as expected for some time, as #3171 was originally contributed in #3006 and many customers were having to patch BLT with #3006 for several months. In taking a look at the logs across customers for when there was an uptick in issues for to what now appears to be an issue impacting most - it looks like #2945 may have introduced this regression and is the likely primary root cause of the issues i raised in #3584. 

When cloud hooks and BLT commands have been firing inadvertently, this has resulted in losing the service container which drush and core assume to be available, so the recent error messages on ACSF codebases such as this make much more sense now:

```
In Drupal.php line 130:
\Drupal::$container is not initialized yet. \Drupal::setContainer() must be 
 called with a real container.
```

Drush is failing to create the directory because in some instances the variables  are now complecting with the environment so when db-update and post-install fire for a site if it happens to be the update environment context, drush is attempting to write to both a file system and a database that it thankfully does not have access to, and in some cases is returning null for values. The approach in #2945 and this 'fix' in #3554 should likely be revisited or reverted for BLT 9.2.x and 10.x to unblock customers. @danepowell when is the BLT team expecting to implement manual and automated testing for changes like this which impact ACSF? It was my understanding that for BLT and other products it is a priority to be multisite/ACSF compatible and I have not seen updates suggesting this has changed.

PR #3549, opened 2019-04-12
Alter run_tests not to use --yes as argument

nedsbeds:
Fixes #3548 
--------

Changes proposed
---------
Update the run_tests command to use --no-interaction rather than --yes on blt setup and blt tests:all

Previous (bad) behavior, before applying PR
----------
[Symfony\Component\Console\Exception\RuntimeException]  
  The "--yes" option does not exist.

Expected behavior, after applying PR and re-running test steps
-----------
Build passes

PR #3560, opened 2019-04-19
run_tests should not use the --yes argument on Pipelines

anavarre:
This is a follow-up to #3549 which addressed TravisCI only.


danepowell:
Good catch, thanks!

PR #3566, opened 2019-04-22
See what breaks if we unpin Behat version

### LINKED ISSUE #3559: DT-589: Support Behat 3.4+
I am unable to upgrade the `behat/behat` library.

Hoping we can remove the restrictions on the behat versions that appear to be put in 1.5 years ago to work with Lightning.
I see the restrictions were introduced in https://github.com/acquia/blt/pull/1993 clearly to address a problem with lightning. As I, like many other BLT users, do not use lightning, this is a tough version contraint to deal with. I also think it's very odd that BLT should be responsible for specifying versions of behat, and that cannot be left up to lightning to handle. Could they use maybe the conflicts section of their composer file? Not sure if that would work...? As I mentioned this was introduced a long time ago, so maybe this does not even exist as a problem any longer...?

Leave as is?

danepowell:
Fixes #3559
--------

Changes proposed
---------
- Allow BLT users to use Behat 3.4 or 3.5, and potentially future minor releases as well.

I have no idea what might break with this. As far as I know, no one has tested Behat versions above 3.3 with BLT. I don't see anything obviously risky in the Behat release notes. Lightning tests with Behat 3.5.0 so we should probably shoot for that as well.

### DIFF
--- a/src/Robo/Commands/Tests/BehatCommand.php
+++ b/src/Robo/Commands/Tests/BehatCommand.php
@@ -154,11 +154,11 @@ class BehatCommand extends TestsCommandBase {
       // @todo replace base_url in behat config when internal server is being used.
       $task = $this->taskBehat($this->getConfigValue('composer.bin') . '/behat')
         ->format('pretty')
-        ->arg($behat_path)
         ->option('colors')
         ->noInteraction()
         ->printMetadata(FALSE)
         ->stopOnFail()
+        ->dir($behat_path)
         ->option('strict')
         ->option('config', $this->getConfigValue('behat.config'))
         ->option('profile', $this->getConfigValue('behat.profile'))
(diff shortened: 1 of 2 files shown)


danepowell:
I confirmed BLT's internal test suite is pulling Behat 3.5.0 with this, and is passing. I'd still like more manual testing and reports from folks about whether this works in the field.

danepowell:
I can't figure out what's going on here, and I don't think it's worth spending much more time on.

@dpagini have you successfully gotten Behat 3.4 or 3.5 to run at all? They seem pretty unstable. As you can see in the tests, I'm hitting these bugs, which are pretty severe and have been open for years with no obvious solution:
- https://github.com/Behat/Behat/issues/1076
- https://github.com/Behat/Gherkin/pull/129

I debugged this for a while and got nowhere. I don't see a pressing need to spend more time on this until Behat gets that taken care of, or unless you know of a workaround.

PR #3570, opened 2019-04-23
Use a unified environment detector.

### LINKED ISSUE #3565: Cloud Hooks fail in ACSF multi-stack environments
Cloud Hooks detect whether they are running in an ACSF environment and should gracefully abort here: https://github.com/acquia/blt/blob/9.2.x/src/Robo/Commands/Artifact/AcHooksCommand.php#L85

However, apparently some ACSF environments are named like '02devup' which fails that regex. I'm not sure why, or if something changed to cause that, I'm asking around.

We should fix that, and also look for any other places in the code that detect ACSF environments. Maybe we need a unified function for that. blt.settings.php uses a file_exists and slightly different regex to check ACSF envs: https://github.com/acquia/blt/blob/10.x/settings/blt.settings.php#L100

danepowell:
Fixes #3565
--------

This fixes #3565 and also adds a utility class to encapsulate environment detection methods so that this kind of error hopefully does not recur (it could have been avoided if we hadn't implemented three different methods of ACSF env detection in three different places).

This requires a change record since any project settings relying on these variables will no longer work.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -89,8 +89,8 @@ if (EnvironmentDetector::isAcsfInited()) {
     $config_initializer = new ConfigInitializer($repo_root, $input);
     $blt_config = $config_initializer->initialize();
 
-    // The hostname must match the pattern local.[sitename].com, where
-    // [sitename] is a value in the multisites array.
+    // The hostname must match the pattern local.[site-name].com, where
+    // [site-name] is a value in the multisites array.
     $domain_fragments = explode('.', $http_host);
     $name = array_slice($domain_fragments, 1);
     $acsf_sites = $blt_config->get('multisites');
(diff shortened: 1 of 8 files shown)


danepowell:
@alexxed I'd appreciate a review. Also let me know if you have any opinions on the following:
- Should we continue to use static methods for everything here, or make them non-static and allow things like $ah_env to be set via a constructor? Currently, you have to pass variables like $ah_env to any method where you want to override what might be auto-detected, which is a little ugly but simple and performant.
- Should we just get rid of all of the $is_whatever variables in settings files, and replace them with direct calls to EnvironmentDetector methods? Seems like so...

settings/blt.settings.php
```diff
@@ -70,6 +72,7 @@
 /**
  * CI envs.
  */
+// TODO: Incorporate all of these as well?
```
alexxed: Yes, they look like they belong there.

settings/blt.settings.php
```diff
@@ -83,21 +86,23 @@
  * Note that the values of environmental variables are set differently on Acquia
  * Cloud Free tier vs Acquia Cloud Professional and Enterprise.
  */
+// TODO: Remove all of these variables, call detector methods directly instead?
```
alexxed: Yes, just call them directly.

danepowell:
Thanks a lot for the last-minute review!

lcatlett:
@danepowell - how was this deployed and tested on/for an ACSF environment? I am attempting to upgrade a test codebase and cant get drupal to bootstrap the correct site, I think because of the `$_SERVER` globals being used

PR #3572, opened 2019-04-23
Add documentation for checkDirty deploy step.

danepowell:
Hopefully this helps folks having issues with their deploys.


danepowell:
@anavarre @mikemadison13 is this helpful? Any feedback?

anavarre:
Thanks, @danepowell - I think that's helpful.

> If deploying via CI, you'll need to examine your CI scripts to determine what might be causing these files to change during the test process.

We should indicate where those scripts live and what to look out for. We should also indicate that what you see locally is not what you get in a CI env. due to those scripts.

In particular when I ran into the issue I had to `blt blt:init:settings` which I didn't know was even a thing (I never set up this BLT instance myself). We probably need to clarify that.

mikemadison13:
@danepowell only feedback is that the ignore dirty flag cannot be easily set for deployments occurring during the CI process, not sure how we would ignore it in that circumstance (or if we want to). 

otherwise this is great, thanks!

PR #3577, opened 2019-04-24
Fixes #3568: Don't run Drupal tests on Pipelines.

### LINKED ISSUE #3568: Drupal tests fail in Acquia Pipelines
If you try to run the standard `blt tests:all` suite in a 10.x project (upgraded from 9.2.x, but I think this would happen for fresh projects too) it will fail because of some sort of jankiness in the new DrupalTest command. I think it has something to do with launching / killing Selenium needlessly.

@greylabel can you comment on what might be going on here? Notice that `tests:drupal:run` has the `@launchWebServer` tag, which will start the server automatically: https://github.com/acquia/blt/blob/10.x/src/Robo/Commands/Tests/DrupalTestCommand.php#L109

But it then goes on to explicitly launch the webserver when the tests are actually run: https://github.com/acquia/blt/blob/10.x/src/Robo/Commands/Tests/DrupalTestCommand.php#L80

It seems like that shouldn't be happening twice. And is Selenium even necessary here?

danepowell:
Fixes #3568 

Pipelines doesn't have Java, so it can't run Selenium tests. Maybe this needs to be better documented somewhere as well... not sure where...

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -179,7 +179,7 @@ The `tests` configuration variable has following properties:
  * `chromedriver.port`: Port for `chromedriver` WebDriver for Chrome, default is `9515`
  * `chromedriver.args`: Args for `chromedriver` WebDriver for Chrome, default is `null`
 
-### PHPUnit
+## PHPUnit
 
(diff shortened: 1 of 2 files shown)


danepowell:
Anyone who would like to run Drupal tests in Acquia Pipelines should submit a Support ticket requesting that Pipelines include all of the dependencies (such as Java) necessary to run Selenium. Unless/until Pipelines adds that support, this isn't something that BLT can work around.

PR #3578, opened 2019-04-24
Integrate with ORCA for automated testing

TravisCarden:
This (in-progress) pull request begins implementing ORCA for automated testing for BLT. Here's what it does so far:

- It implements ORCA's standard jobs and features and isolates pre-existing testing code from them.
- It eliminates any features that ORCA already provides, such as configuring the PHP memory limit or validating `composer.json`.
- It fixes a YAML validation error that caused the static code analysis job to fail.
- It [normalizes](https://github.com/localheinz/composer-normalize) `composer.json` files.

The following remains to be done:

1. Review the changes, obviously. I'm happy to walk someone through them so they fully understand the new system. Among other things, you should consider whether or not you like where I put the new scripts (`bin/travis`).
1. Fix the two failing jobs that aren't _allowed_ to fail. They're both standard ORCA jobs, and they're failing for the same reason: they run BLT's PHPUnit tests, but they don't do the pre-test setup they depend on. (ORCA doesn't do any special setup for _any_ packages per its guiding principle: use packages as a customer would.) You can deal with this in two ways:
    1. Refactor the tests to contain their own setup so they can be safely run in standard ORCA jobs.
    1. [Tag them `orca_ignore`](https://github.com/acquia/orca/blob/v1.0.0-alpha11/docs/getting-started.md#tagginggrouping) so that ORCA doesn’t run them at all and continue to run them in your custom job.
1. Consider whether ORCA renders any of your custom tests or features redundant and remove them.
1. OPTIONAL (but appreciated): Provide feedback. :)


danepowell:
Thanks so much for getting the ball rolling on this! I've been underwater coming back from PTO and trying to get a release out the door today, but I'll review as soon as I can.

TravisCarden:
You're welcome, @danepowell! I think I've done about everything I can do on my own here. I'd love to talk through it with you sometime, and I'm happy to do anything else I can to help you get it over the finish line. Ping me when you're ready to chat! :)

danepowell:
@TravisCarden I think I'm close to getting this running, but hitting a snag with the failing tests and wondering if you have guidance.

To run all of its PHPUnit tests, BLT requires a PHPUnit bootstrap file that manages a sandbox Drupal installation to run all of the tests against. You can see that BLT requires this bootstrap file in its phpunit.xml.dist file here: https://github.com/acquia/blt/blob/a296b52c797bbedff3a5cf67bb5ba344bf38b67b/phpunit.xml.dist#L5

I _think_ the reason that ORCA tests are failing now is because they are running BLT's PHPUnit tests without using BLT's own configuration or bootstrap file. Any ideas how to fix that?

TravisCarden:
Ah. Let me look into that, @danepowell.

TravisCarden:
Okay, @danepowell. Having looked into the issue, I have a couple of thoughts: First, in principle, I don't think a bootstrap file is the place to do fixture setup like that. It's designed for autoloading and the like. It would probably be better to use a [test listener](https://phpunit.de/manual/6.5/en/extending-phpunit.html#extending-phpunit.PHPUnit_Framework_TestListener) or [`setUpBeforeClass()` template methods](https://phpunit.de/manual/6.5/en/fixtures.html#fixtures.more-setup-than-teardown). But, more practically, I don't think ORCA should be running those tests, to begin with. They're not interacting with the ORCA fixture at all. They neither require nor use ORCA for anything except the one thing it doesn't do. Any such tests I suggest [tagging `orca_ignore`](https://github.com/acquia/orca/blob/v1.0.0-alpha11/docs/getting-started.md#tagginggrouping) so they run only in your custom jobs. That should make your build pass. Then you can refactor your tests later if you choose.

danepowell:
Thanks for the feedback, we'll look into that in https://github.com/acquia/blt/issues/3630. I'm not familiar with the history of why the bootstrap was used in the first place.

Ignoring the tests worked great.

Now I just need two approving reviews before I can merge. @TravisCarden if you wouldn't mind being one of the reviewers I'd really appreciate it. @adrianthedev or @alexxed if you could be the other.

.travis.yml
```diff
@@ -1,60 +1,78 @@
+---
 language: php
-dist: xenial
 
-php:
-  - 7.1
-  - 7.2
+php: 7.1
```
alexxed: Is there a reason why this PR removes testing with 7.2?

PR #3580, opened 2019-04-24
Remove support for PhantomJS.

danepowell:
PhantomJS is long abandoned, and the most recent versions are by all accounts unstable. BLT also hasn't seen any issues related to it in several years, leading me to believe that few people are using it.

I propose dropping support for it.

If anyone is still using PhantomJS, speak now or forever hold your peace. Let me know if there's some reason we need to keep maintaining this instead of encouraging folks to move to Chromedriver and/or Selenium.


danepowell:
Thanks for the feedback everyone.

PR #3595, opened 2019-04-26
Update cloud hooks for previous 10.x users.

danepowell:


mikemadison13:
is this redundant with https://github.com/acquia/blt/blob/10.x/src/Update/Updates.php#L709 ?

danepowell:
@mikemadison13 this update hook is intended for users updating from 10.0.0-rc3 or other 10.0 releases, who need to get the changes added here: https://github.com/acquia/blt/pull/3594

Unfortunately it will be redundant for users coming from 9.2.x, but I'm not really sure what to do about that. Maybe we could encapsulate this in an update method that only runs once even when called multiple times. But it seems like a lot of effort when there's no real harm I can see in running it twice.

mikemadison13:
gotcha, yeah short of changing the update hook numbering to force the 10.x update to run for rc1, i'm not sure what else to do otherwise.

danepowell:
Okay, this stuck in my craw so I decided to fix it. This should only run updates once regardless of where users are coming from.

PR #3596, opened 2019-04-26
Recommend installing zaporylie/composer-drupal-optimizations globally.

danepowell:
BLT's composer.json already depends on zaporylie/composer-drupal-optimizations, but unfortunately because of the way Composer plugins work it has no effect on the initial install (i.e. during project creation). It does work great on subsequent updates.

We should recommend that folks install it globally before installing BLT to get a much better experience. On my machine, this reduced `composer create-project` runtime from ~60s to ~15s and memory usage from 1600mb to 400mb. Huge improvement.

We already recommend that folks install the prestissimo plugin as a prerequisite, so this seems like a natural extension.

@alexxed you might think about having ADS install these global plugins before spinning up BLT projects.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -31,7 +31,8 @@ Then install the minimum dependencies for BLT. The preferred method is via Homeb
 
         /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
         brew install php71 git composer drush
-        composer global require "hirak/prestissimo:^0.3"
+        composer global require hirak/prestissimo:^0.3
+        composer global require zaporylie/composer-drupal-optimizations:^1.1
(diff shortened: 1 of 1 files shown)


lcatlett:
Hasnt `composer global require` proven to be a challenge on BLT and in general on  large Drupal codebases? Why would BLT not use a more modern approach like the  https://github.com/consolidation/cgr package, or just running composer install on `zaporylie/composer-drupal-optimizations` via a composer plugin or script pre-install script so that it is executed before the symfony dependencies which are usually the issue? 

It is my understanding that `zaporylie/composer-drupal-optimizations` is designed to [only work in the project root](https://github.com/consolidation/cgr/issues/20#issuecomment-292761397) so this change seems like it will result in a nightmare of autoloader conflicts and issues...

PR #3599, opened 2019-04-29
Fixes #3527: Change settings file include order.

### LINKED ISSUE #3527: Settings file include order
The order that BLT includes settings files get included leaves some room for improvement, in my opinion. I am mainly referring to `blt.settings.php` where all settings files from a projects `docroot/sites/settings/*` get included.

So we have a situation where we load settings from BLT, then settings from the project, then more BLT settings (in the form of CI files), then the local includes. I am suggesting we reverse the order of the CI to be above the `docroot/sites/settings/*` load. I would think we should load all BLT settings _first_, which gives projects **more** flexibility to override BLT specific things.

I ran into this issue because I just switched over to using the `docroot/sites/settings/*.settings.php` load provided by BLT as opposed to including all my project specific settings from `docroot/sites/$sitedir/settings/includes.settings.php`, which is also loaded by BLT. Previously, I had a global setting file that set my CI database to the site directory name. When I switched the new BLT way, that file was being included before the CI specific BLT files, which were then overwriting my database name to a static name for all sites.

I know this is a specific use case, but it just seems to me like it would make more sense to load ALL library (BLT) settings files first, including CI, and then let the project take-over.

N/A

danepowell:
Fixes #3527
--------

Changes proposed
---------
- CI-specific settings were included immediately before site-specific settings. Now CI settings are _after_ site settings.
- You can now override these CI settings by creating `sites/settings/ci.settings.php`
- Stop globbing settings files in `sites/settings/*.settings.php` to avoid picking up things like local.settings.php, ci.settings.php, etc... For consistency with how site-specific includes work, suggest that users define an array of files instead of globbing.

Additional details
-----------
@dpagini will this accommodate your use case? I need to know by tomorrow so we can get it into the release Wednesday. cc @greylabel since you were involved in the original discussion.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -176,37 +176,25 @@ if (file_exists($deploy_id_file)) {
 }
 
 /**
- * Include custom global settings file.
+ * Include custom global and site-specific settings files.
  *
  * This provides an opportunity for applications to override any previous
  * configuration at a global or multisite level.
- *
- * This is being included before the CI and site specific files so all available
- * settings are able to be overridden in the includes.settings.php file below.
  */
 $settings_files[] = DRUPAL_ROOT . '/sites/settings/global.settings.php';
-
-/**
- * Include optional site specific includes file.
- *
- * This is intended for to provide an opportunity for applications to override
- * any previous configuration.
- *
- * This is being included before the local file so all available settings are
- * able to be overridden in the local.settings.php file below.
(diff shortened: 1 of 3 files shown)


dpagini:
Yes, this would work great. I think not globbing makes a lot of sense along with this change... make users be somewhat responsible for adding a global settings file.

PR #3603, opened 2019-04-30
Speed up Composer operations in Travis.

danepowell:
I'm curious how much this will speed up our tests, which churn quite a bit on Composer operations. We should probably merge it regardless, since we officially recommend these packages, so we should also test with them.


danepowell:
Shaves off about a minute, or ~10 minutes across the matrix. I'll take it!

PR #3613, opened 2019-05-02
Fixes #2938: Support running BLT outside of a VM.

### LINKED ISSUE #2938: Allow users to run BLT commands from their host environment
## User Story

As a user of BLT, I want to run BLT commands from my host environment so that I don't have to keep ssh'ing into the VM.

## Discussion

In BLT 8x, we allowed users to run blt commands from their host environment. That feature was dropped in BLT 9x and users are now required to ssh into their VM to run most commands. That is an impediment to the development process and we've gotten pretty dissed for going that way.  Let's reverse course and go back to allowing blt commands from the host machine.

danepowell:
Fixes #2938
--------

Changes proposed
---------
- Add a configuration parameter `vm.blt-in-vm` (default: true) that you can disable if you want to run BLT on the host machine instead. This will just get rid of the warnings when you run commands on the host.
- Get rid of the validateMysqlAvailable hook, which was kind of a clumsy check to begin with. Not just in this use case, but other ones too (such as running blt commands against a remote alias). If that's really something we need, there are probably more targeted ways to do it.

Steps to replicate the issue
----------
1. Apply this patch.
2. In blt.yml, change `drush.aliases.local` to match the alias for your VM (i.e. `foo.local`)
3. In blt.yml, set `vm.blt-in-vm` to false.
4. Try running commands like `blt setup` from the host machine.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -264,5 +264,9 @@ validate:
 vm:
   enable: false
   config: ${repo.root}/box/config.yml
+  # By default, BLT should only be run from inside a VM, if it exists. You can
+  # choose to run it on the host instead by setting this to false, and changing
+  # drush.aliases.local to an alias that can SSH into the VM.
+  blt-in-vm: true
   vagrant:
       hostname: ${project.local.hostname}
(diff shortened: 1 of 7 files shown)


danepowell:
This would actually be really handy to test some other issues since it avoids having to mount BLT inside the VM (pain). So I'm going to go ahead and merge. But I still welcome any feedback if there's ways to make it better.

Issue #3204, opened 2018-11-01
Drush alias failure with consolidation/site-alias requires change to BLT


shelane:
The lightning profile includes consolidation/site-alias in composer requirements. Apparently, BLT has to change how drush alias files are created.
https://github.com/consolidation/site-alias/issues/18

greg-1-anderson:
I decided to go ahead and roll back the breaking bugfix on the 1.x branch. BLT users may now update to site-alias 1.1.10 to continue using the current version of BLT. Pinning to 1.1.5 is still also possible, but updating to 1.1.10 is preferred.

BLT should update to site-alias ^2 as soon as convenient, and make the changes described above. There will be no further development on the 1.x branch.

mikemadison13:
@greg-1-anderson  thanks for jumping in on this. i think we can update the version no problem. I scanned our codebase and the `$alias->root()` method didn't immediately jump out at me, is there anything we need to do other than update consolidation/site-alias? 

The only real place BLT actually "generates" aliases is here:
https://github.com/acquia/blt/blob/10.0.x/src/Robo/Commands/Generate/AliasesCommand.php

greg-1-anderson:
Try upgrading to ^2; then the exception that @shelane encountered will return. I think she may have more info about the code path in blt where this is encountered.

shelane:
I simply tried to do an install with `blt drupal:install` which ended up returning an error:
```
In AliasRecord.php line 76:
                                                      
  [Exception]                                         
  Site alias @default.local does not specify a root.  
```
Details are listed in the linked issue for consolidation/site-alias in my original message above.

mikemadison13:
@shelane how long has it been since you generated your aliases (what version of blt / drush, etc.)? 

i am looking at my project alias for what i'm on right now, and i have roots defined for all my aliases for both local and acquia cloud alias files, and i see it as an option here:

https://github.com/acquia/blt/blob/10.0.x/src/Robo/Commands/Generate/AliasesCommand.php#L339

shelane:
Drupal version   : 8.6.2  
BLT 9.2.1

I just deleted my local.drush.yml file, ran ` blt:init:settings` and a new drush file was created with no root.

mikemadison13:
Ah, I see the issue. Was thinking actual drush aliases, not the local.drush file. Thanks! Will roll a PR for testing shortly.

mikemadison13:
Actually, it looks like drush has this locked into version 1.5.x... 

[...]

PR #3614, opened 2019-05-03
Improve multisite generate command.

danepowell:
There were a few odd things happening with this command. Mainly, it wasn't setting the local db settings correctly. Also it was setting the remote drush alias as local. Defaults could be set better. Lots of code standards violations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
     - COMPOSER_BIN=$TRAVIS_BUILD_DIR/vendor/bin
     - BLT_DIR=$TRAVIS_BUILD_DIR
     - DRUPAL_CORE_HEAD=^8.7.0-alpha1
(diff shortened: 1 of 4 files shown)


danepowell:
See if we can fix generated local Drush aliases to include a `root`, pending discussion in #3204

danepowell:
Bummer... this is now actually blocked by #3204. I think the problem is that we changed the default local alias from `self` to `@foo.local`, which wouldn't be a problem except that we're hitting the bug in #3204.

greg-1-anderson:
Workaround: if you are expecting the cwd to be always at the project root, and your Drupal root is always at docroot, then you could set root to `${env.CWD}/docroot` for your local aliases. Didn't try it, but should work.

danepowell:
Awesome thanks, I think it's actually `${env.cwd}` (lowercase) but that did the trick.

PR #3619, opened 2019-05-03
Fixes #3618 to set composer to 10.x instead of 10.x-dev.

### LINKED ISSUE #3618: BLT 10.x Rewrites composer.json to 10.x-dev
**Describe the bug**

**To Reproduce**
1. Update composer.json to `acquia/blt: ^10.0`
2. Run `composer update acquia/blt --with-all-dependencies`
3. After blt update runs
4. Verify composer.json has now changed

**Expected behavior**
BLT shouldn't change my stable ^10.0 to 10.x-dev

**Detailed error output**
N/A

**BLT doctor output**
N/A

mikemadison13:
Fixes #3618 
--------

Changes proposed
---------

- sets the template composer.json to ^10.0 instead of 10.x-dev

### DIFF
--- a/subtree-splits/blt-project/composer.json
+++ b/subtree-splits/blt-project/composer.json
@@ -32,7 +32,7 @@
     },
     "require": {
         "php": ">=7.1",
-        "acquia/blt": "10.x-dev",
+        "acquia/blt": "^10.0",
         "acquia/drupal-spec-tool": "^2.0.0",
         "acquia/lightning": "^3.1.0",
         "drupal/acquia_connector": "^1.5.0",


danepowell:
Confirmed the bug. I think we should take a different approach to fix, which would be to fix the update hook merging this template composer.json.

Using 10.x-dev for blt-project is necessary to make sure that users of `acquia/blt-project:10.x-dev` can pull the dev version of BLT as well. We temporarily change this to a stable version only when tagging new stable release of blt-project. I know it's a little confusing as a maintainer but it's better for end users (although I'm open to suggestions).

This is described in the RELEASE workflow: https://github.com/acquia/blt/blob/10.x/docs/RELEASE.md#update-the-blt-project-repo

PR #3625, opened 2019-05-04
Tracing the Drupal requests in drupal-requests.log on acquia cloud.

joshirohit100:
Acquia cloud has drupal-requests.log as extra logs and this will trace drush command in drupal-requests.logs and makes it easy to identify.


danepowell:
Thanks for the contribution. In order for this to be merged, it's just going to need a few things:
1. This settings file won't have any effect on its own. Presumably it needs to be included somewhere. I would suggest [blt.settings.php](https://github.com/acquia/blt/blob/10.x/settings/blt.settings.php) if it should always be active.
2. If this isn't going to "just work" out of the box, i.e. because people need to specifically include it as mentioned above, then it needs documentation to that effect.
3. I need a set of testing steps so I can verify that this operates as intended (specifically, what commands you are running in Acquia Cloud to test this and examples of the log output before and after).

danepowell:
Thanks again for the contribution and testing steps, they were helpful. I confirmed it works as expected and improves log output.

It seems like this functionality should really be built into Cloud, not provided via BLT. I'll ping you on an internal ticket regarding that.

We just need one other review before this can be merged.

PR #3629, opened 2019-05-07
Preserve comments in DrupalVM config when adding multisites.

danepowell:
Workaround for lack of comment support in Symfony YAML parser: https://github.com/symfony/symfony/issues/22516

This isn't tied to a particular ticket, just something that bothered me when refactoring the multisite command previously.

To test, run `blt multisite` on a BLT project using DrupalVM, and opt to let it update your DrupalVM configuration. You should see that most (not all) of the comments in `box/config.yml` are preserved.


danepowell:
I think the failing tests are unrelated but I'm looking into it

danepowell:
We can't test DrupalVM in TravisCI or Pipelines. The best we could do is to write a unit test with a mock DrupalVM config file to make sure `consolidation/comments` rewrites it as expected. Of course, at that point we've basically just duplicated `consolidation/comments`'s own tests: https://github.com/consolidation/comments/blob/master/tests/TestYamlComments.php

Seems like that would have negligible value.

alexxed:
Right, I was thinking of a test that edits a regular drupalvm yml file with comments and checks that it does not break it, e.g. only what was changed is changed.

PR #3632, opened 2019-05-08
Fixes #3617: Run cache clears for config strategy 'none'.

### LINKED ISSUE #3617: Follow up to #3363: Drush CR
**Describe the bug**
I started using the config strategy 'none' and there were some changes made to make sure that `drush updb` was run when the config strategy was none. These work great.
However, during our production deploys, there is now no longer a `drush cr` performed for our sites since it only gets run in features/config import steps.
I have some small hesitation here, wondering if a `drush cr` is _really_ even needed. I guess if we have new frontend assets, for example, we want those being served up immediately after a deployment, so I think we do want that.

**To Reproduce**
1. From a BLT project...
2. Set config strategy to 'none'
3. When an artifact is deployed on ACE, cloud hooks will run `blt artifact:ac-hooks:post-code-deploy ...` and no drush CR will be performed.

**Expected behavior**
A general cache-clear should take place to ensure the deployment works as promised.

danepowell:
Fixes #3617
--------

Changes proposed
---------
Run cache clear even if config strategy is none.
If config strategy is anything other than 'none', the only functional change is that the cache clear now happens after the post-config-import step. I doubt this will be breaking change for anyone.

### DIFF
--- a/src/Robo/Commands/Setup/ConfigCommand.php
+++ b/src/Robo/Commands/Setup/ConfigCommand.php
@@ -200,11 +200,9 @@ class ConfigCommand extends BltTasks {
   /**
    * Checks whether core config is overridden.
    *
-   * @param string $cm_core_key
-   *
    * @throws \Acquia\Blt\Robo\Exceptions\BltException
    */
-  protected function checkConfigOverrides($cm_core_key) {
+  protected function checkConfigOverrides() {
     if (!$this->getConfigValue('cm.allow-overrides') && !$this->getInspector()->isActiveConfigIdentical()) {
       throw new BltException("Configuration in the database does not match configuration on disk. This indicates that your configuration on disk needs attention. Please read https://github.com/acquia/blt/wiki/Configuration-override-test-and-errors");
     }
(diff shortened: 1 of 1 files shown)


danepowell:
@mikemadison13, @dpagini, @alexxed can you please review?

mikemadison13:
I'm skeptical about removing the cache rebuild post config import. I don't love the redundant cache rebuild but... I think importing config and not clearing cache is also bad

dpagini:
I mean, this makes sense to me and solves my issue... but I know I'm in the minority here with the `none` config strategy...

danepowell:
That's fine, I think we can satisfy everyone at the cost of a little cyclomatic complexity. The latest commits preserve the old behavior for users of configuration management, and only add a cache clear if the config strategy is none.

src/Robo/Commands/Setup/ConfigCommand.php
```diff
@@ -46,68 +48,74 @@ public function update() {
...
-      $cm_core_key = $this->getConfigValue('cm.core.key');
-      $this->logConfig($this->getConfigValue('cm'), 'cm');
-      $task = $this->taskDrush();
+    if ($strategy == 'none') {
+      // Still clear caches to regenerate frontend assets and such.
+      $result = $this->taskDrush()->drush("cache-rebuild")->run();
+      return $result;
+    }
 
-      $this->invokeHook('pre-config-import');
```
danepowell: Sorry the git diff is mangled from here on. All I did was to remove the rest of this function from the conditional.

PR #3634, opened 2019-05-09
Adds the --copy-links flag to copy symlinks as files.

### LINKED ISSUE #3635: Add the --copy-links flag to Deploy command's rsync
Symlinks in the `docroot` folder are currently copied as symlinks, which has the potential for unintended side effects.
Add the `--copy-links` flag to `DeployCommand`'s `rsync` call.

jhedstrom:
Fixes #3635  
--------

I am integrating BLT into a project that utilizes symlinks of custom code into `docroot`. Adding this flag allows the `deploy` command to still work as expected since it treats those symlinks as files and correctly copies them into the artifact.

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -365,7 +365,7 @@ class DeployCommand extends BltTasks {
 
     $this->setMultisiteFilePermissions(0777);
     $this->say("Rsyncing files from source repo into the build artifact...");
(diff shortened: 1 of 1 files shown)


danepowell:
I'm just trying to anticipate any negative side effects this might have. The only thing I can think of is that the size of the repo might increase slightly if files symlinked from within the repo are now duplicated. Is there any other way this might impact unsuspecting users?

jhedstrom:
That's a good question. Now that I have been thinking about it, I'm not even sure this is necessary since the `deploy` command bundles up everything, not just the `docroot` directory.

danepowell:
I'm closing this for now since the use case isn't clear, and we haven't been able to demonstrate that the impact on existing projects would be minimal. If we can resolve those issues feel free to reopen.

PR #3640, opened 2019-05-11
Don't use bootstrap for sandboxmanager and phpunit tests.

### LINKED ISSUE #3630: Refactor how sandboxmanager is bootstrapped in PHPUnit tests
Feedback received regarding how the sandbox manager is bootstrapped in #3578:

> In principle, I don't think a bootstrap file is the place to do fixture setup like that. It's designed for autoloading and the like. It would probably be better to use a [test listener](https://phpunit.de/manual/6.5/en/extending-phpunit.html#extending-phpunit.PHPUnit_Framework_TestListener) or [`setUpBeforeClass()` template methods](https://phpunit.de/manual/6.5/en/fixtures.html#fixtures.more-setup-than-teardown).

The challenge is that the sandbox needs to be set up only once for the entire test suite, whereas PHPUnit setUp methods are centered around individual tests, which I'm supposing is why it was handled via the bootstrap file. 

Might need to get creative to make that work. Maybe if we define setUpBeforeClass() as a static method in BltProjectTestBase? I haven't looked to see if it can actually be used statically.

danepowell:
Fixes #3630
--------

Changes proposed
---------
- Stop using a bootstrap file to set up the master test sandbox, in order to make our tests self-contained and exportable to e.g. ORCA in the future


danepowell:
@TravisCarden since you propose this change, if you'd be so kind as to take a quick look.

danepowell:
I guess another way to do this would have been with a test listener. Not sure if one way or the other is necessarily better.

tests/phpunit/src/BltProjectTestBase.php
```diff
@@ -57,10 +57,24 @@ abstract class BltProjectTestBase extends TestCase {
...
+   * Track whether our master sandbox has been initialized.
+   */
+  protected static $initialized = FALSE;
+
   /**
    * {@inheritdoc}
+   * @throws \Exception
    */
   public function setUp() {
+    if (!self::$initialized) {
```
alexxed: Since you don't need anything on the instance, why not put this in the static `setUpBeforeClass()` which is called once before the first test?

danepowell: The problem with that is that BltProjectTestBase is never invoked directly, it's only a base class for all of our other test cases. _Each time_ one of those classes is run, it will also run setUpBeforeClass(), which is a huge waste of resources. It only needs to run once for the entire suite.

This is the only way I know of to ensure this only runs once for the entire suite instead of once per class.

alexxed: That makes sense. Still, if you put the same solution in setUpBeforeClass it will be less calls to this method, but that's nitpicking now.

You could put this in a static method so that its obvious why, as you said in this comment, e.g. `bootstrapSandboxOncePerSuite()`

danepowell: Yep you are right. Thanks for being persistent :) I swear I tried using setupBeforeClass originally and it didn't work for some reason, but I just tried it again and it worked great :man_shrugging: I'll push a change.

PR #3641, opened 2019-05-13
Fixed infinite loop of api calls from setting credentials

raghunat:
BLT shouldn't infinitely loop while trying to set API credentials. The setApiClient method actually tries to make API calls to ensure they are valid, so we should probably break after a few times.

---------

- Break the loop after 5 times, and alert the user that it failed when generating aliases.
- Suggest they check their provided credentials.

Steps to replicate the issue
----------
1.  Generate aliases without a creds file present
2. Enter incorrect credentials

----------
It would infinitely loop

-----------
It would error out, not generate aliases, and warn to check credentials


danepowell:
This whole 'while' loop just seems janky and unnecessary. Additionally, I think we could handle errors better in the API call itself. Can you review #3646 as an alternative solution to this?

PR #3646, opened 2019-05-13
Fixed drush alias generation loop.

danepowell:
Changes proposed
---------
- Fix infinite loop that could occur when entering invalid credentials
- Improve overall error handling of API connection

Steps to replicate the issue
----------
1. Ensure you have no cached credentials at `~/.acquia/cloud_api.conf`
1. Run `blt recipes:aliases:init:acquia`
2. Enter invalid creds ("asdf" / "asdf")

----------
Infinite loop of somewhat cryptic error messages

-----------
A single clear error message, and command exits cleanly:
> Invalid credentials. Failed to authenticate with Acquia Cloud API.

Additional details
-----------
Replaces #3641

### DIFF
--- a/src/Robo/Commands/Generate/AliasesCommand.php
+++ b/src/Robo/Commands/Generate/AliasesCommand.php
@@ -3,6 +3,8 @@
 namespace Acquia\Blt\Robo\Commands\Generate;
 
 use Acquia\Blt\Robo\BltTasks;
+use Acquia\Blt\Robo\Exceptions\BltException;
+use Acquia\Hmac\Exception\MalformedResponseException as MalformedResponseExceptionAlias;
 use AcquiaCloudApi\CloudApi\Client;
 use AcquiaCloudApi\CloudApi\Connector;
 use Symfony\Component\Yaml\Yaml;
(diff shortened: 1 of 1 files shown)


src/Robo/Commands/Generate/AliasesCommand.php
```diff
@@ -146,19 +148,20 @@ protected function loadCloudApiConfigFile() {
...
     $this->say("You may generate new API tokens at <comment>https://cloud.acquia.com/app/profile/tokens</comment>");
     $key = $this->askRequired('Please enter your Acquia cloud API key:');
     $secret = $this->askRequired('Please enter your Acquia cloud API secret:');
-    do {
-      $this->setCloudApiClient($key, $secret);
-      $cloud_api_client = $this->getCloudApiClient();
-    } while (!$cloud_api_client);
-    $config = array(
+
+    // Attempt to set client to check credentials (throws exception on failure).
```
raghunat: Hmm, this removes the loop entirely, which perhaps is probably okay, but I am curious as to why we had to add a loop in the first place?

danepowell: Grasmick originally wrote this command along with the loop here :https://github.com/acquia/blt/pull/2168

There's no documentation there as to why a loop was used. I vaguely recall something about authentication sometimes failing on the first try due to problems with the hmac library, but I think all of those issues have since been resolved. I tested this extensively and didn't experience any problems authenticating on the first try (indeed, if the API was behaving in a stateful way like that it would be a much bigger issue).

raghunat: Looks good to me then!

PR #3648, opened 2019-05-14
Combine template with blt-project subtree split.

danepowell:
Currently we store "template" files for new BLT projects in two places. The older / original place that we stored these is in the `/template` directory, which has pretty much everything except the template composer.json, which lives in `subtree-splits/blt-project`.

Now that blt-project is a subtree-split of BLT and can be managed at the same time, it seems like we should keep all template files in a single directory to simplify things. This would essentially make `blt-project` the canonical reference for what an unbuilt BLT project should look like, which seems in keeping with its name and purpose.

@alexxed @grasmash does this seem wise to you?

The only thing I haven't figured out is how to manage the README for blt-project (the repo, as it appears on Github) separately from the template README that should go into created projects.


alexxed:
Sounds good to me.

src/Robo/Commands/Blt/UpdateCommand.php
```diff
@@ -351,26 +342,6 @@ protected function convertWindowsPathToCygwinPath($path) {
     return str_replace('\\', '/', preg_replace('/([A-Z]):/i', '/cygdrive/$1', $path));
   }
 
-  /**
-   * Updates project BLT .yml files with new key value pairs from upstream.
-   *
-   * This WILL NOT overwrite existing values.
-   */
-  protected function mungeProjectYml() {
```
danepowell: I removed this function because I can't imagine why a project would have a pre-existing blt.yml file that contains customizations that need to be preserved.

danepowell:
Eep... stumbled onto some dark corners in the project sandboxing and creation process that I had to clean up to make this work, but it should be all good to go now. Would still appreciate a review especially from @grasmash who wrote all this in the first place.

danepowell:
Bah... looks like we'll also need to patch ORCA. It currently clones `acquia/blt-project` directly from Github, when I think it should instead clone it out of the subtree of `acquia/blt` for situations like this where we are actively developing on it. Heads up @TravisCarden . I'll be diving deeper on this tomorrow.

.travis.yml
```diff
@@ -61,6 +61,9 @@ matrix:
 before_install:
   # Exit build early if only documentation was changed in a Pull Request.
   - source ${BLT_DIR}/scripts/travis/exit_early
+  # Prepare local testing version of blt-project so it can be used by ORCA.
+  - cp -r subtree-splits/blt-project /tmp/
+  - composer config --global repositories.blt path /tmp/blt-project
```
TravisCarden: That is _beautiful_ simplicity. I can't wait to see how this works.

danepowell: I know, right? I was shocked that this worked on the first try!

danepowell:
Now I'm discovering one reason we may have not done this before... unavoidably, there are certain artifacts that _have_ to exist in a Github repo that aren't necessarily going to be part of the project template itself. Things like the LICENSE and README, which are different than e.g. the README you'd want for an actual built project. There's only a few of these exceptions though and I accounted for them, so I think this is still worth it.

PR #3649, opened 2019-05-14
Improve dev:link-composer command.

danepowell:
The `blt:dev:link-composer` command is incredibly useful when developing BLT itself. It lets you link a local development version of BLT into a BLT-based project via Composer.

This improves it in a few ways:
1. Works around an issue where composer fails to create the symlink via the path repo when the package already exists on disk (requiring removing the vendor diretory)
2. Provides default `blt-path` argument if not set.
3. Mounts the local version of BLT in the VM if possible.

The last one is super tricky to do unless directories are arranged in just the right way, which is why for now we only support this when BLT exists at `../../packages/blt`. Theoretically we could be a little more flexible but it would take some really tricky heuristics to figure out the right paths and map them to what's possible in DrupalVM.

To test this PR, you'd need to start with a new BLT project with DrupalVM. Then both (1) apply this patch to BLT in that project via composer, and (2) check out the danepowell:issue-devcommand branch in a BLT repo located at a relative path of `../../packages/blt`. Then you can run `blt blt:dev:link-composer` in the project and it should both symlink in the development version of BLT and mount it via Vagrant.


alexxed:
Can you give an example with full paths, I don't get why the double parent to packages.

danepowell:
Yeah of course. Sorry I know it's a little confusing.

On my machine in my home directory, I have a `packages` directory that contains any packages/libraries I'm developing, and a sites directory that contains any fully-built Drupal codebases that I'm developing on.
```
~
├── packages
|   ├── blt
|   ├── blt-project
|   ├── lightning
├── sites
|   ├── blt-canary
```

So in that case, you'd be running this command from the `blt-canary` directory.

alexxed:
Right, that explains it. It might be a good idea to have just one condition, blt-canary in the same dir as blt-canary, as I suspect that's pretty common. You have a more fine grained system. I'll give it a test.

danepowell:
I pushed a change to support the scenario of having the `blt` directory as a sibling to your project directory. i.e. `../blt` instead of `../../packages/blt`.

alexxed:
Manual test:

I had blt 10.0.1 installed, and patching failed:
```
Updating dependencies (including require-dev)
    1/1:	https://codeload.github.com/acquia/blt/legacy.zip/8a7e9bc99715875ef9550359857ed239ced34bf8
    Finished: success: 1, skipped: 0, failure: 0, total: 1
Package operations: 1 install, 1 update, 0 removals
Gathering patches for root package.
Gathering patches for dependencies. This might take a minute.
  - Installing acquia/blt (10.0.1): Loading from cache
  - Applying patches for acquia/blt
    https://patch-diff.githubusercontent.com/raw/acquia/blt/pull/3649.patch (testing-patch)


                                                                       
  [InvalidArgumentException]                                           
  Package is not installed: acquia/blt-10.9999999.9999999.9999999-dev  
```

I updated acquia/blt to 10.0.0-dev and then patching worked. I'm not sure if that's right.

```
acquia@localhost vagrant1]$ vendor/bin/blt blt:dev:link-composer
[warning] Drupal VM is locally initialized, but is not running.
[Exec] Running rm -rf vendor &&
 composer update acquia/blt --with-dependencies in /home/acquia/vagrant1
Dependency "acquia/drupal-spec-tool" is also a root requirement, but is not explicitly whitelisted. Ignoring.
Dependency "drupal/core" is also a root requirement, but is not explicitly whitelisted. Ignoring.
Dependency "acquia/drupal-spec-tool" is also a root requirement, but is not explicitly whitelisted. Ignoring.
Dependency "drupal/core" is also a root requirement, but is not explicitly whitelisted. Ignoring.
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 236 installs, 0 updates, 0 removals
  - Installing zaporylie/composer-drupal-optimizations (1.1.0): Loading from cache
...
See https://www.drupal.org/security-advisory-policy for more information.
PHP CodeSniffer Config installed_paths set to ../../drupal/coder/coder_sniffer/
Updating BLT templated files...
 > blt blt:update --ansi --no-interaction
[warning] Drupal VM is locally initialized, but is not running.
Copying files from BLT's template into your project...
Removing deprecated files and directories...
> blt:init:shell-alias
The BLT alias is already installed and up to date.
[Exec] Done in 44.343s
[Exec] Running vagrant halt &&
 vagrant up in /home/acquia/vagrant1
```

Unfortunately I'm on kernel 5.1 and VirtualBox's module doesn't compile, so doesn't work.

Looking at box/config.yml, the synced folder don't look right:
```
vagrant_synced_folders:
    -
        local_path: .
        destination: /var/www/vagrant1
        type: nfs
    -
        local_path: ../../packages/blt
        destination: /var/packages/blt
        type: nfs
```

I have /home/acquia/blt and /home/acquia/vagrant1 (my blt project).

adrianthedev:
One thing that I've noticed is that it will write to the box config.yml even if the config is there already.

image

src/Robo/Commands/Blt/DevCommand.php
```diff
@@ -13,10 +13,14 @@ class DevCommand extends BltTasks {
...
+   *
    * @command blt:dev:link-composer
+   *
    * @param array $options
    */
-  public function linkComposer($options = ['blt-path' => InputOption::VALUE_REQUIRED]) {
+  public function linkComposer($options = ['blt-path' => '../../packages/blt']) {
     $composer_json_filepath = $this->getConfigValue('repo.root') . '/composer.json';
     $composer_json = json_decode(file_get_contents($composer_json_filepath));
     $composer_json->repositories->blt = [
```
danepowell: I just realized that instead of muddling with json here, I think I can just shell out to `composer config repositories.blt path $options['blt-path']`

danepowell:
@adrianthedev yeah and additionally it nukes any comments in config.yml. I wonder if I need to abstract the comment handling from #3629 so we can use it here.

danepowell:
Okay I added logic to prevent writing multiple duplicate nfs mounts, and to preserve comments in the config.yml file.

danepowell:
BTW I'm aware that this produces some non-functional code style changes in config.yml, that's not something we can control unfortunately: https://github.com/geerlingguy/drupal-vm/issues/1932

danepowell:
@adrianthedev @alexxed please review when you can, let me know in slack if you need more help testing

adrianthedev:
I think it didn't work for me before because I tried running it with an absolute path `/Users/adrian/packages/blt` (with the project being on another drive, not somewhere in my home directory).

PR #3657, opened 2019-05-15
Standardize on \Drupal instead of Drupal.

danepowell:
Looks like core has decided to standardize on `\Drupal` instead of `Drupal` even when referencing in a procedural context. This should have no functional change but just aligns with core standards. This might require an upstream change to Coder standards since I'm not sure if they are in line with this standard.


danepowell:
Upstream issue: https://www.drupal.org/project/coder/issues/3054969

PR #3662, opened 2019-05-17
Fixes #3661: Abort deploy if Git repo is inaccessible.

### LINKED ISSUE #3661: BLT deploy does not stop on error event
**I want to...**
I have BLT tool integrated into our CI/CD process. One of the pipeline steps is to make sure `deploy --dry-run` is ok.
I have removed ssh key for the repository to imitate a git fetch fail.

**It's not working because...**
Whole command `blt artifact:deploy -vvv --ignore-dirty --no-interaction --dry-run` finshed succesfully, but here is what I have in the verbose output:
```
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
[ExecStack]  Exit code 128  Time 1.493s
```
but it kept on executing the next commands and eventually gave me the state "Done".  Jenkins detects this step as a successful one.

I expect when I run BLT deploy, any failed Robo Tasks would stop the whole process and report back the general fail. Is there any possibility of configuring this behavior?

Please let me know if I'm missing something.

danepowell:
Fixes #3661 
--------

Changes proposed
---------
- Instead of blindly fetching and merging from a Git remote branch, make sure that a remote branch actually exists first
- Fail the deploy task if the fetch and merge fail.
- Also fail the deploy task if the later checkout fails.

Steps to replicate the issue
----------
1. Add a bogus remote to blt.yml so that git fetches fail.
2. Run `blt artifact:deploy --ignore-dirty --dry-run`

----------
Deploy task finishes with exit code 0

-----------
Deploy task fails.

Additional details
-----------
In searching for other places where the deploy task might pass even though subcommands fail, I found one other instance (the `git checkout -b` step). That was originally allowed to fail in https://github.com/acquia/blt/pull/373 . There's not much context there, I suspect it was specific to an older version of Git. With lots of testing on git 2.20.1, I could never get `git checkout -b` to fail even if the target branch already existed.

@dimdiden can you please review and test?

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -259,7 +259,6 @@ class DeployCommand extends BltTasks {
    * Adds remotes from git.remotes to /deploy repository.
    */
   protected function addGitRemotes() {
-    // Add remotes and fetch upstream refs.
     $git_remotes = $this->getConfigValue('git.remotes');
     if (empty($git_remotes)) {
       throw new BltException("git.remotes is empty. Please define at least one value for git.remotes in blt/blt.yml.");
@@ -271,9 +270,10 @@ class DeployCommand extends BltTasks {
 
   /**
    * Adds a single remote to the /deploy repository.
+   *
+   * @param $remote_url
    */
   protected function addGitRemote($remote_url) {
-    $this->say("Fetching from git remote $remote_url");
     // Generate an md5 sum of the remote URL to use as remote name.
     $remote_name = md5($remote_url);
     $this->taskExecStack()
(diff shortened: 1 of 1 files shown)


dimdiden:
@danepowell Thank you for this fix! I have tested it and now it works as expected.

However, I've noticed another bugs with the same behavior:
I'm running source code inside a `php docker container` with the volume bound to this source code
1. After the stage `[ExecStack] git ls-remote --exit-code --heads...` I see the following Exit code:
```
Warning: Permanently added the RSA host key for IP address  to the list of known hosts.
[ExecStack]  Exit code 2  Time 2.323s
```
and it doesn't stop deploy process. I believe the correct behavior would be to return no errors as adding  RSA key is not an error.

2. On the stage `[ExecStack] rsync -a --no-g --delete --delete-excluded...` I have rsync issue (probably because of running inside docker container): 
```
rsync: failed to set permissions on "/home/wodby/app/deploy/vendor/acquia/blt/docs/drush.md": Not supported (95)
rsync: failed to set permissions on "/home/wodby/app/deploy/vendor/acquia/blt/docs/git-hooks.md": Not supported (95)
```
which ends up with the following error:
```
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1189) [sender=3.1.3]
[ExecStack]  Exit code 23  Time 10:42
```
But this doesn't stop the deploy process and it finishes without issues.

Let me know if you need any additional details on that matter. Thank you.

danepowell:
The first error is expected and only indicates that the branch you are pushing to doesn't exist remotely. I don't think it has anything to do with the fingerprint warning. The error shouldn't appear unless you are using very verbose logging, and it's not something we can squelch entirely due to limitations in the Robo result logger.

The second error seems more related to https://github.com/acquia/blt/pull/3670

PR #3666, opened 2019-05-20
Minor settings cleanup: removed unnecessary logic.

danepowell:
This conditional seems unnecessary and in fact risky. Drupal hasn't generally bootstrapped far enough at this point to allow access to the service container, so if the logic in this conditional ever actually got called it would throw an exception:
> \Drupal::$container is not initialized yet. \Drupal::setContainer() must be called with a real container.

Test this for yourself by simply setting `$site_path = null` prior to that line. Fortunately as far as I know it's not possible for `$site_path` to be undefined so this logic is never called.

Even Drupal's [default.settings.php](https://git.drupalcode.org/project/drupal/blob/8.8.x/sites/default/default.settings.php#L675) uses `$site_path` without making any such check.

If there is some kind of edge case were are trying to support here, we need clear testing steps for it, a comment in the code, and some guard code such as `if (\Drupal::hasService('kernel'))...`

Hold to verify on ACSF.

### DIFF
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -64,7 +64,9 @@ if ($ip) {
 }
 
 $repo_root = dirname(DRUPAL_ROOT);
-/** @var $site_path **/
+if (!isset($site_path)) {
+  $site_path = 'default';
+}
 $site_dir = str_replace('sites/', '', $site_path);
 
 // Special site name detection for ACSF sites being developed locally.
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -64,7 +64,10 @@ if ($ip) {
 }
 
 $repo_root = dirname(DRUPAL_ROOT);
-/** @var $site_path **/
+/**
+ * @var $site_path
+ * This is always set and exposed by the Drupal Kernel.
+ */
 $site_dir = str_replace('sites/', '', $site_path);
 
 // Special site name detection for ACSF sites being developed locally.
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -64,10 +64,7 @@ if ($ip) {
 }
 
 $repo_root = dirname(DRUPAL_ROOT);
-
-if (!isset($site_path)) {
-  $site_path = \Drupal::service('site.path');
-}
+/** @var $site_path **/
 $site_dir = str_replace('sites/', '', $site_path);
 
 // Special site name detection for ACSF sites being developed locally.
--- a/settings/blt.settings.php
+++ b/settings/blt.settings.php
@@ -64,9 +64,7 @@ if ($ip) {
 }
 
 $repo_root = dirname(DRUPAL_ROOT);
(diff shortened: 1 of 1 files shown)


settings/blt.settings.php
```diff
@@ -64,10 +64,7 @@
 }
 
 $repo_root = dirname(DRUPAL_ROOT);
-
-if (!isset($site_path)) {
-  $site_path = \Drupal::service('site.path');
-}
+/** @var $site_path **/
```
alexxed: Well it's odd to use a variable that is not defined in this file. Where does it come from ? Should we not check for it?

greylabel: Drupal provides in `DrupalKernel::initializeSettings` -- https://git.drupalcode.org/project/drupal/blob/8.8.x/core/lib/Drupal/Core/DrupalKernel.php#L1081

danepowell: Yeah I agree it's weird. I know core defines `$site_path` as Grant pointed out, but I don't understand how it becomes exposed to the global namespace in settings files. But clearly somehow it does, or we'd have hundreds of ACSF customers complaining about this code causing a WSOD.

danepowell: Just to be super safe, I changed it to check if `$site_path` is defined first. Although it _must_ be defined, because even core's default.settings.php uses `$site_path` without checking if it's set first.

greylabel: The call stack looks roughly like:

`Index.php -> Drupal\Core\DrupalKernel::handle -> Drupal\Core\DrupalKernel::initializeSettings -> Drupal\Core\Site\Settings::initialize -> [require()](https://git.drupalcode.org/project/drupal/blob/8.8.x/core/lib/Drupal/Core/Site/Settings.php#L125)`

`$site_path` is in scope at when `settings.php` is called.

settings/blt.settings.php
```diff
@@ -64,9 +64,8 @@
 }
 
 $repo_root = dirname(DRUPAL_ROOT);
-
 if (!isset($site_path)) {
-  $site_path = \Drupal::service('site.path');
+  $site_path = 'default';
```
greylabel: This is probably unnecessary if settings.php files are included correctly by Drupal and should probably fail if not rather than assuming the `$site_path` is `default` without additional checks or context.

alexxed: Yeah, I'm happy if I see a throw error here, just to be sure this file was included in a way that site_path is set.

settings/blt.settings.php
danepowell: Makes sense, thanks for the explanation!

settings/blt.settings.php
danepowell: I just reverted the most recent commit, so there's no more check or default value. I think it's perfectly safe to assume this is set by core, as even core itself does this. I don't see how throwing an additional error would add value.

PR #3667, opened 2019-05-20
Fixes #3665: Drupal Spec Tool dependency version.

### LINKED ISSUE #3665: Loosen relationship with Drupal Spec Tool
It's not you, it's me. I just need a little space. 😉

I'd like to move `acquia/drupal-spec-tool` from requirements to suggestions because it's not a strict requirement--it's an optional nice-to-have. I'd also like to loosen the version constraint by at least making it "greater than or equal to". (The current constraint will require a coordinated release for the next version, which is no good for anyone.) These changes should have no impact on pre-existing sites, of course--only new projects. What do you think, @danepowell?

danepowell:
Fixes #3665
--------

Changes proposed
---------
- Remove Drupal Spec Tool as a hard dependency of BLT
- Change version constraint in project template to be less strict

This essentially makes Drupal Spec Tool a "suggested" dependency, since any BLT project can just remove it from their composer.json if they don't want it, same as all of the other modules.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "grasmash/yaml-expander": "^1.2.0",
         "oomphinc/composer-installers-extender": "^1.1",
         "sensiolabs/security-checker": "^5.0.0",
-        "symfony/config": "^4.2",
+        "symfony/config": "^3.0",
         "symfony/console": "^3.4.0",
         "symfony/twig-bridge": "^3.3",
         "symfony/yaml": "^3.2.8",
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,6 @@
         "php": ">=7.1",
         "ext-json": "*",
         "composer-plugin-api": "^1.1.0",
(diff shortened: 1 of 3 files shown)


danepowell:
@TravisCarden do you have any idea what's causing the test error? It seems related to ORCA, but I can't fathom why this PR would trigger such an error. I also haven't seen that error in any other recent test runs. Restarting didn't clear it.

danepowell:
Weeeird. Bumping symfony config from 4 to 3 fixed it. I _though_ we were already using 4 before anyway. I guess maybe this conflicted with a dependency of some other Acquia product. ORCA doing its job I suppose :)

danepowell:
Oh... the error related to Symfony Config was probably triggered by the fact that BLT Project requires Behat <3.4, and Behat <3.4.2 requires Symfony Config <4. So when ORCA tried to require BLT via Composer, Composer grabbed some other version of BLT (not the one from this PR) that would satisfy those constraints.

PR #3676, opened 2019-05-22
Fixes #3674 to repair site splits in 10.x.

### LINKED ISSUE #3674: Site Splits Broken
**Describe the bug**
This is an issue impacting both BLT 9.2.x and 10.x (and perhaps previous versions as well).

Currently, config splits based on a site (or site splits) should be functional when a config split has the exact same name as a site on ACSF. 

After https://github.com/acquia/blt/issues/3362 got fixed (which resolved a site name detection issue) I had hoped we were good, but after extensive troubleshooting, this is still not working. 

**To Reproduce**
1. Create a "site split" that properly matches the site name of at least one acsf site
2. Deploy code to acsf
3. Confirm that the site split didn't get activated
4. Using `drush php` print out the global $config array
5. see output:

```
global $config;
>>> print_R($config);
Array
(
    [config_split.config_split.local] => Array
        (
            [status] => 
        )

    [config_split.config_split.dev] => Array
        (
            [status] => 
        )

    [config_split.config_split.stage] => Array
        (
            [status] => 
        )

    [config_split.config_split.prod] => Array
        (
            [status] => 
        )

    [config_split.config_split.ci] => Array
        (
            [status] => 
        )

    [config_split.config_split.ah_other] => Array
        (
            [status] => 1
        )

    [config_split.config_split.g/files/<directory>] => Array
        (
            [status] => 1
        )
```

**Expected behavior**
I would expect an additional item in the config array above that corresponds with the site split. I could understand if the split wasn't getting activated for "some reason" but the fact that it's not in the config object at all is highly suspect given https://github.com/acquia/blt/blob/10.x/settings/config.settings.php#L89 

**Detailed error output**
N/A

**BLT doctor output**
N/A

**System information**
* BLT version: 9.2.3 (but persists across all 9.2.x and 10.x versions)

**Additional context**
This issue fits into a broader issue with config splits and the current landscape. While not directly related, https://github.com/acquia/blt/issues/3575 is a similar set of broken config split detection (but for profiles).

I think environment splits have been fine all along (because BLT can detect that). Site splits should _also_ be fine, once we merge these PRs, as BLT can easily detect the site on ACSF (thanks to the post-sites factory hook). However, as mentioned, profile splits require bootstrapping Drupal (so are a slightly different ball of wax).

**Solution**
We tested this by adding a global $_acsf_site_name, re-deployed, and confirmed that the site splits were actually working.

mikemadison13:
Fixes #3674 
--------

Changes proposed
---------

- added a global prior to checking for the existence of the site name variable

### DIFF
--- a/settings/config.settings.php
+++ b/settings/config.settings.php
@@ -87,6 +87,7 @@ if ($split != 'none') {
 $config["$split_filename_prefix.$site_dir"]['status'] = TRUE;
 
 // Set acsf site split if explicit global exists.
+global $_acsf_site_name;
 if (isset($_acsf_site_name)) {
   $config["$split_filename_prefix.$_acsf_site_name"]['status'] = TRUE;
 }


danepowell:
I'm having trouble testing this because the stock db-update.sh factory hook (before even applying this patch) is returning a pretty cryptic error: 
```
Could not retrieve the site standard_domain from the database.
``
Not sure if you've seen that before. This is an entirely vanilla BLT 10 project.

mikemadison13:
I haven't but haven't tried BLT on ACSF either so...

danepowell:
Found the source of the error, not related to this PR, updated docs to account for it: #3680

I was able to verify that the original bug exists as described, and that this PR fixes it. Thanks!

At some point we should probably refactor this whole system to avoid slinging around so many global variables, but this will do for now.

PR #3678, opened 2019-05-24
Removed docs on updating from non-Composer versions.

danepowell:
This is in response to an internal PUB ticket pointing out that these docs are outdated.

BLT has used Composer for going on 4 years now, so it's extremely unlikely that these docs are still relevant to anyone. Let's get rid of them and avoid confusion and the additional maintenance burden.


danepowell:
@alexxed please review when you can

danepowell:
Since this is only a docs change and we need to work with the docs team on this, I'm going to go ahead and merge.

PR #3679, opened 2019-05-24
pipelines-deploy should fail gracefully.

danepowell:
The `pipelines-deploy` command (own by Cloud / Pipelines, not BLT) can fail for a variety of benign reasons, for instance due to transient issues with AWS (500 / 504 errors). It will also fail for any customer using Pipelines but not CDEs, or when they run out of CDEs (403 errors).

In these situations, the entire Pipelines job is marked as failure, and that's a bad customer experience as it renders the automated tests basically worthless (especially if every job starts failing).

Since spinning up a CDE is a "nice to have" but not nearly as important as the outcome of Pipelines automated tests, which are rendered useless by these failures, I think we should just swallow any such errors.

### DIFF
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -45,14 +45,14 @@ events:
     steps:
       - deploy:
           script:
-            - pipelines-deploy
+            - source ${BLT_DIR}/scripts/pipelines/pipelines_deploy_safe
   pr-merged:
     steps:
       - deploy:
           script:
-            - pipelines-deploy
+            - source ${BLT_DIR}/scripts/pipelines/pipelines_deploy_safe
   pr-closed:
     steps:
       - deploy:
           script:
-            - pipelines-deploy
+            - source ${BLT_DIR}/scripts/pipelines/pipelines_deploy_safe
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -50,9 +50,11 @@ events:
     steps:
       - deploy:
           script:
+            - composer install --ansi
             - source ${BLT_DIR}/scripts/pipelines/pipelines_deploy_safe
   pr-closed:
     steps:
       - deploy:
           script:
+            - composer install --ansi
             - source ${BLT_DIR}/scripts/pipelines/pipelines_deploy_safe
--- a/scripts/pipelines/pipelines_deploy_safe
+++ b/scripts/pipelines/pipelines_deploy_safe
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -v
+set +e
+
+pipelines-deploy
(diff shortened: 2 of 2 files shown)


danepowell:
@mikemadison13 I asked you about this the other day, this is the best fix I could think of for ACSF customers running Pipelines.

danepowell:
@raghunat the problem with that approach is that a single BLT-based project can be deployed to multiple Cloud subscriptions. For instance, many of our largest customers use ACE for testing and development (with CD) and ACSF for production. So the same pipelines file needs to apply to first, i.e. Pipelines _has_ to either detect dynamically whether CD is present or fail gracefully.

How would you propose we solve this?

danepowell:
@alexxed @adrianthedev could use your review, as well as your thoughts on Stephen's comment above

raghunat:
@danepowell Thanks for the context! I acknowledge that there are BLT users out there who use ACE  as a transition platform to ACE, but I have generally seen that pipelines is not on the ACSF side but the ACE side only in those situations (I could be off base here).  That being said, we want to help our ACSF folks to CICD, so knowing which applications are ACSF and ACE is something we could make pipelines aware of, and expose accordingly so that a yaml could be crafted to either run or not run those steps. With Cloud API v2, we could perhaps discover which are ACE or ACSF in the interim?

danepowell:
A Pipelines container is pretty isolated, I don't _think_ it would generally have access to the API.

I think the best way to do this would be with environment variables. I'm pretty sure on ACSF PIPELINE_CLOUD_REALM is generally `enterprise-g1`. We can disable CD deployments based on that.

Keep in mind that this PR isn't just about ACSF vs ACE though. There's several other transient reasons that CD environment might fail to start (capacity, maintenance, AWS issues) and I'm not sure they all warrant failing the entire job. I still think there'd be value in failing gracefully.

alexxed:
Let me see if I get this right, so if pipeline-deploy does this:

- a (exit 0)
- b (exit 1)
- c (exit 0)

with this patch, the deploy job in pipeline will not be marked as failed right?

danepowell:
I'm not sure what you mean by that. Are you talking about the fact that pipelines-deploy is run in three places (post-deploy, pr-merged, and pr-closed)?

Each of those events corresponds to a separate Pipelines job. For instance, when you open a PR it runs the post-deploy event, which calls pipelines-deploy to create a CDE. Then when you merge _or_ close a PR, it runs the pr-merged or pr-closed event, which calls pipelines-deploy to tear down the CDE. But each of these runs in a separate Pipelines build job.

Regardless, if `pipelines-deploy` fails at any point the job will see show as passing, as long as all of the prior tests passed.

danepowell:
Closing per discussions with Stephen. Unfortunately there's no general way for BLT to work around this right now. The closest we could come is #3695, but that doesn't generalize because there are other ACSF realms besides enterprise-g1.

PR #3680, opened 2019-05-24
Updated ACSF documentation.

danepowell:
I discovered a fun little roadbump in setting up ACSF sites while testing #3676, thought it would be helpful to document. I also thought we should remove the now-obsolete info on setting up profiles with ACSF, since that will soon be supplanted by Mike's KB article.

### DIFF
--- a/docs/acsf-setup.md
+++ b/docs/acsf-setup.md
@@ -3,15 +3,17 @@
 To configure a project to run on ACSF, perform the following steps after initially setting up BLT, but _before_ creating any sites in ACSF:
 
 1. Execute `blt recipes:acsf:init:all` from the project root and commit any changes.
-1. Create one or more custom profiles:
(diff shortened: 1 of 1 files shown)


mikemadison13:
a couple of thoughts

* we should likely mention that creating more than one profile can have an impact on configuration management strategies (and can link to my KB article once it's done, that shouldn't hold up this PR though).
* we should warn them that acsf config and modules must be present in configuration splits (as requiring it in the profile is great for installation, but not sufficient for updates)
* not sure what the preferred drush alias strategy is at this point given acsf tools, but we should probably mention that as well.

i'm not sure that drush and config need to be extensively documented here, but there are some acsf specific things that should likely be called out.

danepowell:
Thanks, I added notes to that effect.

PR #3682, opened 2019-05-28
Prevent patched htaccess from being overwritten for ACSF customers.

danepowell:
Changes proposed
---------
- When customers run `blt acsf:init`, add a key to `composer.json` that will prevent drupal-scaffold from overwriting .htaccess on future updates.

Steps to replicate the issue
----------
1. Initialize ACSF on a project and commit changes: `blt acsf:init`
2. Make a change that triggers Drupal scaffold, such as updating or reverting Drupal core versions

----------
Notice that the ACSF-specific patch to .htaccess is lost / overwritten as part of the update.

-----------
.htaccess customizations are preserved as part of updates.


mikemadison13:
i'm wondering if this should be a more broad thing... not just acsf. i use drupal scaffold for ace sites to do this as well, just to ensure that customizations to the htaccess file aren't overwritten. should blt just "do" this by default? should we have a command for it that can be called outside of acsf init if not?

adrianthedev:
That is a good point. Should blt do that by default? That would mean having the update in a different command, not just under `acsf:init`.

mikemadison13:
I was thinking we could have a command for this, an acsf:init could then call that command. Obviously not "all" sites will need to do this, but anyone who forces HTTPS via htaccess, or perhaps has custom redirect rules, etc. will definitely need it.

greylabel:
@mikemadison13 is there a particular advantage vs just pathing with `"scripts": {"post-drupal-scaffold-cmd": [],}` in `composer.json`?

mikemadison13:
honestly, no probably not. hadn't considered that :)

danepowell:
It seems like overkill to create a command that just adds a single line to your composer.json file, which is a trivial change to make otherwise. If you are advanced enough as a developer to know "hey I should run this command because I made a change to .htaccess", you are also competent enough to just make the change yourself.

I feel like BLT should only manage this in the context of stopping you from shooting yourself in the foot, i.e. as part of the ACSF setup command where we constantly see customers clobbering the htaccess changes and generating support tickets.

A documentation fix might be better. Ideally I think I'd add a comment to composer.json explaining how to do this, but json doesn't support comments so :man_shrugging: I feel like if we put this anywhere else, people wouldn't find it any more easily than they would by RTFM for Drupal Scaffold.

src/Robo/Commands/Acsf/AcsfCommand.php
```diff
@@ -71,6 +71,12 @@ public function acsfInitialize($options = ['acsf-version' => '^2.47.0']) {
       $project_config['modules']['local']['uninstall'][] = 'acsf';
     }
     YamlMunge::writeFile($project_yml, $project_config);
+
+    // Exclude .htaccess from further updates due to ACSF patch.
```
alexxed: Is there a reference to more text? I can't tell what ACSF patch this is referring to.

alexxed: Oh, now from the comments I get it. It's about ACSF patch of .htaccess. Can we just say in the comment `// .htaccess was patched, excluding from further updates` if that is the case?

PR #3692, opened 2019-05-30
DT-485: Don't optimize Composer autoloader locally.

### LINKED ISSUE #3591: Optimize autoloader for deploys, but don't optimize locally
Matt added Composer config and CLI arguments in #3312 to optimize the autoloader in an attempt to fix #3148. We now know that the root cause of #3148 was unrelated, and this change had no effect at best, and may be disrupting local development at worst.

According to the [official Composer docs](https://getcomposer.org/doc/articles/autoloader-optimization.md#autoloader-optimization), you should never optimize the autoloader in development environments. We should revert this change.

Not do this? I don't know that I've seen any clear problems either way, but I'd prefer to stay on the reservation and do what Composer officially recommends unless there's a clear reason not to. It seems like we got into this situation by essentially spitballing solutions, which I'd like to undo.

Even though this has nothing to do with #3148, since it does touch the autoloader we should make sure that once removed the class errors don't come pack when using the upstream composer patch: https://github.com/composer/composer/pull/8108

danepowell:
Fixes #3591 
--------

Changes proposed
---------
- Don't optimize the autoloader locally since the Composer documentation expressly forbids doing so

To test, run `blt setup` locally and ensure that it doesn't behave any differently than before.

I also verified that this in no way interacts with https://github.com/composer/composer/pull/8085

### DIFF
--- a/src/Robo/Commands/Setup/BuildCommand.php
+++ b/src/Robo/Commands/Setup/BuildCommand.php
@@ -102,7 +102,7 @@ class BuildCommand extends BltTasks {
   public function composerInstall() {
     $result = $this->taskExec(
       (DIRECTORY_SEPARATOR == "\\") ? 'set' : 'export' .
-        " COMPOSER_EXIT_ON_PATCH_FAILURE=1 && composer install --ansi --no-interaction --optimize-autoloader --apcu-autoloader"
+        " COMPOSER_EXIT_ON_PATCH_FAILURE=1 && composer install --ansi --no-interaction"
     )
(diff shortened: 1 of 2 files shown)


dpagini:
Hey @danepowell - do you think there should be some kind of update script for this? Or maybe a call-out in the release notes? Or else this fix will only affect new projects, or users who look into each change... right?

danepowell:
Yeah. We didn't write an update hook originally because it seemed like a pretty trivial change, but I guess we should just too keep projects consistent: #3721

PR #3694, opened 2019-05-31
DT-482: Fix deprecated SUT tests

### LINKED ISSUE #3638: Task: review allowed test failures
Now that we've integrated with ORCA, we should review the allowed failures (such as deprecation tests) to see if there's anything we need to address.

danepowell:
Fixes #3638 
--------

Changes proposed
---------
- Add robofile to Composer autoloader so phpstan can find it.
- Move all tests into a common directory so they can be autoloaded properly

Steps to replicate the issue
----------
1. View automated tests

----------
deprecated SUT tests failing

-----------
deprecated SUT tests pas


danepowell:
Unfortunately I don't think we'll be able to resolve all deprecation warnings because of https://github.com/doctrine/annotations/issues/270

danepowell:
Per @TravisCarden it may also not be easy/possible to fix the fact that phpstan is scanning files that intentionally aren't being autoloaded, because they are template / example files. I guess we need to decide if that's acceptable or not.

PR #3695, opened 2019-05-31
Don't run pipelines-deploy on ACSF.

danepowell:
Changes proposed
---------
- When Pipelines is run in an ACSF subscription, don't run `pipelines-deploy` because CD environments cannot exist on ACSF, and this command will always fail.


danepowell:
This doesn't generalize because there are other ACSF realms besides enterprise-g1.

PR #3701, opened 2019-06-05
Fixes #3148: Frivolous composer notices.

### LINKED ISSUE #3148: Resolve "Could not scan for classes inside" notices

danepowell:
Fixes #3148 
--------

Changes proposed
---------
- Add a post-autoload-dump script to be invoked by the project template, which will re-sort Composer's install.json file to ensure that composer/installers gets loaded prior to blt on subsequent composer operations.

Testing steps
----------
1. Starting from a normal BLT-based project, run any composer command (`composer --version`) and get all of the notices about "could not scan for classes...".
2. Apply this patch via Composer: https://patch-diff.githubusercontent.com/raw/acquia/blt/pull/3701.diff (note you MUST use the .diff format, not .patch like normal)
3. Run any composer command (`composer --version`) and verify that the notices are no longer generated. 

Additional details
-----------
I would consider this a workaround until/if this is fixed upstream: https://github.com/composer/composer/issues/8065


christopher-hopper:
Is there some way I can send @danepowell a hug?
This was such a clever find. I think I am starting to cry a little. It's made me all emotional.

Nice work!

yyss8:
I start seeing those warnings after updating composer from `1.8.6` to `1.9.0` https://github.com/composer/composer/releases/tag/1.9.0

image

danepowell:
Uggggh. Okay, thanks for reporting, I'll investigate this downstream in BLT and try to confirm whether it's a BLT issue, Composer issue, or both: https://github.com/acquia/blt/issues/3807

ndobromirov:
Just an update related to `BLT 9.2.8`
For me it was a composer issue. 
After updating `composer` from `1.8.6` to `1.9.1` and reinstalling the vendor with 
```
rm -rf ./vendor && composer install
```
The problematic notices went away.

PR #3705, opened 2019-06-10
Work around Drush 9.5 bug with path based domains

rmuit:
Fixes # (Acquia internally reported issues, possibly)
--------

Changes proposed
---------
Add slash to ACSF db-update.sh hook so that path domains should be able to work with Drush 9.3 / 4 / 5 (I believe Drush 9.2 and older will never work. Not sure anymore. Drush8 works fine.)

Steps to replicate the issue
----------
1. Have codebase with Drush 9.5; have ACSF site with path based domain
2. Update code.

----------
Likely: error related to database "default" not being found?

-----------
Normal update.

Additional details
-----------
I filled in the template but actually details were discussed with @danepowell in internal tickets.

### DIFF
--- a/scripts/factory-hooks/db-update/db-update.sh
+++ b/scripts/factory-hooks/db-update/db-update.sh
@@ -18,7 +18,8 @@ site="$1"
 env="$2"
 # database role. (Not expected to be needed in most hook scripts.)
 db_role="$3"
-# The public domain name of the website.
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks. I reproduced the original issue using Drush 9.5, and also confirmed that it doesn't exist in 9.7. I then confirmed that this PR fixes Drush 9.5 and doesn't have any negative impact on 9.7.

Issue #3704, opened 2019-06-10
All chromedriver tests fail with 400 Bad Request error


bobbygryzynger:
**I want to...**
Address `FunctionalJavascript` (chromedriver) test failures.

This is likely an upstream issue, but perhaps a workaround can be investigated for BLT users pending an upstream fix.

**It's not working because...**
All tests which require a user to log in fail at the login step with: `The requested URL returned error: 400 Bad Request`. The same thing is occurring within BLT's internal tests. See [here](https://travis-ci.com/acquia/blt/jobs/206614349#L2503) for instance.

**System information**
* Operating system type: `Linux`
* Operating system version: `Ubuntu 14.04.5`
* BLT version: `10.x`

bobbygryzynger:
It looks like the latest release of [`acquia/orca`](https://github.com/acquia/orca/releases/tag/v1.0.0-alpha14) will fix the internal build failures, but that still leaves open the question of how to address the build failures in BLT projects. 

It seems like one viable solution is to add [`lbaey/chromedriver`](https://github.com/lbaey/chromedriver) as a dependency (which `acquia/ocra` does) to allow projects to pin their version of `chromedriver` if necessary. 

This would then make the [`chromedriver` download script](https://github.com/acquia/blt/blob/10.1.0/scripts/linux/install-chrome.sh) obsolete.

danepowell:
This is due to a breaking change in Chrome 75, not anything that BLT is responsible for. There may be multiple problematic upstream libraries, at least one of which might be Drupal Extension: https://github.com/jhedstrom/drupalextension/issues/535

bobbygryzynger:
It’s true that this is an upstream issue. But, the one thing that BLT is responsible for is downloading the latest version of `chromedriver` and, currently, if projects use BLT’s setup scripts, there’s no way to use anything other than the latest version of `chromedriver`.

danepowell:
I think the only situation in which BLT downloads Chromedriver for you is if you are using TravisCI. I've opened a PR to address that: #3708 

Otherwise you need to provide Chromedriver yourself in whatever environment is running BLT, and I would definitely recommend pinning to version 74 until this upstream issue is resolved:
https://github.com/minkphp/MinkSelenium2Driver/issues/293

PR #3706, opened 2019-06-10
Cleaned up test matrix.

danepowell:
I found the way we'd defined the test matrix to be really hard to reason through, and predict what's going to run. It also contains a lot of cruft, plus now Drupal tests are completely broken due to #3704. So here are a few changes:

- Remove all references to the 'long' group since we don't define any tests in that group
- Remove all references to the 'requires-vm' group since that's already excluded by default in phpunit.xml, except when we're resetting PHPUNIT_EXCLUDE_GROUP
- Allow Drupal tests to fail because there's no way we can fix them for now
- Change the Drupal core HEAD version of 8.8.x and allow it to fail since that's very bleeding-edge and not supported by BLT 10.


raghunat:
Approved given the fact that:

We didn't really test HEAD before, but now will, even if it fails and that this will help mainline development continue as we work with our upstream dependencies

PR #3711, opened 2019-06-13
Cleaned up TravisCI cruft

danepowell:
We had a _ton_ of junk in .travis.yml and related files that is no longer needed or relevant. This cleans it up.


.travis.yml
```diff
@@ -7,19 +7,13 @@ addons:
   chrome: stable
 
 cache:
-  bundler: true
```
danepowell: We don't use Ruby so we don't need to cache Bundler. Presumably this is a holdover from when Compass was all the rage.

.travis.yml
```diff
@@ -7,19 +7,13 @@ addons:
...
-  apt: true
```
danepowell: This is not even a valid key in TravisCI

.travis.yml
```diff
@@ -7,19 +7,13 @@ addons:
...
   directories:
     - "$HOME/.composer/cache"
-    - "$HOME/.console"
     - "$HOME/.drush/cache"
-    - "$HOME/.nvm"
-    - "$TMPDIR/blt-sandbox-instance/.rules"
```
danepowell: We don't use Drupal Console, NVM, or long phpunit tests any more. The history on these are here:
- https://github.com/acquia/blt/commit/d48c277d7479c5e266d821d5010471d50d7e4994
- https://github.com/acquia/blt/commit/71cdb5ea7e987a313a0f208dabf52ca7b1a514c5
- https://github.com/acquia/blt/pull/2545

scripts/travis/.travis.yml
```diff
@@ -1,57 +1,41 @@
-sudo: true
```
danepowell: Not necessary, containerized builds now have sudo anyway

scripts/travis/.travis.yml
```diff
@@ -1,57 +1,41 @@
...
 language: php
 dist: xenial
 
 # Adjust the version of PHP to match your production environment.
 # Making this version number greater than the production environment can have unintended consequences
 # including a non-functional prod environment.
-php:
-  - 7.2
-
-matrix:
```
danepowell: Not relevant since we don't have any build matrix or allow_failures jobs

scripts/travis/.travis.yml
```diff
@@ -1,57 +1,41 @@
...
-  fast_finish: true
+php: "7.2"
 
 env:
   global:
     - COMPOSER_BIN=$TRAVIS_BUILD_DIR/vendor/bin
     - BLT_DIR=$TRAVIS_BUILD_DIR/vendor/acquia/blt
-    - BUILD_DIR=$TRAVIS_BUILD_DIR
-
-jdk:
```
danepowell: Not necessary since BLT now uses the dmore behat chrome extension (direct Chrome debugging) rather than Selenium.

scripts/travis/.travis.yml
```diff
@@ -1,57 +1,41 @@
...
-# - slack: '<account>:<token>#[channel]'
+# notifications:
+#   - hipchat: [api token]@[room id or name]
+#   - slack: '<account>:<token>#[channel]'
 
 before_install:
   # Disable xdebug.
   - phpenv config-rm xdebug.ini
   - composer self-update
-  - composer validate --no-check-all --ansi
```
danepowell: Not necessary since `blt validate` now does this

.travis.yml
```diff
@@ -7,19 +7,13 @@ addons:
...
     - "$TMPDIR/phpstan/cache"
 
 services:
   - mysql
-  - xvfb
```
danepowell: Not necessary without Selenium

danepowell:
> Looks good, but I'd remove all commented statements like notifications, ssh known hosts, etc.

@alexxed can you suggest a better way to guide people on how to add those things? Maybe do it as part of the Travis CI init script? The problem is that forgetting to add ssh known hosts is one of the top reasons db syncs and deploys fail from Travis, we really have to shove it in people's faces for them to notice.

Issue #3715, opened 2019-06-17
Add isset check on $_SERVER['PWD'] superglobal in find_repo_root()


brooke-heaton:
**Describe the bug**
On local environments that do not explicitly declare a $_SERVER['PWD'] superglobal variable (I'm not sure how this gets declared - does this original with DevDesktop?), the blt:update (or even the blt --version) method will fail in `find_repo_root()` (vendor/acquia/blt/bin/blt-robo.php:23). This was previously noted and resolved in https://github.com/acquia/blt/issues/3425. I'm not sure why this regressed or why this was overwritten. 

**To Reproduce**
Steps to reproduce the behavior, ideally starting from a fresh install of BLT:
1. Running a local Lando environment with BLT 10.0.0-beta2
2. Run `lando blt --version`, `lando blt blt:update` or nearly ANY blt command that runs `find_repo_root()` 
3. See error

**Expected behavior**
BLT should work in environments that don't explicitly declare a $_SERVER['PWD'] superglobal variable. This variable appears to be fairly local setups specific though I'm not clear on where this gets declared. In short, it is overly assuming. 

**Detailed error output**
PHP Notice:  Undefined index: PWD in /app/vendor/acquia/blt/bin/blt-robo.php on line 23

Notice: Undefined index: PWD in /app/vendor/acquia/blt/bin/blt-robo.php on line 23

**BLT doctor output**
PHP Notice:  Undefined index: PWD in /app/vendor/acquia/blt/bin/blt-robo.php on line 23

Notice: Undefined index: PWD in /app/vendor/acquia/blt/bin/blt-robo.php on line 23
+---------------------------+--------------------------------------------------+
| Property                  | Value                                            |
+---------------------------+--------------------------------------------------+
| %paths.%root              | /app/docroot                                     |
| %paths.%site              | sites/default                                    |
| %paths.%modules           | sites/all/modules                                |
| %paths.%themes            | sites/all/themes                                 |
| %paths.%config-sync       | /app/config/default                              |
| alias-searchpaths.0       | /app/drush/sites                                 |
| blt-version               | 10.0.0-beta2                                     |
| composer-version          | Composer version 1.8.4 2019-02-11 10:52:10       |
| config-sync               | /app/config/default                              |
| db-driver                 | mysql                                            |
| db-hostname               | localhost                                        |
| db-name                   | drupal                                           |
| db-password               | drupal                                           |
| db-port                   | 3306                                             |
| db-username               | drupal                                           |
[...]

PR #3717, opened 2019-06-17
Add isset check on $_SERVER['PWD'] in find_repo_root()

brooke-heaton:
This addresses https://github.com/acquia/blt/issues/3715

Not all local environments declare a $_SERVER['PWD'] superglobal and this results in a PHP Notice that wrecks numerous BLT commands. This fails, for instance, in Lando environments. This should be made agnostic so that $_SERVER['PWD'] is only included as a possible_repo_root if declared.

Fixes # .

Changes proposed:

Remove $_SERVER['PWD'] from $possible_repo_roots
Add an isset check to $_SERVER['PWD']
if set, add $_SERVER['PWD'] to $possible_repo_roots via array_unshift to maintain the original array order.

### DIFF
--- a/bin/blt-robo.php
+++ b/bin/blt-robo.php
@@ -20,11 +20,14 @@ require_once __DIR__ . '/blt-robo-run.php';
  */
 function find_repo_root() {
   $possible_repo_roots = [
-    $_SERVER['PWD'],
     getcwd(),
     realpath(__DIR__ . '/../'),
     realpath(__DIR__ . '/../../../'),
   ];
+      // Check for PWD - some local environments will not have this key.
+    if (isset($_SERVER['PWD'])) {
+        array_unshift($possible_repo_roots, $_SERVER['PWD']);
+    }
   foreach ($possible_repo_roots as $possible_repo_root) {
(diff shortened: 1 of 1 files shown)


danepowell:
@brooke-heaton I think you must be using an outdated version of BLT 10.x. Notice the merge conflict on this PR? That's because this was already fixed back in March, and I can't see that it's changed at all since then:
https://github.com/acquia/blt/blob/aeedc1f5d57cf223361defd458007155140414cb/bin/blt-robo.php#L27

brooke-heaton:
@danepowell yikes. my bad! I just upgraded to 10 so I must be out of date. My bad and thanks!

brooke-heaton:
@danepowell likely due to me following the upgrade instructions pretty religiously. I had used this command `composer require acquia/blt:^10.0.0 --no-update && composer update` and didn't end up on the latest version. 

https://github.com/acquia/blt/wiki/What's-new-in-BLT-10.x#how-to-upgrade

danepowell:
`composer require acquia/blt:^10.0.0 --no-update && composer update` should definitely get you the latest version (10.1.0). Maybe you had some other composer dependency that conflicted with later BLT versions?

If you're able to reproduce a problem there feel free to open a new issue.

brooke-heaton:
Nah, let's close this. My bad. I'll figure it out locally.

PR #3720, opened 2019-06-18
Fixes #3639: DT-529: Use Acquia coding standards

### LINKED ISSUE #3639: Task: share code standards with ORCA
Right now BLT provides its own phpcs.xml.dist. Now that we are using ORCA, we should share our code standards with other Acquia products. Travis is working on sharing those standards so we can incorporate them. Will hopefully just need to modify phpcs.xml.dist to inherit from ORCA and customize as necessary to exclude some of the parts of our codebase that are special.

danepowell:
Fixes #3639 
--------

Changes proposed
---------
- Use Acquia coding standards
- Clean up code to comply with standards

This should all be strictly non-functional changes.


scripts/drupal-vm/post-provision.php
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env php
 <?php
-
+// phpcs:ignoreFile
```
danepowell: Has to be ignored for now because PHPCS gets really confused when a shell file uses PHP

PR #3721, opened 2019-06-19
DT-485: Update hook for composer autoload change.

danepowell:
Follow-up to #3692
--------

Changes proposed
---------
- Automatically update existing users' composer.json to incorporate change from #3692
- Add a newline when writing composer.json files (they have a newline by default, and all files should have a newline per code standards)

Steps to replicate the issue
----------
1. Start with a project generated by BLT version 10.0 or earlier, which has the old config keys.
2. Run `blt blt:update` and see that the keys are removed.

Requires a change record because people might not understand why these keys are getting removed and get alarmed that we are removing "optimizations".

### DIFF
--- a/src/Update/Updater.php
+++ b/src/Update/Updater.php
@@ -444,7 +444,7 @@ class Updater {
       ksort($contents['require-dev']);
       $contents['require-dev'] = (object) $contents['require-dev'];
     }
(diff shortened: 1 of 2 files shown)


danepowell:
@dpagini please review when you get a chance

PR #3724, opened 2019-06-19
DT-536: Fix ACSF stack detection.

danepowell:
ACSF stacks can appear as 02dev, 03dev, etc...


SirHawbly:
Solved issues with private file paths in *.02dev instances z825825.

PR #3725, opened 2019-06-24
Update INSTALL.md: don't recommend installing drush with brew

erinrasmussen:
removing drush from line 33  `brew install php71 git composer drush` 
since the current line generates the following error: 
Error: No available formula with the name "drush"

This makes some sense, the best practice is to install drush as a dependency of your drupal project with Composer.

Fixes # 
--------
the current brew install command generates the following error: Error: No available formula with the name "drush"
Changes proposed
---------

-
removing drush from line 33  `brew install php71 git composer drush` 
-

Steps to replicate the issue
----------
1.  
2. 

----------
brew generates the error Error: No available formula with the name "drush"

-----------

brew installs php71 git nd composer 

Additional details
-----------


danepowell:
Thanks!

PR #3712, opened 2019-06-13
Update ORCA to alpha18.

danepowell:
https://github.com/acquia/orca/releases/tag/v1.0.0-alpha17

This is awesome because we should get full deprecation testing now.

Blocked by:
- [x] #3627 to fix code style errors 
- [x] #3738 to fix deprecation tests
- [x] Figure out what to do about deprecated Behat context: https://github.com/jhedstrom/drupalextension/issues/537
- [x] https://github.com/acquia/orca/issues/27 to fix other deprecation tests
- [x] alpha17 release of ORCA to fix overzealous code style scanning (i.e. MD files)
- [x] Update coding-standards dependency version to match ORCA dependency in ~alpha17~ whatever version includes the library
- [x] Think about removing native PHPCS checks from BLT, since they would now duplicate ORCA.
- [x] Remove version constraint `mglaman/phpstan-drupal": "0.11.7"` introduced in #3739
- [x] Update to Symfony 4.x if we can to work around a deprecation scanning bug with ORCA

PR #3726, opened 2019-06-25
Use AcquiaDrupalStrict coding standards

danepowell:
acquia/coding-standards-php recently split its coding standards. We should go with the stricter version in order to unblock #3712.

### DIFF
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -460,7 +460,7 @@ class RoboFile extends Tasks implements LoggerAwareInterface {
     $lines = array_filter(explode("\n", $output));
     $changes = [];
     foreach ($lines as $line) {
-      $num_matches = preg_match("/([a-f0-9]{40}) (.+)/", $line, $matches);
+      preg_match("/([a-f0-9]{40}) (.+)/", $line, $matches);
       $commit_hash = $matches[1];
       $changes[$commit_hash] = $matches[2];
     }
--- a/RoboFile.php
+++ b/RoboFile.php
@@ -17,9 +17,32 @@ class RoboFile extends Tasks implements LoggerAwareInterface {
 
   use LoggerAwareTrait;
 
+  /**
+   * BLT root.
+   *
+   * @var string
+   */
   protected $bltRoot;
+
+  /**
+   * Binary.
+   *
+   * @var string
+   */
   protected $bin;
+
+  /**
+   * Drupal PHPCS standard.
+   *
+   * @var string
+   */
   protected $drupalPhpcsStandard;
+
+  /**
+   * PHPCS paths.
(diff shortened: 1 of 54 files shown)


danepowell:
@alexxed can you please review?

RoboFile.php
```diff
@@ -17,9 +17,32 @@ class RoboFile extends Tasks implements LoggerAwareInterface {
 
   use LoggerAwareTrait;
 
+  /**
+   * BLT root.
```
alexxed: Oh really, a comment is required for each var? I feel that this is a pretty bad habit, these usually just say the same thing as the var name.

settings/blt.settings.php
```diff
@@ -70,6 +71,7 @@
  * @var $site_path
  * This is always set and exposed by the Drupal Kernel.
  */
+// phpcs:ignore
```
alexxed: I can't immediately tell what's the ignored problem here.

RoboFile.php
danepowell: For now, yeah they are mostly just the variable name because there are _so_ many and I don't have time to document them better. Long-term, they should be documented more thoroughly.

Unfortunately we don't have much control over this. This is the [Drupal code comment standard](https://www.drupal.org/node/1354#var). I've looked and we can't disable it without disabling the requirement for _all_ short descriptions, or else all variable comments (including type hinting).

settings/blt.settings.php
danepowell: `$site_path` is a global variable, but PHPCS doesn't know that so it thinks it's undefined. We can't simply declare it as global either, because it's named inappropriately (not our fault as mentioned in the doc comment, it comes from Drupal core). I don't know of anything else to do besides ignore it.

danepowell:
@alexxed yep I'll create a ticket to get rid of the ignores.

settings/blt.settings.php
TravisCarden: Do note that [you can limit only specific sniffs](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file) to make the impetus clearer, e.g.:
```php
// phpcs:disable Generic.Commenting.Todo.Found
$xmlPackage = new XMLPackage;
$xmlPackage['error_code'] = get_default_error_code_value();
// TODO: Add an error message here.
$xmlPackage->send();
// phpcs:enable
```

danepowell: Thanks, it's a little more verbose and pretty much all of the ignores are for the same two reasons (global vars or shell exec) so I'll probably document it in the follow-up ticket instead.

PR #3093, opened 2018-09-12
Fixes #3089. More changes to update memcache settings to fix error

sugaroverflow:
Fixes #3089
--------

Changes proposed:
---------
(What are you proposing we change?)

-remove `memcache.yml` file
-remove reference to `memcache.yml` in `memcache.settings.php`

Steps to replicate the issue:
----------
(If this PR is not fixing a defect/bug, you can remove this section.)

1. Install fresh copy of BLT 9.1.6
2. Deploy code to Acquia Cloud
3. Navigate to `/admin/config/development/configuration`, run into error. 

Steps to verify the solution:
-----------
(How does someone verify that this does what you think does?)

1. Patch with recommendation in PR (remove `memcache.yml` and reference of it from `memcache.settings.php`. Make sure you're up to date with BLT `9.16` and that your memcache is pinned to 2.0-alpha7`.
2. Deploy to code to Acquia Cloud and navigate to `/admin/config/development/configuration`, everything should be working!

Issue #3089, opened 2018-09-12
MemcacheLockFactory does not have a method 'getPersistent' from BLT memcache.yml


sugaroverflow:
Hello 👋 

I was facing the same issue as #3074 was watching #3076 and #3079 as memcache.settings config changes were being made. 

I just updated to acquia/blt 9.16 from 9.15 hoping that the issue would be resolved, but am seeing a new memcache error.

 My system information:
* BLT version: 9.1.6

When I run this command:
```
drush status
```
or navigate to my site (white screen of `The website encountered an unexpected error. Please try again later and the following error output in the logs)

I get the following output:
```
ArgumentCountError: Too few arguments to function Drupal\memcache\MemcacheBackend::__construct(), 3 passed in /mnt/www/html/sitename/docroot/core/lib/Drupal/Component/DependencyInjection/PhpArrayContainer.php on line 101 and exactly 4 expected in Drupal\memcache\MemcacheBackend->__construct() (line 64 of /mnt/www/html/sitename/docroot/modules/contrib/memcache/src/MemcacheBackend.php).
 [error]  Drush command terminated abnormally due to an unrecoverable error.
```

And I expected this to happen:
```
[success] Cache rebuild complete.
```

I was hoping this might be related to the previous issues with the memcache settings and that someone might be able to help me move forward. Thank you!

ba66e77:
@sugaroverflow, can you check the version of `drupal/memcache` you have? On 9.1.6, we're [pinning to the alpha7 release](https://github.com/acquia/blt/blob/9.1.6/composer.suggested.json#L16) which has a constructor for the [MemcacheBackend class that takes 3 arguments](https://cgit.drupalcode.org/memcache/tree/src/MemcacheBackend.php?h=8.x-2.x&id=f30543ba2d991e5fd068963ebfe5bb00e163c818#n53). 

Looking at memcache's commit logs, it looks like a [commit was added yesterday that adds a 4th argument](https://cgit.drupalcode.org/memcache/commit/src/MemcacheBackend.php?h=8.x-2.x&id=3b645a4849c7689d8cd454e8de46e8cf232d6561). That change is in their beta1 release, though.

sugaroverflow:
Thanks for your help @ba66e77 ! 

I noticed that even though BLT was pinning the alpha7 release for memcache, my `composer.json` was installing the 2x-dev. I fixed that - and was able to get the site to load and access backend admin pages. 

[...]

PR #3730, opened 2019-06-25
Memcache updates

### LINKED ISSUE #3257: Support for drupal/memcache 2.0.0 (stable)
My system information:
* Operating system version: 16.04LTS
* BLT version: 9.1.x, 9.2.x

Acquia BLT does not support the current stable `drupal/memcache:^2.0.0` (8.x-2.0). This has been noted previously in #3220 with the following response:

@lcatlett said:
> BLT currently locks to the 2.0-alpha7 version of the memcache module because Acquia does not yet recommend that customers update to 2.0.

Acquia has recommended we update to `drupal/memcache:^2.0@stable` package to address issues with memcache we are seeing on the Acquia Cloud Enterprise platform. I note that some Acquia documentation has contrary advice. 

## Suggested Solution

- Provide updated BLT settings for the stable `drupal/memcache:^2.0.0` (8.x-2.0).
- Backport these settings to earlier BLT versions.
- Update BLT documentation where necessary.

lcatlett:
Fixes #3257 #3220  
--------

Changes proposed
---------
This PR updates the memcache module to the latest stable release and patches the memcache module to use the finally RTBC (but not yet merged) patch for large objects (see drupal.org/project/memcache/issues/435694). We/Acquia unfortunately have not yet been successful in getting this critical patch merged by the maintainers (@pat-patterson-acq can provide more background information internally if needed).

After several months of collaboration between product / engineering / CS this change also updates the BLT memcache settings at acquia/blt:settings/memcache.settings.php@10.x to what is the "new" Acquia supported memcache recommendation that will be in docs. See internal tickets PF-2147, SL-11016, PUB-19493 just to name a few.

This PR also removes the use of `DRUPAL_ROOT` from blt's memcache settings, as this constant is deprecated in favor of `\Drupal::root()` and/or `app.root`. See PF-1710, the core issue at https://www.drupal.org/project/drupal/issues/2328111 and CR at https://www.drupal.org/node/2330441


settings/memcache.settings.php
```diff
@@ -51,6 +60,10 @@
           'factory' => ['@memcache.backend.cache.factory', 'get'],
           'arguments' => ['container'],
         ],
+        'lock.container' => [
+          'class' => 'Drupal\memcache\Lock\MemcacheLockBackend',
+          'arguments' => ['container', '@memcache.backend.cache.container'],
+        ],
```
fiasco: So this means we're forcing locking into non-persistent store by default?
I've rarely come across a legitimate reason to put locking into memcache. Usually its an early indicator that an application is trying to do an atomic operation too regularly. Such operations have performance implications as too many requests have to wait to acquire the lock or let the atomic operation complete before continuing. This is why we see it as an "issue" so often. But really, its the symptom which this change would then hide.

subtree-splits/blt-project/composer.json
```diff
@@ -73,7 +73,11 @@
         "patchLevel": {
             "drupal/core": "-p2"
         },
-        "patches": []
+        "patches": {
+            "drupal/memcache": {
+                "12987574 - >1M data writes incompatible with memcached (memcached -I 32M -m32): #435694": "https://www.drupal.org/files/issues/2019-02-26/435694-118-d8-memcache-large-items.patch"
```
fiasco: 👍 🎉

anavarre: Ideally, to get the memcache issue merged, we should provide before/after stack metrics over the course of a few days. The one reason why memcache maintainers would be reluctant to merge this is if there would be a negative performance impact. Memcache is too critical in people's infrastructures to introduce performance regressions. If we could demonstrate that it doesn't have any, then I'm sure we'd get more support.

mikemadison13:
@lcatlett can you please confirm if there are any necessary update steps here for ACE / ACSF? I'm assuming just "apply the patch" and "update the memcache module" to 2.x and we're "good"?

mikemadison13:
this version of the code deployed cleanly (where several of the others that tried to call the Drupal services more directly did not. doing some additional testing now.

mikemadison13:
also did a deployment to ACSF with this version of the code without issue.

@lcatlett and @pat-patterson-acq were able to confirm that memcache was running server side in both of the hosting environments, which is a good sign!

danepowell:
More feedback and testing from everyone is certainly welcome, however I don't think we can merge this until it's publicly published in the Cloud documentation. BLT is not, nor should it be, the canonical reference for memcache settings implementation.

pat-patterson-acq:
@danepowell we can't document this on Cloud docs until the changes to BLT are published.  What else will it take to push this through?

alexxed:
@danepowell I think we can merge this, BLT docs should be updated together with the release.

danepowell:
At the very least, before we can merge this we need testing steps, as per the contribution guidelines.

fiasco:
Does this implicitly turn memcache on for all ACSF sites too? Because this setup could be bad for multisite installations if there isn't enough memory allocated to memcache (e.g. 128MB x N sites)

in D7, ACSF had a way to enable memcache on a per site basis (could reconfirm with them that D8 is supported too).

danepowell:
I'm currently waiting on the outcome of an internal discussion about who can vouch for, test, and support these changes going forward, since the BLT team has not traditionally been responsible for this.

pat-patterson-acq:
@danepowell we are evaluating a different approach to solving this problem - thanks for your patience.  Please keep this issue open until we have a path forward.

danepowell:
Long-term I think we should remove memcache from BLT entirely: https://github.com/acquia/blt/issues/3772

danepowell:
Thanks to everyone who provided feedback on these changes.

The Cloud memcache documentation has now been fully reviewed and updated:  https://docs.acquia.com/acquia-cloud/performance/memcached/enable/

Instead of trying to chase the sun and match these changes, BLT will likely take a different tack in future releases and provided a command that helps to initialize memcache settings on a per-project basis. Follow https://github.com/acquia/blt/issues/3777 for updates. I know that folks are anxious to update to the new memcache release and configuration, so this is a high priority.

settings/memcache.settings.php
stevenlafl: So your argument is to make the symptom -more pronounced-?

stevenlafl:
@danepowell Locking as a feature of the memcache module exists in Acquia documentation for Drupal 7. This is a feature regression for D8.

The way I see it is:
The change to remove it began here https://github.com/acquia/blt/issues/3089 where several people had tested removing PersistentLock from the yml file and reported this solves it, which is in line with suggestion here: https://www.drupal.org/project/memcache/issues/2995907. When @ba66e77 asked @sugaroverflow to submit a PR in that issue, @sugaroverflow created https://github.com/acquia/blt/pull/3093 which removes the yml file entirely. I see that this was introduced here as well.
While it does fix the problem (as does simply removing PersistentLock), it also removes the rest of the regular memcache locking feature. I do not see any discussion justifying the removal of the feature entirely. Can anyone provide insight here?

PR #3733, opened 2019-06-27
Remove acsf-tools from being required by default for acsf projects.

lcatlett:
Changes proposed
---------
ACSF tools is not an official Acquia project and neither support nor the ACSF engineering team are able to assist customers when they run into issues, nor is it recommended by the ACSF team. As it has caused significant issues in support, BLT should not include acsf-tools by default when running the `blt acsf` command.

The PR removes the package, so once merged the BLT release notes should likely call out that customers who wish to continue using the toolset should `composer require acquia/acsf-tools` in their root project composer.json.

### DIFF
--- a/src/Robo/Commands/Acsf/AcsfCommand.php
+++ b/src/Robo/Commands/Acsf/AcsfCommand.php
@@ -48,14 +48,6 @@ class AcsfCommand extends BltTasks {
     $this->invokeCommand('internal:composer:require', $package_options);
     $this->say("In the future, you may pass in a custom value for acsf-version to override the default version, e.g., blt recipes:acsf:init:all --acsf-version='8.1.x-dev'");
     $this->acsfDrushInitialize();
-    $this->say('Adding acsf-tools drush module as a dependency...');
-    $package_options = [
-      'package_name' => 'acquia/acsf-tools',
-      'package_version' => 'dev-9.x-dev',
-    ];
-    $this->invokeCommand('internal:composer:require', $package_options);
-    $this->say('<comment>ACSF Tools has been added. Some post-install configuration is necessary.</comment>');
-    $this->say('<comment>See /drush/Commands/acsf_tools/README.md. </comment>');
     $this->say('<info>ACSF was successfully initialized.</info>');
     $this->say('Adding nedsbeds/profile_split_enable module as a dependency...');
     $package_options = [


danepowell:
I'm hesitant to remove support for a tool that might be used by a substantial fraction of BLT + ACSF users. I don't have hard metrics on the overlap between BLT and acsf-tools users, but according to packagist, [acsf-tools](https://packagist.org/packages/acquia/acsf-tools/stats) is nearly as popular as [BLT itself](https://packagist.org/packages/acquia/blt/stats). To me, this strongly suggests we should continue including acsf-tools.

As to the arguments for removing it:
- BLT includes over 250 Composer packages that aren't directly supported by Acquia. I'm not sure that "being unsupported" by itself is a valid criteria for not including a package.
- I definitely understand that having "acsf" in the package name might imply a level of support that's not there, but that's not really BLT's problem. Instead of removing this module, should we just add a disclaimer that "this module is not supported by Acquia"? Even better might be if acsf-tools added a post-composer-install step that printed that disclaimer.
- If the package itself is broken or causing problems that's not good. But I don't see evidence of instability or bugs right now, and the extremly low ratio of open issues to packagist popularity (compared to BLT) would suggest this is not the case.

I believe @sdelbosc and @msherron maintain acsf-tools, would they like to weigh in on this? And @mikemadison13 as someone who might use acsf-tools more frequently than me.

lcatlett:
@danepowell - I am not sure the packagist numbers are an accurate representation of the customers actually using acsf-tools, as it is likely it appears as popular as blt simply because blt is including it for acsf projects by default when running the blt acsf command. there are a number of issues with commands such as acsf-tools-ml which cause a substantial amount of support issues. There was an internal email thread which went around on March 26 where deprecating acsf-tools was discussed which you may want to re-review for context

sdelbosc:
I agree with @lcatlett on the usage.
If the link between BLT and acsf-tools is only what is in this PR I think it would make sense to merge it.

alexxed:
I suggest a phased approach. 
Phase 1: update should not remove acsf-tools, but print out a deprecation message during blt-10, install should not include it
Phase 2: remove acsf-tools in blt-11

danepowell:
I can find nothing in the acsf-tools repo to indicate that it's deprecated or unstable: https://github.com/acquia/acsf-tools

I think it would be really odd and inappropriate for BLT to unilaterally say "hey this other product over here is unsupported, don't use it".

If I see a deprecation notice for ACSF Tools go up on the project itself, I would definitely remove it from BLT immediately. Otherwise, I agree with Alex that the best approach would be to wait for BLT 11 to remove it. That would give folks time to transition over to ADS, which may duplicate much of the ACSF Tools functionality.

PR #3736, opened 2019-06-28
Added some clarifying language to Install.md

jprasmussen:
Changes proposed
---------
The documentation for BLT on [Acquia Docs](https://docs.acquia.com/blt/install/) has slightly diverged from the docs in this repo. Some wording has been changed there and implies that the user has installed BLT when they have actually just installed the prerequisites. In an effort to clarify this documentation I've open a ticket and PR for the Docs team and now doing the same here.

In this PR, the changes made are after the System Requirements and are to help a user more clearly understand what the next steps are with installing BLT.

Additional details
-----------
- https://docs.acquia.com/blt/install/
- https://backlog.acquia.com/browse/DOC-19034

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -86,9 +86,13 @@ Linux is fully supported by BLT and DrupalVM and shares many of the same depende
 
 # Installing BLT
 
-Choose your own adventure:
+After you have successfully installed the prerequisites, you can now proceed with installing BLT.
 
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks. Just to be clear, I still consider the docs in this repo to be the system of record for changes, but it seems like the docs team has been making changes to docs.acquia.com and not syncing them back. We're still working on a process for that.

I think these changes make sense regardless, so... mergy mergy. Thanks again.

PR #3737, opened 2019-06-28
Fixes #3645: Allow EnvironmentDetector to be overridden.

### LINKED ISSUE #3645: Extensible EnvironmentDetector "service".
BLT "ships" with environment detection for dev, test and prod. In our Acquia account, we have an "Acceptance" environment, and a "Sandbox" environment that we use on a regular basis, and even "RA". In the past, I have included a settings file after `blt.settings.php` that says something along the lines of... `$is_dev_env = ($is_dev_env || $ah_env == 'sandbox');` to make my sandbox environment behave like DEV in all BLT code. However, as of #3570, this method really no longer works, as I have to call `EnvironmentDetector::isAhDev()` for example, and can no longer treat my sandbox environment like dev. For the time being, I'm going to just "re-set" all the variables that went away in 3570, and set them via the `EnvironmentDetector` class, and then include my overrides.

Unsure?

I think I'm going to, and this may be the ultimate solution, but for now I'll `extend` the `EnvironmentDetector` class, and overwrite the methods I want, and then just use MY class to resolve any IF statements I was previously using BLT variables for...
Thoughts?

### LINKED ISSUE #3689: Move Tugboat integration to custom plugin
I'd like to move the Tugboat integration (as well as other non-core CI integrations) to a separate composer plugin, since it's not something that we can directly support.

This would also be really good to serve as a reference / template for creating BLT plugins.

The problem is that there is no way to override the environment detector to detect Tugboat as a CI environment, so this is blocked by #3645

danepowell:
Fixes #3645 
--------

Also fixes #3689 although we need to open-source that plugin before merging.

Changes proposed
---------
- Modify the environment detector to automatically call any subclassed versions of a few specific methods that we know customers want to override.
- Remove Tugboat integration (it has been moved to a separate plugin that demonstrates the above functionality).

In order to override the detector, downstream projects (either a BLT-based project or a BLT plugin such as the linked tugboat plugin) simply need to provide their own class that extends EnvironmentDetector. It needs to have the namespace / file structure `Foo\Blt\Plugin\EnvironmentDetector\FooDetector`. For instance, here's the TugboatDetector:
```php
<?php

namespace Acquia\BltTugboat\Blt\Plugin\EnvironmentDetector;

class TugboatDetector extends EnvironmentDetector {
  public static function getCiEnv() {
    return isset($_ENV['TUGBOAT_URL']) ? 'tugboat' : null;
  }

  public static function getCiSettingsFile() {
    return sprintf('%s/vendor/acquia/blt-tugboat/settings/tugboat.settings.php', dirname(DRUPAL_ROOT));
  }
}
```
Additionally, downstream projects need to add that to the classmap with this composer key:
```js
"autoload": {
    "classmap":["src/Blt/Plugin/EnvironmentDetector/TugboatDetector.php"]
}
```

This needs a lot of testing, and we also need to open-source the tugboat plugin (private right now: https://github.com/acquia/blt-tugboat).


danepowell:
@dpagini you probably want to be the first to test this. Let me know if it solves your use case.

dpagini:
Thanks @danepowell! Let me give this a shot and report back...

danepowell:
@dpagini thanks! Note that I pushed a bug fix just a few seconds ago to fix a possible segfault.

danepowell:
Awesome, the tugboat plugin is now available here: https://github.com/acquia/blt-tugboat

We'll just need to add a change record that upgrading users will need to `composer require acquia/blt-tugboat` if they want to continue using that.

PR #3739, opened 2019-07-01
Fixes #3738: Failing deprecation tests.

### LINKED ISSUE #3738: Fix broken deprecation tests
Started breaking with this PR: https://github.com/acquia/blt/pull/3710

I think it's a version conflict with phpstan and an mglaman package (ORCA provides its own versions of these).

danepowell:
Fixes #3738 
--------

Changes proposed
---------
- There seems to be a problem with "mglaman/phpstan-drupal": "0.11.9" causing fatal errors with neon packages. Not sure what, but pinning it to 0.11.7 helps. Before we merge this we should probably have a plan to revert it / update to the latest version. ORCA will likely face the same problem, coordinate with them.

This isn't intended to make all deprecation tests pass, just to eliminate the meta-error caused by phpstan. We will fix all of the deprecation tests in #3712.

### DIFF
--- a/settings/drush-request-trace.settings.php
+++ b/settings/drush-request-trace.settings.php
@@ -19,7 +19,7 @@ if (isset($_ENV['AH_SITE_ENVIRONMENT']) && PHP_SAPI === 'cli') {
   putenv('HTTP_HOST=' . $_SERVER['HTTP_HOST']);
 
   if (function_exists('drush_get_context')) {
-    $cli_args = drush_get_context('argv');
+    $cli_args = $GLOBALS['argv'];
     $cli_args[0] = 'drush';
 
     // Ensure each argument is wrapped in quotes.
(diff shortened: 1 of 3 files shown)


settings/drush-request-trace.settings.php
```diff
@@ -19,7 +19,7 @@
   putenv('HTTP_HOST=' . $_SERVER['HTTP_HOST']);
 
   if (function_exists('drush_get_context')) {
-    $cli_args = drush_get_context('argv');
+    $cli_args = $GLOBALS['argv'];
```
danepowell: This is the only deprecated function call we can fix at the moment. I've tested it in the canary application and it's safe.

danepowell:
New ORCA release has a hotfix that may let us undo the phpstan changes: https://github.com/acquia/orca/releases/tag/v1.0.0-alpha17

PR #3740, opened 2019-07-02
Fixes the sanitize function to use the configured docroot.

b-sharpe:
Simply changes the sanitize function of deployment to use the configured docroot instead of being hard-coded.

### DIFF
--- a/src/Robo/Commands/Deploy/DeployCommand.php
+++ b/src/Robo/Commands/Deploy/DeployCommand.php
@@ -493,6 +493,7 @@ class DeployCommand extends BltTasks {
    * Removes sensitive files from the deploy dir.
    */
   protected function sanitize() {
+    $docroot = $this->getConfigValue('docroot');
     $this->say("Sanitizing artifact...");
 
     $this->logger->info("Find Drupal core text files...");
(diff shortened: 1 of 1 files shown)


danepowell:
Why would you want to do this?

`docroot` is typically defined as `${repo.root}/docroot`: https://github.com/acquia/blt/blob/736e40ff107e69ed39a27d462578b1275fe338de/config/build.yml#L60

Wouldn't this change then cause deploy operations to happen in `${repo.root}/deploy/${repo.root}/docroot`, which is a nonsensical path?

b-sharpe:
@danepowell Because docroot should be configurable. Just because Acquia uses it as it's default doesn't mean every host does. Pantheon for example uses "web" and some other hosts use "html"

Every other function seems to allow it to be configurable, but since santize does not and is part of the artifact:deploy script, it prevents deployment if not using docroot

danepowell:
I see. That makes sense, but I think the implementation is going to need some work because as I mentioned above, `docroot` is typically defined to include the `repo.root`. You would need a separate configuration variable for just the docroot name.

danepowell:
Closing due to failing tests and lack of response. If you'd like to respond to the feedback above and clean up the PR, feel free to reopen.

PR #3746, opened 2019-07-08
Removed unused functions and improved docs

danepowell:
I realized we have a ton of vestigial functions in Robofile from older testing processes. We can safely remove them now.

We've also changed out we recommend that people link BLT to test projects, I've updated documentation accordingly.


danepowell:
@alexxed still need your review here

PR #3758, opened 2019-07-18
Bump PHP version in use on Travis CI

TravisCarden:
In light of the fact that Acquia will be retiring support for PHP 7.1 on October 1, the ORCA product team has decided to increase the PHP version in use on ORCA tests to 7.2 today. If you wish to run non-default jobs on another version of PHP, you can do so via matrix settings, e.g.:

```yaml
matrix:
  include:
    - { name: "PHP 7.1 job", env: SOME_VARIABLE=1, php: "7.1" }
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: php
 
-php: "7.1"
+php: "7.2"
 
 addons:
   chrome: stable


danepowell:
Thanks. The rest of BLT (DrupalVM, Pipelines config, etc...) already defaults to 7.2 I think so this brings everything in line.

PR #3763, opened 2019-07-19
Removed ORCA workaround script.

danepowell:
Deprecation tests should still pass here as https://github.com/acquia/orca/pull/33 has been updated to obviate this workaround.


danepowell:
Travis is implementing an upstream change to work around the side effect we discovered here

danepowell:
I think the upstream change caused regressions elsewhere (with Lightning) and was thus reverted. We'll come back to this if it gets fixed upstream.

PR #3768, opened 2019-07-22
DT-577: Ensure after_failure runs during Travis tests.

### LINKED ISSUE #3687: after_failure is not triggered when travis build fails on behat tests
I'm using blt 10.0.1 with behat running on travis ci. travis.yml is generated by blt with some minor tweaks.

I want to trigger some tasks when behat tests fail on travis build, for example: uploading screenshot and test reports to S3. The tasks are added in `after_failure` step on .travis.yml

Somehow `vendor/acquia/blt/scripts/travis/run_tests` script in `script` step stops the whole build process when `blt tests:all` returns a failed test result. `after_failure` step is never triggered.

Some articles mentions about `set -e` issue on travis script, but doesnt seem to help. I've already tried a few things, including removing `set -e`, handling error code manually, calling `blt tests:all` directly in `after_failure`, but none triggers that `after_failure` step. I have not tried to alter `tests:behat:run` command and take out the `->stopOnFail()` yet, not sure if that has something to do with this issue.

There are 2 articles in the past for this same problem but it doenst seem to have a working solution:
https://github.com/acquia/blt/pull/3043
https://github.com/acquia/blt/pull/2521

Does anyone have similar problem and happen to know how to solve it?

danepowell:
Fixes #3687 
--------

Note that the run_tests script is `source`d via .travis.yml. That means it operates in the same shell context as the master Travis script. So when it exits, the whole shell exits and that's why subsequent steps don't run.

To avoid that, let's return instead of exiting. This prevents subsequent test steps from running, but allows the master Travis build to continue. We also need to be good bash citizens and unset verbose mode before returning.

### DIFF
--- a/scripts/travis/run_tests
+++ b/scripts/travis/run_tests
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
-# Ordinarily we'd use set -e here but that breaks travis_wait, so we use exit
+# Ordinarily we'd use set -e here but that breaks travis_wait, so we use return
 # pattern below.
 
 set -v
 
-blt validate:all --no-interaction || exit 1
-travis_wait blt setup --define drush.alias='${drush.aliases.ci}' --environment=ci --no-interaction --verbose || exit 1
-blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --environment=ci --no-interaction --verbose || exit 1
+blt validate:all --no-interaction || { set +v && return 1; }
+travis_wait blt setup --define drush.alias='${drush.aliases.ci}' --environment=ci --no-interaction --verbose || { set +v && return 1; }
+blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --environment=ci --no-interaction --verbose || { set +v && return 1; }
 
 set +v


scripts/travis/run_tests
```diff
@@ -1,12 +1,12 @@
...
-# Ordinarily we'd use set -e here but that breaks travis_wait, so we use exit
+# Ordinarily we'd use set -e here but that breaks travis_wait, so we use return
 # pattern below.
 
 set -v
 
-blt validate:all --no-interaction || exit 1
-travis_wait blt setup --define drush.alias='${drush.aliases.ci}' --environment=ci --no-interaction --verbose || exit 1
-blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --environment=ci --no-interaction --verbose || exit 1
+blt validate:all --no-interaction || { set +v && return 1; }
```
alexxed: I think this should avoid the repetition
```bash
{
set -v
blt validate:all --no-interaction
travis_wait blt setup --define drush.alias='${drush.aliases.ci}' --environment=ci --no-interaction --verbose
blt tests:all --define drush.alias='${drush.aliases.ci}' --define tests.run-server=true --environment=ci --no-interaction --verbose
set +v
return 0
} || { set +v && return 1 }

```

danepowell: I think that will always return exit code 0. If any of validate/setup/test fail, the script will just keep running, there's nothing to force it to stop running.

PR #3773, opened 2019-07-24
DT-633: blt:dev:link-composer lando support

danepowell:
Might be a touch tricky to test. Start by checking out this branch in your local BLT repo. Set up a test site as a supported sibling directory (i.e. either `~/blt`+`~/my-blt-project` or `~/packages/blt`+`~/sites/my-blt-project`). Make sure the new site has Lando initialized.

Run `blt blt:dev:link-composer` once, which will run using the old version of this command and link BLT, but fail during the VM setup. That's expected.

Run it a second time and it should configure your Lando file without error to mount your local version of BLT. Now you can confirm that BLT commands still run inside Lando.

Long-term, I think we need to find a home for this command outside of BLT since it's such a chicken-and-egg problem to test.


danepowell:
@ionutz2k2 Adrian already manually tested this if you just want to review the code changes.

PR #3780, opened 2019-07-30
Remove internal TravisCI tests from acquia/blt-project

danepowell:
The acquia/blt-project package has internal Travis CI tests that run as a legacy from when it was a separate project from BLT. Now that it's a subtree these tests get run along with the rest of the ORCA suite. There's no need to run them separately, and they create a non-trivial maintenance burden.


danepowell:
Tests temporarily failing, unrelated to this PR. Upstream issue: https://github.com/acquia/orca/issues/37

PR #3782, opened 2019-07-30
DT-584: Support custom filesets via BLT plugins

### LINKED ISSUE #3642: Custom filesets on BLT 10.x
**Describe the bug**
I was previously using custom filesets in BLT. The upgrade to 10.0 appears to have broken this functionality. The example filesets uses a namespace of `Acquia/Blt/Custom` but I believe that is broken with the new "BLT Plugin" functionality.

**To Reproduce**
_Follow the instructions on the [Acquia documentation page](https://docs.acquia.com/blt/extending-blt/#adding-or-overriding-filesets)._

**Expected behavior**
My custom fileset is available to linting functions.

**Detailed error output**
image

danepowell:
Fixes #3642 
--------

Changes proposed
---------
- Auto discover fileset definitions exposed as BLT plugins (in the `*\Blt\Plugin\Filesets` namespace with filename `*Filesets.php`)
- Update examples and documentation to match

The idea here is that fileset definitions become plugins just like custom commands and hooks. All you have to do is name and namespace them according to the above pattern and expose them via PSR4. This could be in the default `blt/src` directory, another directory in a BLT project, or even a separate Composer package.

Steps to replicate the issue
----------
1. Create a new BLT project
2. [curl this file](https://github.com/danepowell/blt/blob/9ac85fa499a0a3d41413522a9e7e6ac36e35a2dc/scripts/blt/examples/Filesets/ExampleFilesets.php) and place it at `blt/src/Blt/Plugin/Filesets/ExampleFilesets.php`
3. Follow the documentation to create a custom fileset definition and configure a BLT linter to use it.
4. Run the linter.

----------
1. Error because custom fileset definition cannot be found.

-----------
1. Linting occurs without error.

### DIFF
--- a/docs/extending-blt.md
+++ b/docs/extending-blt.md
@@ -1,18 +1,20 @@
 # Extending / Overriding BLT
 
-BLT uses Robo to provide commands.
+BLT provides an extensive plugin and configuration system to allow for customization.
 
 ## Adding a custom Robo Command or Hook
 
-Robo uses the [Annotated Command](https://github.com/consolidation/annotated-command) library to enable you to add commands as well as hook into existing BLT commands. This allows you to execute custom code in response to various events, typically just before or just after a BLT command is executed.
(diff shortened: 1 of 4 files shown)


danepowell:
@dpagini @longmtran can you review this, since you originally reported the bug?

dpagini:
By looks alone, this looks pretty good to me. I'll try to give a shot in practice.

PR #3792, opened 2019-08-02
Clear cache before sql-sanitise on sync

leonk:
Fixes # 3791
--------

Changes proposed
---------
Clear drupal cache before sql-sanitise is run, to ensure plugins are registered.


leonk:
Hi @danepowell I tested with the `drush cr` removed, and this is still working for me.  I've updated the PR.

PR #3794, opened 2019-08-02
Removed redundant method override.

danepowell:
Robo recently updated the parent definition of confirm() to be exactly the same as ours, so there's no need to override it anymore: https://github.com/consolidation/Robo/blob/e5a6ca64cf1324151873672e484aceb21f365681/src/Common/IO.php#L132


danepowell:
Blocked by upstream issue: https://github.com/acquia/coding-standards-php/pull/3

PR #3801, opened 2019-08-07
DT-652: Migrate memcache settings to standalone package.

### LINKED ISSUE #3777: Change memcache settings management strategy
Currently, BLT provides built-in memcache settings based on those recommended by the Cloud team: https://docs.acquia.com/acquia-cloud/performance/memcached/enable/

The fact that these are built into the BLT codebase causes a lot of pain for maintainers as well as customers:

- They cannot be directly modified by customers
- Even customers who don't use memcache will still have the dependency in their codebase
- Everyone (customers, other Acquia teams) gets confused about who "owns" these settings and who should Support them
- Every time the recommendations change, the BLT team has to update these settings and inevitably deal with support requests as fallout.
- BLT updates are tightly coupled to memcache updates, customers can't upgrade one independently of the other.

I propose that we remedy this by removing the baked-in settings, and replacing them with a command "recipes:memcache:init". This command will curl/wget the recommended settings files directly from the Cloud docs and install them into a customer's project codebase, and add memcache as a dependency to their composer.json.

I think we can do this as part of a minor release as long as we run the initialization automatically for upgrading customers, and provide a clear changelog.

This will be technically simple but require a fair bit of testing. I also suspect we'll find clashes between the ACE and ACSF versions of the memcache settings, since they don't appear isomorphic (the ACE settings will be enabled on ACSF). We may need to work with the docs team on that.

danepowell:
Fixes #3777 
--------

Changes proposed
---------
- Remove memcache settings from BLT
- Replace with a dependency on https://github.com/acquia/memcache-settings

This shouldn't impact existing users since we automatically depend on and include the new settings. However it does require a change record because the settings themselves (as well as the memcache version) have been updated as part of this change.


settings/cache.settings.php
```diff
@@ -18,38 +18,11 @@
...
- * @see https://www.drupal.org/node/2766509
+ * Use memcache as cache backend if Acquia configuration is present.
  */
-if ((EnvironmentDetector::isAhEnv() || EnvironmentDetector::isCiEnv()) &&
-  // phpcs:ignore
-  array_key_exists('memcache', $settings) &&
-  // phpcs:ignore
-  array_key_exists('servers', $settings['memcache']) &&
+$repo_root = dirname(DRUPAL_ROOT);
+$memcacheSettingsFile = $repo_root . '/vendor/acquia/memcache-settings/memcache.settings.php';
```
greylabel: Should this still be wrapped in `EnvironmentDetector::isAhEnv()` or similar so it only includes on Acquia?

danepowell: The [upstream memcache settings file](https://docs.acquia.com/_downloads/cloud-memcache-d8.php) already includes this check in its preamble, so it would be redundant for BLT to check as well. Let me know if you think that would be insufficient.

I figured this would cause confusion, maybe I should have noted it in the code. Oh well :smile:

docs/memcache.md
```diff
@@ -2,26 +2,9 @@
...
-If you are upgrading from a previous version of BLT to 9.2.x, make sure and re-generate your factory hooks using:
-
-```
-recipes:acsf:init:hooks
-``` 
-
-This will create a new memcache factory hook for use on ACSF.
+## Acquia Cloud and Site Factory
 
+To enable memcache integration in a Cloud hosting environment, ensure that you've installed the `acquia/memcache-settings` Composer package. If this package is present, BLT will automatically include the relevant settings files that enable memcache integartion.
```
anavarre: s/integartion/integration

docs/memcache.md
```diff
@@ -2,26 +2,9 @@
...
-As of BLT 9.2, the factory hooks contain the necessary code to handle memcache integration with ACSF provided that your subscription and hardware are properly configured. [Using Memcached on Acquia Cloud](https://docs.acquia.com/acquia-cloud/performance/memcached/) provides additional information about this.
-
-If you are upgrading from a previous version of BLT to 9.2.x, make sure and re-generate your factory hooks using:
-
-```
-recipes:acsf:init:hooks
-``` 
-
-This will create a new memcache factory hook for use on ACSF.
+## Acquia Cloud and Site Factory
```
anavarre: s/Site Factory/Acquia Cloud Site Factory/

subtree-splits/blt-project/composer.json
```diff
@@ -8,14 +8,14 @@
         "acquia/blt": "10.x-dev",
         "acquia/drupal-spec-tool": "*",
         "acquia/lightning": "^4.0.0",
+        "acquia/memcache-settings": "*",
```
alexxed: I see this module includes 2.0, not 2.0-alpha7. Is that a problem? There was a discussion on another PR I remember.

danepowell: Yep 2.0 is the version that we now publicly recommend, and that works with the recommended memcache settings.

jedgar1mx:
For `multisites`, do we just add 
`
$repo_root = dirname(DRUPAL_ROOT);
$memcacheSettingsFile = $repo_root . '/vendor/acquia/memcache-settings/memcache.settings.php';
if (file_exists($memcacheSettingsFile)) {
  require $memcacheSettingsFile;
}
`
to every settings.php

Thanks

danepowell:
You don't need to manually include memcache like that for any type of codebase (single or multi site) as long as you are using BLT. Just make sure that your settings.php includes BLT's core settings file (it should if you generated the project/site with BLT in the first place):

`require DRUPAL_ROOT . "/../vendor/acquia/blt/settings/blt.settings.php";`

PR #3803, opened 2019-08-09
DT-575: Fixes #3802: Improve documentation of multisite Behat config and profile overrides 

### LINKED ISSUE #3690: Document how to run tests against multiple sites
We have separate documentation on how to set up multisite projects, ACSF projects, CI, and testing for projects in general, but what is lacking is specific guidance on how to construct and run test suites for multisite installations, especially in the context of core CI tools (Travis / Pipelines).

Obviously there are a multitude of strategies for testing multisite projects, so I think we should focus on the simplest scenario to start, which is to run a separate Behat suite per-site.

The documentation should make sure to clarify that this is only one possible strategy (the other main strategy being to only test a single "canary" site or core platform features), and should also make sure to address the whole-stack setup process (including how to configure CI tools to e.g. install multiple sites / dbs).

Also make sure to capture the feedback from #3275, and maybe use my comment there as a basis for the documentation.

### LINKED ISSUE #3802: BLT Multisite Behat testing does not change local.extensions.Behat\MinkExtension.base_url when --site flag is passed
**Describe the bug**
I have a Drupal 8 multisite. I have the following site structure:

docroot/sites/site1/blt.yml
docroot/sites/site2/blt.yml

In each of the multisite `blt.yml` files, I have explicitly defined each sites `project.local.uri` in the sites `blt.yml` file. Each multisite has this value. In theory, running `blt tests:behat:run --site=site1' should change the `local.extensions.Behat\MinkExtension.base_url` - at least [per the comment on 2860](https://github.com/acquia/blt/issues/2860#issuecomment-480052280) (as this is not documented). This is not happening. Further, a better solution would be to allow passing different Behat `--profile` values via BLT. Tags are already passed. Why in the world not profiles? (e.g. `blt test:behat:run --site-site1 -d profile=site1`)

After running a `blt tests:behat:run --site=site1`, I get an error due to a mismatch between the `local.extensions.Behat\MinkExtension.base_url` defined in `local.yml` and the `project.local.uri`. Because I have two sites in my multisite, the `local.Behat\MinkExtension.base_url` can only exist for ONE URI event though I have two sites to test. It's either going to be, for example, http://local.site1.com or http://local.site2.com or null if left blank. BLT never intervenes to change this value per the note on https://github.com/acquia/blt/issues/2860#issuecomment-480052280 It is supposed to? I've never once seen this happen. 

> For instance, if you have a site `bar` with local uri `http://local.bar.com`, then running `blt tests:behat --site=bar` will rewrite local.yml such that the base_url is `http://local.bar.com`. I just tested this to verify it works exactly as expected.

I have never ever seen this work successfully. This does not work in my setup. 

```
blt tests:behat:run --site=site1
[warning] Behat is not configured properly.
BLT can (re)generate tests/behat/local.yml using tests/behat/example.local.yml.
 Do you want (re)generate local Behat config in tests/behat/local.yml? (y/n) n 
// Note I do not ever want to rewrite my entire local.yml. Why would I? I have paths to features defined per site etc. 
[warning] project.local.uri in blt.yml does not match local.extensions.Behat\MinkExtension.base_url in local.yml.
[warning] project.local.uri = http://local.site1.com
[warning] local.extensions.Behat\MinkExtension.base_url = https://local.site2.com
[error]  Behat is not configured properly. Please run `blt doctor` to diagnose the issue. 
```

Note that the error of [warning] `local.extensions.Behat\MinkExtension.base_url = https://local.site2.com` could just as easily be an error that it has no value (if left empty). If I explicitly change it to the site I WANT to test by rewriting the yaml, then it will work. But It's just supposed to work via BLT, right? It doesn't.

**To Reproduce**
1. Ensure you have a multisite and that multisite1 has a `blt.yml` file with `project.local.hostname` defined
2. Define a `local.Behat\MinkExtension.base_url` for your OTHER multisite or leave blank in `tests/behat/local.yml` <-- only one profile is allowed, which is a huge problem. We need multiple for multisites. 
3. Run `blt tests:behat:run --site=site1`
4. See BLT error that `project.local.hostname` does not match `local.Behat\MinkExtension.base_url`

**Expected behavior**
I expect BLT to identify that each multisite has a unique base_url and for Behat to test off the multisite's `project.local.url` defined in the multisites `blt.yml`. However it is very much NOT possible to have more than one `base_url` defined in a `local` profile and BLT has never overridden this value per the comment on issue . BLT must be able to handle more than just the local profile or Behat testing will not work. I'd much prefer being able to change the Behat profile used. 

**BLT doctor output (note BLT doctor is not very useful fur multi-site configurations)**
```
+---------------------------+--------------------------------------------+
| Property                  | Value                                      |
+---------------------------+--------------------------------------------+
| blt-version               | 10.2.1                                     |
| composer-version          | Composer version 1.9.0 2019-08-02 20:55:32 |
| root                      | /app/docroot                               |
| stacks.drupal-vm.inited   | false                                      |
| stacks.dev-desktop.inited | false                                      |
| uri                       | default                                    |
+---------------------------+--------------------------------------------+
+------------------------------------------+--------------------------------------------------------------+
| Check                                    | Problem                                                      |
+------------------------------------------+--------------------------------------------------------------+
| BehatCheck:checkBaseUrl:uri              | base_url in tests/behat/local.yml does not match the site    |
|                                          | URI.                                                         |
|                                          |   Behat base_url is set to                                   |
|                                          | https://naswa.lndo.site.                                     |
|                                          |   Drush site URI is set to default.                          |
| ComposerCheck:composer.autoload.psr-4    | The Composer configuration for autoload.psr-4 differs from   |
|                                          | BLT's default, recommended values.                           |
|                                          |   Change your configuration to match BLT's defaults in       |
|                                          |                                                              |
|                                          | vendor/acquia/blt/subtree-splits/blt-project/composer.json.  |
| DbCheck:checkRequiredKeys                | drush status is missing the 'db-hostname' key.               |
| DrupalCheck:checkDrupalBootstrapped      | Could not bootstrap Drupal via drush without alias.          |
| DrupalCheck:checkDrupalInstalled         | Drupal is not installed.                                     |
|                                          |                                                              |
|                                          | Run `blt drupal:install` to install Drupal locally.          |
| FileSystemCheck:checkFileSystem:%files   | Public files directory is not set.                           |
| FileSystemCheck:checkFileSystem:%private | Private files directory is not set.                          |
| FileSystemCheck:checkFileSystem:%temp    | Temporary files directory is not set.                        |
| NodeCheck:checkNodeVersionFileExists     | .node-version file exists                                    |
| SimpleSamlPhpCheck:checkSimpleSamlPhp    | /app/vendor/simplesamlphp/simplesamlphp/metadata/saml20-idp- |
|                                          | remote.php is missing. Run `blt                              |
|                                          | source:build:simplesamlphp-config`.                          |
| WebUriCheck:checkUri                     | Site URI is not set                                          |
|                                          | Is options.uri set correctly in                              |
|                                          | /app/docroot/sites/default/local.drush.yml?                  |
+------------------------------------------+--------------------------------------------------------------+
[error]  BLT Doctor discovered one or more critical issues. 
```
**System information**
* Operating system version: [19.04
* BLT version: 10.2.1  

**Additional context**
I'm happy to hop on a Zoom call to demonstrate this failure.  

I will note that I have found a way to get things working that does not conform to comment https://github.com/acquia/blt/issues/2860#issuecomment-480052280 nor documentation (as this is not documented functionality).

The way I have managed to make things work is to define additional keys in my `sites/sites1/blt.yml` file. Specifically I am now defining `behat.config` and `behat.profile` and my `behat.config` is a fully qualified path to the file `site1.behat.yml` which defines a `site1` profile key. I am then able to run the `lando blt tests:behat:run --site=site1` command successfully.

brooke-heaton:
Add an explicit example of how multiple Behat configuration files and profiles can be defined in multisite specific blt files.

Fixes #3802
Fixes #3690

Changes proposed
---------
There's been much confusion over how to properly configure BLT for Behat tests in a multisite configuration.  This documentation makes it explicitly clear how multiple configuration files and profiles can be used and how each multisite can then differentiate the Behat tests configuration per site.

### DIFF
--- a/docs/multisite.md
+++ b/docs/multisite.md
@@ -46,7 +46,20 @@ drush:
     remote: mysite.test
 ```
 
-Then, to refresh your local site, you could run: `blt drupal:sync --site=mysite`.
+If you wish to use a Behat profile other than `local.yaml` for each site, you can also override the configuration source file and profile values in in `docroot/sites/mysite/blt.yml`:
+
+```yaml
+behat:
+  tags: mysite
+  # When tests:behat:run target is executed, this Behat profile will be used.
+  # @see tests/behat/example.local.yml for Behat profile definition.
+  profile: mysite
+  config: /app/tests/behat/mysite.behat.yml
+```
+
+Per the above configuration, a functional 'mysite.behat.yml' configuration file would need to exit at `app/tests/behat/` and would need to include correct configuration for a `mysite` profile. This would permit more granular configuration of Behat suites, contexts and extensions per multisite.
+
+Running the following command would then trigger Behat to use the `mysite` configuration and profile: `blt tets:behat:run --site=mysite`.
 
 ## Acquia Cloud setup
 
--- a/docs/multisite.md
+++ b/docs/multisite.md
@@ -13,7 +13,7 @@ There are two parts to setting up a multisite instance on BLT: the local setup a
     * Sets up a new drush alias.
     * Sets up a new vhost in the box/config.yml file.
     * Sets up a new MySQL user in the box/config.yml file.
-    
+
     Some manual configuration is still required via the following steps.
 
 1. Set the new site's local database credentials in the `docroot/sites/{newsite}/settings/local.settings.php` file to ensure your new site connects to the correct database.
@@ -45,6 +45,8 @@ drush:
     local: mysite.local
     remote: mysite.test
 ```
+
+Then, to refresh your local site, you could run: `blt drupal:sync --site=mysite`.
 
 If you wish to use a Behat profile other than `local.yaml` for each site, you can also override the configuration source file and profile values in in `docroot/sites/mysite/blt.yml`:
 


docs/multisite.md
```diff
@@ -46,7 +46,20 @@ drush:
     remote: mysite.test
 ```
 
-Then, to refresh your local site, you could run: `blt drupal:sync --site=mysite`.
```
danepowell: Was this removed for a particular reason? I think it needs to stay.

docs/multisite.md
```diff
@@ -46,7 +46,20 @@ drush:
...
+  tags: mysite
+  # When tests:behat:run target is executed, this Behat profile will be used.
+  # @see tests/behat/example.local.yml for Behat profile definition.
+  profile: mysite
+  config: /app/tests/behat/mysite.behat.yml
+```
+
+Per the above configuration, a functional 'mysite.behat.yml' configuration file would need to exit at `app/tests/behat/` and would need to include correct configuration for a `mysite` profile. This would permit more granular configuration of Behat suites, contexts and extensions per multisite.
+
+Running the following command would then trigger Behat to use the `mysite` configuration and profile: `blt tets:behat:run --site=mysite`.
```
danepowell: Can you please also add a note to the Behat section of testing.md directing people here for multisite testing guidance?

brooke-heaton: Sorry @danepowell That omission wasn't intentional.

docs/multisite.md
brooke-heaton: Sorry @danepowell That omission wasn't intentional

docs/multisite.md
brooke-heaton: PR: https://github.com/acquia/blt/pull/3812

PR #3808, opened 2019-08-13
Remove Drupal Spec Tool version constraint

TravisCarden:
@danepowell any chance we could remove the version constraint on the Drupal Spec Tool altogether? It's preventing ORCA from requiring the latest dev version under certain circumstances.

### DIFF
--- a/subtree-splits/blt-project/composer.json
+++ b/subtree-splits/blt-project/composer.json
@@ -6,7 +6,7 @@
     "require": {
         "php": ">=7.1",
         "acquia/blt": "10.x-dev",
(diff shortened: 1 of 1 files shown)


danepowell:
It seems like the only new behavior here for end users is that they would start getting dev versions of the spec tool.

Do you want ordinary users pulling dev versions by default? That seems risky to me, but I don't know how you manage your dev workflow / branches / releases.

danepowell:
Oh actually I think the default minimum stability for downstream projects is `stable`, so they wouldn't get a dev version unless they've modified their minimum stability. So I'm fine with this.

TravisCarden:
Thank you. 🙂

PR #3820, opened 2019-08-20
Enabled coveralls.

danepowell:
WIP


danepowell:
Looks like at a minimum, this might require enabling xdebug, which I don't really want to figure out right now. It might be okay during PHPUnit tests but might also kill performance for Behat testing and anything invoking BLT commands directly.

PR #3821, opened 2019-08-21
Check if site can bootstrap before running updb in drupal:update

### LINKED ISSUE #3277: DT-595: When provisioning a new multisite, Acquia code hooks will always fail
Adding a new multisite involves the following:

* Add the site settings folder to dev repo and commit
* Deploy build artifact to host (Acquia)
* Acquia's post code deploy hooks run as deploy artifact is deployed, which check if config needs to be imported or db updates be run for each multisite
* Command fails when it reaches the new multisite, because the site hasn't yet been installed, so drush cache rebuild errors out indicating a DB table is missing
* Acquia reports task as failed

Any workaround for this? I think we need a way for BLT to understand that a site's may not be installed yet (database has no tables) and to skip it when running drupal:config:import?

richardbporter:
Fixes #3277 
--------

Changes proposed
---------
This change checks to see if the site can bootstrap via `drush status` (similarly to https://github.com/acquia/blt/blob/10.x/src/Robo/Commands/Doctor/DrupalCheck.php#L51) before running `drush updb` in the `drupal:update` BLT command. If the site cannot bootstrap, it logs a warning but continues to attempt database updates on other multisites.

Steps to replicate the issue
----------
The steps originally outlined in #3277 suffice but locally you can replicate this by:
- Creating two multisites locally
  - a.com
  - b.com
- Installing Drupal in b.com
- Run `blt auda`

----------
The previous behavior would cause the `blt auda` command to fail on a.com. This actually has potentially serious consequences like leaving b.com or any number of other multisites in a wonky state during AC deployments.

-----------
The `blt auda` command logs a warning, skips running database updates on a.com and proceeds to b.com.

Additional details
-----------
Maybe this should be an opt-in feature via configuration? 

Any feedback would be greatly appreciated. Thanks.

### DIFF
--- a/src/Robo/Commands/Setup/ConfigCommand.php
+++ b/src/Robo/Commands/Setup/ConfigCommand.php
@@ -34,7 +34,8 @@ class ConfigCommand extends BltTasks {
 
     if (empty($status->bootstrap) || $status->bootstrap != 'Successful') {
       $this->logger->warning('Site not bootstrapped. Cannot update database.');
-    } else {
+    }
+    else {
       $task = $this->taskDrush()
         ->stopOnFail()
         // Execute db updates.
--- a/src/Robo/Commands/Setup/ConfigCommand.php
+++ b/src/Robo/Commands/Setup/ConfigCommand.php
@@ -27,7 +27,7 @@ class ConfigCommand extends BltTasks {
       ->drush('status')
       ->option('fields', 'bootstrap')
       ->option('format', 'json')
-      ->silent(true)
+      ->silent(TRUE)
       ->run();
 
     $status = json_decode($result->getMessage());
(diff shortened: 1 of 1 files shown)


shelane:
We would like to see this as well, but... can there be an option to run the install on sites that don’t have a successful bootstrap? (New multisite and database added to code base)

danepowell:
I confirmed that `Inspector::isDrupalInstalled()` is _not_ site aware, because it uses a raw executor to run Drush rather than the site-aware DrushTask. So I think the path forward is as follows:
1. Modify `Inspector::isDrupalInstalled()` to either use the configured Drush URI similar to how `Inspector::isActiveConfigIdentical()` does, or use `$this->taskDrush` instead of the executor to run Drush. The latter requires adding the LoadTasks trait to the Inspector: `use LoadTasks`.
1. Modify `DeployCommand::updateSite()` to run `$this->getInspector()->isDrupalInstalled()` prior to running updates, and just skip the site / print a warning if the site is not installed.

It seems like you've got the chops to do this, but if you want help or don't have the bandwidth let me know.

richardbporter:
I'll give it a shot. Do you prefer I reset this branch, revert the previous commit or just close in favor of another PR?

richardbporter:
Closing in favor of #3831

PR #3826, opened 2019-08-23
Update adding-to-project.md

danepowell:
We want to make it easier to retrofit BLT into existing projects. First step is better documentation of the process.

### DIFF
--- a/docs/adding-to-project.md
+++ b/docs/adding-to-project.md
@@ -1,9 +1,8 @@
 # Adding BLT to an existing project
 
-Adding BLT to an existing project can be much more complex than simply creating a new site with BLT. There are a few reasons for this:
+Adding BLT to an existing project can be more complex than simply creating a new site with BLT since BLT includes a project template and expects the files in your project to match the structure and contents of this template.
 
-1. BLT expects your project to have a particular directory structure and set of files.
-1. Your existing dependencies may conflict with BLT's dependencies.
+BLT will attempt to generate any missing files when it is installed, but it can't account for every possible project structure and some manual reconciliation may be required. Adhering closely to the prerequisites and instructions below should guarantee a good outcome.
 
 Prerequisites:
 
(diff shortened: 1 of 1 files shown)


docs/adding-to-project.md
```diff
@@ -1,9 +1,8 @@
 # Adding BLT to an existing project
 
-Adding BLT to an existing project can be much more complex than simply creating a new site with BLT. There are a few reasons for this:
+Adding BLT to an existing project can be more complex than simply creating a new site with BLT since BLT includes a project template and expects the files in your project to match the structure and contents of this template.
 
-1. BLT expects your project to have a particular directory structure and set of files.
-1. Your existing dependencies may conflict with BLT's dependencies.
```
danepowell: I removed the note about dependency conflicts because this is true for any Composer package and not unique to BLT. Plus, BLT has many fewer dependencies in the 10.x branch than it used to.

docs/adding-to-project.md
```diff
@@ -21,8 +20,17 @@ To add BLT to a pre-existing Drupal project, do the following:
 
 1. Add BLT via composer:
 
-        rm -rf vendor composer.lock
```
danepowell: This was necessary when BLT was using composer merge plugin the 9.x branch, but no longer.

docs/adding-to-project.md
```diff
@@ -21,8 +20,17 @@ To add BLT to a pre-existing Drupal project, do the following:
...
-        composer require acquia/blt:^9.0.0 --no-update
+        composer require acquia/blt:^10.0 --no-update
         composer update
 
+1. Replace your `.gitignore` file with BLT's template `.gitignore`:
```
alexxed: Can't we say merge the files ? for both .gitignore and composer.json.

docs/adding-to-project.md
```diff
@@ -1,9 +1,8 @@
 # Adding BLT to an existing project
```
alexxed: I am missing a paragraph on what benefits do I have if I add BLT to an existing project.

PR #3836, opened 2019-09-03
DT-743: Fixes #3835: Workaround dependency conflicts

### LINKED ISSUE #3835: DT-743: Project creation, updates fail due to conflicting dependencies
This is not a BLT bug per se, but an upstream problem with Drupal, webflo/drupal-core-require-dev, and Mink:
- https://github.com/minkphp/MinkSelenium2Driver/issues/313
- https://www.drupal.org/project/drupal/issues/3078671

This makes it impossible to create new BLT projects, or to run composer install / update in some circumstances. Errors like this:

```
Problem 1
    - Installation request for webflo/drupal-core-require-dev 8.7.6 -> satisfiable by webflo/drupal-core-require-dev[8.7.6].
    - webflo/drupal-core-require-dev 8.7.6 requires behat/mink-selenium2-driver 1.3.x-dev -> no matching package found.
```

The best workaround I see is to add a branch alias as described here: https://github.com/minkphp/MinkSelenium2Driver/issues/313#issuecomment-527165233

I think end users should do this manually for now while we wait to see how the upstream issues are or aren't resolved, and then BLT can do it via an update hook if necessary.

Long term, we should consider removing webflo to decrease the risk of issues like this: https://github.com/acquia/blt/issues/3838

danepowell:
Fixes #3835 via https://github.com/minkphp/MinkSelenium2Driver/issues/313#issuecomment-527165233

The 1.3.x-dev branch of Mink, which is required by webflo/drupal-core-require-dev, no longer exists and is unlikely to come back. This adds an alias so that users can continue to download an equivalent commit hash of Mink.

### DIFF
--- a/src/Update/Updates.php
+++ b/src/Update/Updates.php
@@ -848,4 +848,24 @@ class Updates {
     }
   }
 
+  /**
+   * Version 10.4.0.
+   *
+   * @Update(
+   *   version = "10004001",
+   *   description = "Work around missing Mink branch."
+   * )
+   */
+  public function update_10004001() {
+    $composer_json = $this->updater->getComposerJson();
+    if (array_key_exists('acquia/blt-require-dev', $composer_json['require-dev'])) {
+      $composer_json['require-dev']['behat/mink-selenium2-driver'] = 'dev-master#a38512485e905738321f793b3acd3e7411a85101 as 1.3.x-dev';
+      $this->updater->writeComposerJson($composer_json);
+      $this->updater->getOutput()->writeln("Your composer.json has been updated to include a workaround for an upstream Mink bug. More details are available at https://github.com/acquia/blt/issues/3835");
+      $this->updater->getOutput()->writeln("");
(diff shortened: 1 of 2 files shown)


danepowell:
This works to fix project creation, but we need to figure out how to deal with existing BLT users who will hit this problem next time they run composer update.

The best idea I have at this point is to add an update hook that adds this alias to composer.json if the project is already using `acquia/blt-require-dev` (and hence `webflo/drupal-core-require-dev`).

Then we need a new hotfix or minor release (minor was scheduled for tomorrow anyway).

danepowell:
I think the tests on this are going to continue to fail until we cut a new release. I don't know what else we can do to solve this. Mink really hosed us on this one by deleting such a widely-used release.

danepowell:
I've fixed new project creation via a hotfix. Existing users may still have trouble running composer updates. Before merging this, which is a pretty kludgy workaround and requires manual updates by end users, I'd like to give the upstream issues some time to resolve themselves.

PR #3837, opened 2019-09-03
DT-743: Fixes #3835: Cannot create new BLT projects

### LINKED ISSUE #3835: DT-743: Project creation, updates fail due to conflicting dependencies

danepowell:


danepowell:
Only works in root package: https://getcomposer.org/doc/articles/aliases.md#require-inline-alias

PR #3853, opened 2019-09-18
DT-586: Fixes #3616: Support syncing of private files.

### LINKED ISSUE #3616: DT-586: Support syncing private files from remote
A remote site uses private files. I want to sync these private files locally. The `blt drupal:sync:files` command does not sync private files, only public.

A config option that includes private files when running `blt drupal:sync:files`, and/or an additional command like `blt drupal:sync:files:private`

danepowell:
Fixes #3616 
--------

Changes proposed
---------
- Add `drupal:sync:private-files` command
- Add `sync.private-files` config option
- Modify `blt sync` to call `drupal:sync:private-files` if `sync.private-files` is TRUE.

Steps to replicate the issue
----------
1. Have a BLT project with a remote alias (cloud environment) defined, which has private files
1. Run `blt sync -Dsync.private-files=true` and verify that private files are synced to `<repo-root>/files-private`.

### DIFF
--- a/src/Robo/Commands/Sync/SyncCommand.php
+++ b/src/Robo/Commands/Sync/SyncCommand.php
@@ -37,17 +37,23 @@ class SyncCommand extends BltTasks {
    * Copies remote db to local db, re-imports config, and executes db updates
    * for each multisite.
    *
+   * @param array $options
+   *   Array of CLI options.
+   *
    * @command drupal:sync:default:site
    * @aliases ds drupal:sync drupal:sync:default sync sync:refresh
    * @executeInVm
    */
-  public function sync($options = [
+  public function sync(array $options = [
     'sync-files' => FALSE,
+    'sync-private-files' => FALSE,
   ]) {
-
     $commands = $this->getConfigValue('sync.commands');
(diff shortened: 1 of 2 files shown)


src/Robo/Commands/Sync/SyncCommand.php
```diff
@@ -37,18 +37,24 @@ public function allSites() {
...
+   *   Array of CLI options.
+   *
    * @command drupal:sync:default:site
    * @aliases ds drupal:sync drupal:sync:default sync sync:refresh
    * @executeInVm
    */
-  public function sync($options = [
+  public function sync(array $options = [
     'sync-files' => FALSE,
+    'sync-private-files' => FALSE,
```
alexxed: Is this a public vs private? Can't tell now. If so it would be good to have at least an alias sync:public-files for sync:files.

danepowell: Good point, I'll add an alias and we can make the alias canonical in BLT 11.x

danepowell: https://github.com/acquia/blt/issues/3860

PR #3857, opened 2019-09-19
DT-598: Fixes #2722: ci.blt.yml not consistently applied.

### LINKED ISSUE #2722: DT-598: Yaml placeholders are not correctly expanded in runtime parameters
My system information:
* BLT version: 9.0.5

I've defined the following command hook:
```yml
command-hooks:
 post-deploy-build:
   dir: '${deploy.dir}'
   command: 'rm -rf vendor/drush'
```

BLT builds artifacts on Pipelines by calling this:
```bash
export NEW_BUILD_DIR=/tmp/artifact
blt artifact:build -D deploy.dir=${NEW_BUILD_DIR} --ansi --verbose
```

Thus, I would expect the `rm` command I've defined to run in `/tmp/artifact`. Instead, it actually runs in the Pipelines source directory. This is bad, because the changes to that directory don't get preserved. So as near as I can tell, the post-deploy-build hook is useless.

As a workaround, I've hardcoded the directory to be `/tmp/artifact` but that's obviously not sustainable.

danepowell:
Fixes #2722 
--------

Changes proposed
---------
- Add documentation about how yml properties are expanded
- Fix ci environment not being detected when building artifact on Pipelines

Steps to replicate the issue
----------
1. Add a `post-deploy-build` command to `ci.blt.yml`, i.e.:
```
command-hooks:
  post-deploy-build:
    dir: /tmp/artifact
    command: 'npm install'
```
2. Commit and start a pipelines job.

----------
The post-deploy-build command is not run

-----------
The post-deploy-build command is run

### DIFF
--- a/scripts/pipelines/build_artifact
+++ b/scripts/pipelines/build_artifact
@@ -7,7 +7,7 @@ export PATH=${COMPOSER_BIN}:${PATH}
 # Generate artifact in a separate directory.
 export NEW_BUILD_DIR=/tmp/artifact
 blt deploy:check-dirty --ansi --verbose
-blt artifact:build -D deploy.dir=${NEW_BUILD_DIR} --ansi --verbose
+blt artifact:build -D deploy.dir=${NEW_BUILD_DIR} --environment=ci --ansi --verbose
 # Move git history to artifact directory. Required for pipelines to commit and push.
 mv ${SOURCE_DIR}/.git ${NEW_BUILD_DIR}/
 # Allow dotfiles (hidden) to be globbed (via *).
--- a/subtree-splits/blt-project/blt/blt.yml
+++ b/subtree-splits/blt-project/blt/blt.yml
@@ -1,6 +1,14 @@
 # This file contains your BLT configuration. For a list of all available
 # properties with current values run `blt config:dump`. Default values come
 # from vendor/acquia/blt/config/build.yml.
+#
+# These values can be overridden at the command line using `--define`, i.e.:
+# blt setup --define project.profile.name=minimal
+#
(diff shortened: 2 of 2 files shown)


scripts/pipelines/build_artifact
```diff
@@ -7,7 +7,7 @@ export PATH=${COMPOSER_BIN}:${PATH}
 # Generate artifact in a separate directory.
 export NEW_BUILD_DIR=/tmp/artifact
 blt deploy:check-dirty --ansi --verbose
-blt artifact:build -D deploy.dir=${NEW_BUILD_DIR} --ansi --verbose
+blt artifact:build -D deploy.dir=${NEW_BUILD_DIR} --environment=ci --ansi --verbose
```
danepowell: This is the second or third bug we've caught where this parameter was missing.

There are ways we could refactor this to avoid passing `--environment=ci` every time, such as setting `BLT_ENV=ci` as an environment variable once at the start of the script, or by incorporating `EnvironmentDetector::isCiEnv()` into the DefaultConfig environment detection step. They aren't without risk though, and at any rate are beyond the scope of this PR.

alexxed: CI=true is present in most CI environments today. If not (looking at Jenkins, people can set it).

Why does blt have to act differently in CI vs on your machine ?

danepowell: Good point about the CI variable, we should key off of that. I created this followup issue: https://github.com/acquia/blt/issues/3859

This ticket is itself a good example of why BLT needs to behave differently. Systems paths are difference, available resources are different, you need different modules enabled, etc... In this particular case, the deploy directory is pretty special in Pipelines.

PR #3862, opened 2019-09-27
DT-580: Use standardized class for modifying composer.json files.

### LINKED ISSUE #3683: DT-580: Refactor acsf:init command
After #3682 is merged, add a unified composer.json reader/writer class, use it instead of existing ad-hoc calls that modify composer.json, and also use it instead of multiple composer requires in the acsf:init command (which should save a minute or two off the command and automated tests).

danepowell:
Fixes #3683 
--------

Changes proposed
---------
- Use a single standard class for reading and writing composer.json files, to ensure they are always formatted consistently. Should have no functional change.
- Refactor related methods and doc comments to ensure tests pass.


src/Robo/Commands/Doctor/ComposerCheck.php
```diff
@@ -45,46 +57,21 @@ public function __construct(Config $config, Inspector $inspector, Executor $exec
...
   protected function setComposerJson() {
-    if (file_exists($this->getConfigValue('repo.root') . '/composer.json')) {
-      $composer_json = json_decode(file_get_contents($this->getConfigValue('repo.root') . '/composer.json'), TRUE);
-      $this->composerJson = $composer_json;
-
-      return $composer_json;
-    }
-
-    return [];
+    $composerJson = new ComposerJson($this->getConfigValue('repo.root'));
```
danepowell: This class never writes back to composer.json, so we don't need to store the object itself, just the file contents.

src/Robo/Common/ComposerJson.php
```diff
@@ -0,0 +1,64 @@
...
+    if (!file_exists($this->filepath)) {
+      throw new BltException("Could not find composer.json");
+    }
+    $this->contents = json_decode(file_get_contents($this->filepath), TRUE);
+  }
+
+  /**
+   * Writes contents to file.
+   */
+  public function write() {
```
adrianthedev: Do we use the `read()`/`write()` mnemonic in `blt`? 
An alternative could be `load()`/`save()` as we have it in `ads`.

Just a nitpick.

src/Robo/Commands/Doctor/ComposerCheck.php
```diff
@@ -104,16 +91,6 @@ protected function setComposerLock() {
     return [];
   }
 
-  /**
-   * Get composer.lock.
-   *
-   * @return array
-   *   Array.
-   */
-  public function getComposerLock() {
```
adrianthedev: I see we're not using this anywhere in our codebase. Could someone else use it from an external source?

src/Robo/Commands/Saml/SimpleSamlPhpCommand.php
```diff
@@ -254,18 +255,20 @@ protected function outputCompleteSetupInstructions() {
...
       ->run();
-    $composer_json = json_decode(file_get_contents($this->getConfigValue('repo.root') . '/composer.json'));
-    $composer_json->scripts->{"post-drupal-scaffold-cmd"}[] = "cd docroot && patch -p1 <../patches/htaccess-saml.patch";
-    file_put_contents($this->getConfigValue('repo.root') . '/composer.json',
-      json_encode($composer_json, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES));
+    $composerJson = new ComposerJson($this->getConfigValue('repo.root'));
+    $composerJson->contents['scripts']['post-drupal-scaffold-cmd'][] = 'cd docroot && patch -p1 <../patches/htaccess-saml.patch';
+    $composerJson->write();
     $this->taskExecStack()
-      ->dir($this->getConfigValue('repo.roou'))
```
adrianthedev: Good catch :)

src/Robo/Common/ComposerJson.php
alexxed: There's some logic here that deserves a unit test to make sure we understand requirements in the future.

danepowell: We use write() in the YamlWriter, but I only added that pretty recently. Robo uses load() a few places. I don't see save() anywhere.

I have no strong preference, I'll switch to save() to be consistent with Robo and ADS.

src/Robo/Commands/Doctor/ComposerCheck.php
danepowell: Yeaahhhh.... I guess it's _theoretically_ possible. Just to be safe, I'll put it back.

danepowell:
Doh... I realized Composer already provides a `JsonFile` class that we can use to read/write with all of the same business logic. No need to roll our own class or tests for this. I switched to that.

danepowell:
Upstream issue: https://github.com/acquia/lightning_api/pull/71

If we can sort this, remove ./bin/travis/orca-fix.sh hack

danepowell:
Closing for now since I've spent as much time as I think is warranted trying to debug this, and haven't really made progress. It'll resolve itself naturally with BLT 11, which will require Drupal 8.7/8.8 which have JSON API in core, at which point we can consider trying this again.

PR #3866, opened 2019-10-01
Issue #3865: Move location of logging UI override to appropriate file.

### LINKED ISSUE #3865: Move system logging config override out of cache.settings.php
BLT's [cache.settings.php](https://github.com/acquia/blt/blob/10.x/settings/cache.settings.php) includes the following:

```
if (EnvironmentDetector::isProdEnv() || EnvironmentDetector::isStageEnv()) {
  $config['system.logging']['error_level'] = 'hide';
}
```

This is a strange place for this config override. It has nothing to do with caching.

Move this block of code out of cache.settings.php and into [logging.settings.php](https://github.com/acquia/blt/blob/10.x/settings/logging.settings.php).

**Additional context**
A few times now I've tried to find where this config override is happening from BLT and cache.settings.php is one of the last places I looked.

bkosborne:
Fixes #3865

### DIFF
--- a/settings/cache.settings.php
+++ b/settings/cache.settings.php
@@ -5,7 +5,6 @@
  * Contains caching configuration.
  */
 
-
 /**
  * Use memcache as cache backend if Acquia configuration is present.
  */
--- a/settings/cache.settings.php
+++ b/settings/cache.settings.php
@@ -8,16 +8,6 @@
 use Acquia\Blt\Robo\Common\EnvironmentDetector;
 
 /**
- * Settings.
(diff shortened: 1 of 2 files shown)


settings/cache.settings.php
```diff
@@ -7,16 +7,6 @@
 
 use Acquia\Blt\Robo\Common\EnvironmentDetector;
```
danepowell:
```suggestion
```

This use statement is no longer required, and causing tests to fail.

PR #3867, opened 2019-10-01
testing lightning api errors

danepowell:


danepowell:
if we can sort this, remove ./bin/travis/orca-fix.sh hack

danepowell:
Better demonstrated upstream: https://github.com/acquia/lightning_api/pull/71

PR #3875, opened 2019-10-04
Correct extra.patches data type in blt-project/composer.json

TravisCarden:
The `extra.patches` element in `composer.json` is supposed to be an object, not an array. I doubt anyone will make me defend a fix to something that technically wrong, but I'll just mention without going into details that this makes it much more difficult than it needs to be to safely modify the value from the command-line (i.e., to insert a member rather than clobbering the whole value).

### DIFF
--- a/subtree-splits/blt-project/composer.json
+++ b/subtree-splits/blt-project/composer.json
@@ -76,7 +76,7 @@
         "patchLevel": {
             "drupal/core": "-p2"
         },
-        "patches": []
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks, makes sense. I think we originally put that key in composer.json to work around this issue: https://github.com/cweagans/composer-patches/issues/16

It appears though that patches are being properly applied in automated tests, so this change should be safe.

PR #3877, opened 2019-10-04
DT-458: Removed drush alias generation.

danepowell:
Changes proposed
---------
- Remove Drush alias generation for Acquia Cloud sites, which was deprecated in 10.x and replaced by ADS.
- Remove dependency on the typhonius package that was only being used for alias generation.

Tests will be fixed in #3879


mpodorov:
What is the reason of the alias generation deprecation?

danepowell:
It duplicates functionality that now exists in a superior form in [Acquia Developer Studio](https://docs.acquia.com/dev-studio), as mentioned in the 10.x deprecation message: https://github.com/acquia/blt/pull/3876

If you haven't tried Developer Studio, you really should. If you like BLT, I think you'll like ADS :smile: 

More generally, it's an outlier in terms of BLT functionality in that it is highly specific to Acquia Cloud. We try to make BLT open to any hosting platform, it shouldn't be coupled too tightly to Acquia Cloud.

It also depends on a third-party SDK (typhonius/acquia-php-sdk-v2) that is not maintained by Acquia. Moving this functionality to Developer Studio makes it more stable and secure.

mpodorov:
I'd say some words should be added to BLT documentation about this replacement, such as
"instead of blt aliases command, use use such and such commands of ads".

danepowell:
Yes, we added a deprecation warning to BLT 10.x to that effect: https://github.com/acquia/blt/pull/3876

shelane:
Or for those who are not going to be using ADS because of already having a local dev environment we like, point to where else we can generate alias files. Is there an equal mechanism available directly on ACE?

danepowell:
ADS does more than manage local dev environments, such as providing a unified API interface via the CLI and desktop. If you prefer not to use a CLI tool, you can download aliases via your Cloud profile: https://docs.acquia.com/acquia-cloud/manage/ssh/drush/aliases/

lcatlett:
@shelane I made a BLT plugin which adds this functionality back because I am also using a different local dev stack instead of ADS: https://github.com/lcatlett/blt-drush-alias

typhonius:
Stumbled upon this issue and wanted to confirm that I'll continue to maintain https://github.com/typhonius/acquia-php-sdk-v2/ and the Acquia Cli library for people who prefer to use it e.g. as part of wider PHP libraries.

luzzardi:
Hey!
Why do you remove that so early?
ADS seems to be not ready for ACSF and we miss this command.

If someone else need the alias for acsf: I recommend use the of @lcatlett  plugin:
https://github.com/lcatlett/blt-drush-alias
Thanks a lot!

PR #3878, opened 2019-10-04
DT-655: Removed selenium-server-standalone from blt-require-dev.

### LINKED ISSUE #3779: Remove selenium-server-standalone from blt-require-dev
Currently the blt-require-dev subpackage has a dependency on selenium-server-standalone. This server binary is 20MB+ in size (and worse, uncompressable), taking up a significant amount of disk space and bandwidth, and especially bloating ADS's tarball.

My guess is that most people never use the standalone selenium server since BLT now defaults to headless chrome for behat tests, and DrupalVM includes a global selenium server. Anyone who still needs it can simply add it back as a composer dependency, we would just need to document this.

Since I don't have hard numbers on Selenium usage, I wanted to float this issue for feedback from the community.

danepowell:
Fixes #3779  
--------

Changes proposed
---------
- Remove selenium-server-standalone for the reasons articulated in #3779

This requires anyone wishing to use Selenium for Behat or Drupal tests to install it for their project via `composer require se/selenium-server-standalone`.


danepowell:
Interesting, something seems to be depending on this and breaking tests...

PR #3882, opened 2019-10-08
Update ORCA version

TravisCarden:
ORCA is quite stable at this point. All other packages have been tracking the `master` branch for several releases without incident.


danepowell:
@TravisCarden it looks like the newer release of ORCA (or presumably, the coding standards dependency) is sniffing markdown files as if they were php. Is this a known issue?

It looks like we also need to update our travis.yml and fix some other coding standards, I'll take care of these last two.

danepowell:
Argh... this also needs to go against 11.x, and Github screwed up the rebase, so I opened a new PR: #3884

PR #3884, opened 2019-10-08
Update ORCA version

danepowell:
Changes proposed
---------
- Update ORCA to master
- Fix code style issues due to new coding standards
- Add caching dir to travis.yml according to ORCA release notes: https://github.com/acquia/orca/releases/tag/v1.1.0

Should have no functional impact, this is just improving ORCA integration and a follow up to Travis' PR here: #3882


danepowell:
@TravisCarden we need to figure out why ORCA is scanning markdown files as if they were PHP, I assume this is due to a recent change in ORCA or coding standards?

TravisCarden:
Yes, @danepowell. Acquia Coding Standards recently fixed an issue in https://github.com/acquia/coding-standards-php/issues/1 where a bug in PHPCS was preventing Coder Sniffer from checking various file types it _means to_, including Markdown. This is a case of "fixing ours broke yours". :)

But looking at your actual failures, you have a few "Expected 1 newline at end of file" that are valid. The only problem is `docs/configuration-management.md`, and I think Coder Sniffer is incorrectly trying to check it because it's finding a PHP open tag (`<?php`) in it. I believe that's a bug, and I've reported it at [Sniffer incorrectly applies PHP rules to non-PHP files if they contain PHP tags [#3086639]](https://www.drupal.org/project/coder/issues/3086639). I've confirmed that you can get around it by using [code blocks with syntax highlighting](https://help.github.com/en/articles/creating-and-highlighting-code-blocks#syntax-highlighting), like this:

```markdown
    ```php
    $available_modules = system_rebuild_module_data();
    $dependencies = array();
    foreach ($available_modules as $name => $module) {
      if ($module->info['package'] == 'My Bundle') {
        $dependencies[] = $name;
      }
      \Drupal::service('module_installer')->install($dependencies);
    }
    ```
```

danepowell:
Thanks a lot for the investigation and help, it seems to be working now.

PR #3885, opened 2019-10-11
Show diff for the dirty build.

### LINKED ISSUE #3564: DT-587: Changes to checkDirty behavior during deployments
BLT has long had a function checkDirty that fails deploys if BLT detects that files have changed during the deploy process itself, which could indicate a variety of problems such as bad Composer files, NPM files, problems with gitignore, and/or more generally conditions that might cause the site to behave differently in a remote environment than it did during tests (with undefined and potentially very bad consequences).

In 9.2.x and 10.x we [recently enabled this feature on Pipelines](https://github.com/acquia/blt/pull/3472) (it's always been enabled for Travis and other environments that directly call `blt deploy`. Pipelines is a special case because it doesn't call `deploy`).

It turns out a _lot_ of Pipelines users have those kinds of problems with their codebase, as evidenced by the spate of issues opened after this feature was added. While it's good that these problems are being caught and corrected, it's not great that it started happening unexpectedly, or that it happens during the deploy step (it's really bad for things to fail during a deploy if they didn't fail during tests).

- [x] Add a more helpful error message to both 9.2.x and 10.x branches indicating corrective action if checkDirty fails.
- [ ] See if there's some way to run or simulate this check earlier than the deploy step, and ideally fail during tests.

This last point is tricky because a lot of things (such as the gitignore file, and composer production dependencies) are different during the deploy process, so we can't just run the same command earlier on. I see two options:
- Based on feedback from users who have had problems with their codebase, see if there are additional specific validation steps that we can perform, such as checking for whether settings files are valid.
- Maybe we could _simulate_ a deploy as part of the test process, by running a deploy with the dry-run option? We'd need to make this work on Pipelines as well as Travis (not easy, since they have very different deploy processes), and only in cases when a deploy would otherwise be called for.

andrewhorokhovets:
Fixes #3564
--------

Changes proposed
---------
Print diff information for the dirty build.

Steps to replicate the issue
----------
1. Run dirty build

----------
BLT showed only file was changed and not what actually changed in those files.
It can be difficult to debug what gone un.
Like this https://prnt.sc/phxpjb

-----------
BLT will show changes, like this https://prnt.sc/phxmyz

Additional details
-----------
Those changes can save a lot of time, which needed to resolved unclear bugs.


src/Robo/Commands/Deploy/DeployCommand.php
```diff
@@ -162,6 +162,13 @@ public function checkDirty(array $options = ['ignore-dirty' => FALSE]) {
         $this->logger->warning("There are uncommitted changes on the source repository.");
       }
       else {
+        if ($options['verbose']) {
+          $this->taskExec('git diff --exit-code')
```
alexxed: I'd say `git status` is less verbose. In this case you probably care most about what files are changed, not the diff.

andrewhorokhovets: Hi @alexxed !
Thank you  for your review.
Yes `git status` is less verbose. 
Therefore, it is not always clear what exactly has changed. I gave an example with settings.php
There are several behaviors that can lead to changes in the build.
- If a Drupal installer is running, a line with the type of the installed profile may be added.
- File permissions may change.
- A line with blt is added to the end of the file

Not all such changes are understandable when viewing git status.
In addition, when an error is found, more information gives more power to solve it.

alexxed: Looking at the screenshot I see it's a CI, so it makes sense to get as much info as possible automatically.

andrewhorokhovets: @alexxed great, thanks!
What is the next step to processed it?

PR #3892, opened 2019-10-17
TwigCommand should use configured project docroot.

### LINKED ISSUE #3863: Allow Drupal root to be configurable
The problem is that BLT assumes all project's docroot to be `docroot`. However, there are other standards that use `web` as the Drupal root.

The docroot should be configurable using the `blt.yml` file. Say, support for something as follows should suffice:
```yaml
project.docroot: web
```

I've found no alternatives – I cannot rename the docroot to `docroot` because it's not my project. The team already has their norms and proposing such a big change will make the team not want to use BLT.

wu-edward:
Fixes #3863
--------

Changes proposed
---------

- Use BLT configuration for Drupal docroot instead of hardcoded `docroot` relative to repo root during Twig linting
-

Steps to replicate the issue
----------
1. 
2. 

----------
> tests:twig:lint:files
Linting twig files...
PHP Warning:  file_get_contents(/path/to/project/docroot/core/lib/Drupal/Core/Template

-----------
No path errors during Twig lint command.

Additional details
-----------

### DIFF
--- a/src/Robo/Commands/Validate/TwigCommand.php
+++ b/src/Robo/Commands/Validate/TwigCommand.php
@@ -107,8 +107,8 @@ class TwigCommand extends BltTasks {
   protected function createTwigLintCommand() {
     $twig = new Environment(new FilesystemLoader());
 
-    $repo_root = $this->getConfigValue('repo.root');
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks for the fix!

Issue #3828, opened 2019-08-27
DT-728: Update docs for alternative environments


lcatlett:
**Is your feature request related to a problem? Please describe.**
BLT 10 plugins now exist (see https://github.com/docksal/blt-docksal and https://github.com/lcatlett/blt-ddev) which automate and supersede the current manual setup instructions for both [ddev](https://github.com/acquia/blt/blob/10.x/docs/alternative-environment-tips/ddev.md) and [docksal](https://github.com/acquia/blt/blob/10.x/docs/alternative-environment-tips/docksal.md) so the BLT alternative environment docs should be updated as such. Given https://github.com/acquia/blt/issues/3825 it may make sense to replace these docs entirely in favor of creating a BLT plugins directory page

danepowell:
That's awesome, I didn't even realize those plugins existed. I definitely agree, let's pull those out of the core BLT docs, and add a plugin directory.

PR #3894, opened 2019-10-22
Adds documentation for the Docksal environment

shelane:
Referencing #3828 

Although there is a plugin now listed on the plugins documentation page, I still believe that there needs to be "some" documentation of Docksal as an alternative environment. Included in this PR is documentation on the two methods to setting up Docksal with BLT. One uses the plugin for existing systems, one uses Docksal to create the BLT site - pre-configured for running BLT and Docksal.

Removing it altogether makes it look like Docksal isn't a candidate for an alternative development environment.

### DIFF
--- a/docs/alternative-environment-tips/docksal.md
+++ b/docs/alternative-environment-tips/docksal.md
@@ -1,0 +1,15 @@
+# Setting up BLT with Docksal
+
+By default, BLT is set to work with Drupal VM. It is easy to set up to use with Docksal, and there are two easy ways
+to set this up.
+
+## Setting Your Existing Project to Use Docksal
+
+You can install the [Docksal BLT plugin](../plugins) and run the commands to setup your BLT site to use Docksal.
+
+## Creating a New BLT Project Using Docksal
+
+If you have [Docksal](https://docksal.io) installed and want to create a new BLT project, use `fin project create`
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks a lot for the feedback. I appreciate the effort you've put into the documentation, however the problem is that Docksal integration is not maintained by Acquia, and thus we can't maintain documentation around it either, except to reference community efforts such as the Docksal plugin.

I'd propose that you instead try to improve the documentation included with the [Docksal plugin](https://github.com/docksal/blt-docksal) itself, or create a blog post / tutorial somewhere that we can link to from the BLT docs.

Then we can clean up the bottom section of the "local development" page with better links to resources like this: https://github.com/acquia/blt/blob/11.x/docs/local-development.md#alternative-local-development-environments

We just can't have specific instructions in the BLT docs that would require continuing maintenance and oversight.

shelane:
Thank you for your consideration. We have a blog post that will be updated with the upcoming 1.13.0 release of Docksal. Currently the post only describes the "manual" process of setting up BLT with Docksal. At the release of 1.13.0, it will be updated to include the information that I put in this PR. https://blog.docksal.io/docksal-and-acquia-blt-1552540a3b9f

danepowell:
Sounds good, I've created a ticket to update that documentation and include a reference to your blog post: #3904

PR #3901, opened 2019-10-30
DT-950: Update node version in Pipelines.

### LINKED ISSUE #3900: DT-950: Update node version in Pipelines
See https://github.com/acquia/blt/issues/3897

In Pipelines setup_env, replace the two node-related lines with `nvm install stable --latest-npm`

Install a pinned version of node, or force end users to download their own version.

**Additional context**
Maybe add a comment to the top of setup_env script to indicate that it's internal to BLT, and generally not expected for customers to override.

danepowell:
Fixes #3900 
--------

Changes proposed
---------
- Stop automatically updating Node in Pipelines. We already provide guidance on how end users can manage this themselves.
- Add notes clarifying intended usage of Pipelines scripts

Requires change record since users relying on BLT's node version might be surprised when it goes away.

### DIFF
--- a/scripts/pipelines/acquia-pipelines.yml
+++ b/scripts/pipelines/acquia-pipelines.yml
@@ -18,7 +18,7 @@ events:
               - composer validate --no-check-all --ansi
               - composer install --ansi
               - source ${BLT_DIR}/scripts/pipelines/setup_env
-              # - nvm install v6.9.1
+              # - nvm install v6.9.1 --latest-npm
               # - npm install --global yarn
         - validate:
             type: script
--- a/scripts/pipelines/setup_env
+++ b/scripts/pipelines/setup_env
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+# This file is managed by Composer and should not be modified by end users. To
+# override its default behavior, copy it into your project and modify
+# acquia-pipelines.yml to call your custom version.
 
 set -ev
 
@@ -12,7 +15,6 @@ git config --global user.email "noreply@acquia.com"
 mysql -u root -proot -e "CREATE DATABASE IF NOT EXISTS drupal"
 
 # Setup NVM.
-nvm install v8.1.4
-npm install -g npm
+nvm install stable --latest-npm
 
 set +v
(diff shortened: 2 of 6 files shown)


scripts/pipelines/setup_env
```diff
@@ -12,7 +15,6 @@ git config --global user.email "noreply@acquia.com"
 mysql -u root -proot -e "CREATE DATABASE IF NOT EXISTS drupal"
 
 # Setup NVM.
-nvm install v8.1.4
-npm install -g npm
+nvm install stable --latest-npm
```
adrianthedev: Is this normal? Just going to the latest instead of specifying which version?

danepowell: Yeah I don't know what's optimal here. If we pick a specific version, we have to continually update this file like we're doing now, deal with potentially insecure versions getting installed, and deal with people complaining that it's not the version _they_ happen to want that day.

If we don't pick a specific version, it'll change when a new release comes out which could surprise people and break builds. Although for minor/patch releases, that's not likely to happen often.

Do you have a strong opinion? I guess a compromise would be to at least pin the major version: `nvm install 12 --latest-npm`

danepowell: Actually, I propose that we just take this NPM out of BLT's internal scripts, and move it into our acquia-pipelines.yml template. That way people can manage it themselves, or get rid of it entirely if they don't use node. Nothing in BLT itself uses node, so there's no reason for BLT to manage this.

adrianthedev: That sound good! That's the place for it

scripts/pipelines/acquia-pipelines.yml
```diff
@@ -18,7 +18,7 @@ events:
               - composer validate --no-check-all --ansi
               - composer install --ansi
               - source ${BLT_DIR}/scripts/pipelines/setup_env
-              # - nvm install v6.9.1
+              # - nvm install v6.9.1 --latest-npm
```
alexxed: are you updating node version or npm version ?

`nvm install lts/*` installs the latest lts for example

alexxed: oh, it's a comment :D doh!

PR #3902, opened 2019-10-30
DT-593: Codesniffing via pre-commit hook doesn't respect filesets.

### LINKED ISSUE #3448: DT-593: Codesniffing via pre-commit hook does not respect filesets
This is the root cause of numerous other issues, most recently #3366 (which has my thoughts on the issue and some additional background), but also #2879 and others. We should fix it for real.

danepowell:
Fixes #3448 
--------

Changes proposed
---------
- For git pre-commit hook, stop sniffing an explicit list of modified files (which can result in files getting sniffed that otherwise wouldn't, according to filesets defined in phpcs.xml), and instead use PHPCS' built-in GitModified filter (which does respect phpcs.xml)
- Clarify comments / documentation about interaction between BLT's functions, phpcs.xml, and filesets.
- Fix the wrongly-named gitmodified filter
- Stop passing the current directory as an argument to phpcs, which will also cause it to include files not defined in phpcs.xml.

Steps to replicate the issue
----------
0. Ensure git hooks are installed in your project (you'll see validation running on commits).
1. Create a file with code style errors, such as `new-dir/test.php`.
2. Run `blt phpcs`, observe output.
3. `git add` the new file and attempt `git commit`.

----------
`blt phcs` passes without errors as expected, since `new-dir` is not configured to be sniffed via phpcs.xml. But the Git commit fails unexpectedly.

-----------
Both `blt phpcs` and `git commit` pass without errors. Other files with errors (in directories such as `docroot/modules/custom`) still cause `blt phpcs` and `git commit` to fail.

### DIFF
--- a/subtree-splits/blt-project/phpcs.xml.dist
+++ b/subtree-splits/blt-project/phpcs.xml.dist
@@ -34,9 +34,5 @@
   <exclude-pattern>*/behat</exclude-pattern>
   <exclude-pattern>*/node_modules</exclude-pattern>
   <exclude-pattern>*/vendor</exclude-pattern>
-  <!-- Exclude files provided by ACSF that fail code sniffing. -->
-  <exclude-pattern>factory-hooks</exclude-pattern>
-  <exclude-pattern>hooks</exclude-pattern>
-  <exclude-pattern>docroot/sites/g</exclude-pattern>
 
 </ruleset>
--- a/src/Robo/Commands/Git/GitCommand.php
+++ b/src/Robo/Commands/Git/GitCommand.php
@@ -62,7 +62,7 @@ class GitCommand extends BltTasks {
     $collection->addCode(
       function () use ($changed_files) {
         return $this->invokeCommands([
-          'tests:phpcs:sniff:files' => ['file_list' => $changed_files],
+          'tests:phpcs:sniff:modified',
           'tests:twig:lint:files' => ['file_list' => $changed_files],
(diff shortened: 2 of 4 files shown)


subtree-splits/blt-project/phpcs.xml.dist
```diff
@@ -34,9 +34,5 @@
   <exclude-pattern>*/behat</exclude-pattern>
   <exclude-pattern>*/node_modules</exclude-pattern>
   <exclude-pattern>*/vendor</exclude-pattern>
-  <!-- Exclude files provided by ACSF that fail code sniffing. -->
```
danepowell: These exclude patterns were only added a previous workaround to these issue. Note that directories such as `factory-hooks` are not included in the `file` directives above, and thus will continue to be excluded now that this bug is fixed.

YinLieN:
Hello,

Since this commit, I have an issue with the hook pre-commit, the phpcs sniff is not working at all.

I investigated and found that the new command `vendor/bin/phpcs --filter=GitModified` is applying only on unstaged files. But in order to commit I have to stage all my files and then the hook pre-commit is launched.. with 0 files everytime. I tried with `vendor/bin/phpcs --filter=GitStaged`, it was working.

Do I miss something or is it something that needs to be fixed on blt side?

danepowell:
@agiraud85 thanks for reporting, I think you are right, we'll track that in #3980

PR #3912, opened 2019-11-05
DT-658: Anonymous telemetry

danepowell:
Changes proposed
---------
- Add unified class for storing user configuration (by default at `~/.config/blt`)
- Add unified methods to the environment detector for getting machine info
- Add opt-in reporting for user-run blt commands, as well as more detailed reporting for two specific events (toggle modules and config import)

Steps to replicate the issue
----------
1. Run `blt` and opt-out of reporting.
2. Run any BLT command and verify it is not reported to Amplitude.
3. `rm -rf ~/.config/blt` to reset reporting preferences.
4. Run `blt` and opt-in to reporting.
5. Run any BLT command and verify that the command and exit code are reported to Amplitude.

Additional details
-----------
Follow up to work in #3660


danepowell:
Follow the ADS pattern of asking to opt in to telemetry, store preferences and user data in a user.json file in the home directory, then read from this when sending telemetry and set `setOptOut(true)` if the user has not opted in.

src/Robo/Blt.php
```diff
@@ -88,6 +93,18 @@ public function __construct(
...
+
+    $this->initializeAmplitude();
+  }
+
+  /**
+   * Initializes Amplitude.
+   */
+  private function initializeAmplitude() {
+    $amplitude = Amplitude::getInstance();
+    $amplitude->init('dfd3cba7fa72065cde9edc2ca22d0f37')
```
danepowell: @alexxed I don't seem to have sufficient access to Amplitude to view API keys, so I have no idea to which project this is reporting. Do you know which project is associated with this API key?

danepowell: @raghunat Alex said you might know where this API key came from, so I can check whether this is reporting correctly?

src/Robo/Common/EnvironmentDetector.php
```diff
@@ -253,6 +254,65 @@ public static function isAcsfInited() {
...
+      case FamilyName::LINUX:
+        return shell_exec('( cat /var/lib/dbus/machine-id /etc/machine-id 2> /dev/null || hostname ) | head -n 1 || :');
+
+      case FamilyName::DARWIN:
+        $output = shell_exec('ioreg -rd1 -c IOPlatformExpertDevice | grep IOPlatformUUID');
+        $parts = explode('=', str_replace('"', '', $output));
+        return strtolower(trim($parts[1]));
+
+      case FamilyName::WINDOWS:
+        return shell_exec('%windir%\\System32\\reg query "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Cryptography" /v MachineGuid');
```
alexxed: nice windows wizardry :)

src/Robo/Commands/Setup/ToggleModulesCommand.php
```diff
@@ -69,10 +75,15 @@ protected function doToggleModules($command, $config_key) {
         ->drush("$command $modules_list")
         ->run();
       $exit_code = $result->getExitCode();
+      $eventInfo['active'] = TRUE;
+      $eventInfo['modules'] = md5($modules_list);
```
alexxed: Why would we send an md5 hash to amplitude?

src/Robo/Hooks/CommandEventHook.php
```diff
@@ -41,4 +43,28 @@ public function issueWarnings(ConsoleCommandEvent $event) {
...
+  public function telemetry(ConsoleCommandEvent $event) {
+    $userConfig = new UserConfig(Blt::configDir());
+
+    // Ask to enable telemetry if necessary.
+    if (!$userConfig->isTelemetrySet() && $this->input()->isInteractive()) {
+      $this->say("We strive to give you the best tools for development.\nYou can really help us improve by sharing anonymous performance and usage data.\n");
+      $preference = $this->confirm("Do you want to help us make this tool even better?", TRUE);
+      $userConfig->setTelemetryEnabled($preference);
+      if ($preference) {
+        $this->say("🎉 Awesome! Thank you for helping out!");
```
alexxed: Is there a command to opt out in case you're changing your mind?

src/Robo/Commands/Setup/ToggleModulesCommand.php
danepowell: I want to know if the modules list has been changed from the BLT-provided defaults, or if it matches a few other common scenarios (i.e. if it's empty). But I don't want to send module names that could be potentially sensitive / identifiable.

src/Robo/Hooks/CommandEventHook.php
danepowell: Good point, there is not. I'll create a follow up: https://github.com/acquia/blt/issues/3922

PR #3913, opened 2019-11-07
Remove clear-twig-cache.sh which is now done by the main platform

rmuit:
Changes proposed
---------
The post-theme-deploy hook to clear Twig caches was introduced in PR #2229 / 2440. IMHO the reason this was necessary is, the combination of Drupal Core / Cloud Platform / ACSF Platform wasn't doing a great job of clearing caches.

(I just noticed that there is a proposed @todo mentioned in the PR and comments of #2229, to remove this hook after https://www.drupal.org/node/2752961 would be committed. That turned out to _technically_ not be 100% true, because 
* the Twig cache has been correctly _invalidated_ since Drupal 8.6
* but old stale files were not removed from non-primary webs, so this hook was still good to have around.
)

I'm preliminarily opening this against 11.x because otherwise I'll forget: as of thursday Nov 14, Site Factory 2.96 will be released and the cache clearing situation will be handled by the platform. I promise we now know what we are doing. (Internal ticket searches will show recent tickets from several deparments referring to Twig cache clearing.)

-----------
In the detailed (exported) Site Factory WIP logs for ProcessThemes, the "executePostThemeDeploy" state will _not_ output any message anymore saying that the clear-twig-cache.sh script is getting run. Which is fine, because (as of SF release 2.96) slightly before that state, a new "clearTwigCache" state is visible.

Additional details
-----------
Leaving this script in won't cause adverse effects, it just executes a duplicate action.


danepowell:
Awesome, thanks for the heads up. I might keep this open until Nov 14 and then backport it to 10.x as well. Even though it sounds like it won't do any harm for existing users, we might as well keep things tidy just in case.

PR #3918, opened 2019-11-13
DT-972: Fixes #3916: Broken automated tests.

### LINKED ISSUE #3916: DT-972: Automated tests failing
Lightning Workflow has a new release with a broken simpletest definition, which causes our test running simpletest to fail.

danepowell:
Fixes #3916 
--------

Changes proposed
---------

-
-

Steps to replicate the issue
----------
1. 
2. 

----------
(remove if this is not a bug fix)

-----------

Additional details
-----------


danepowell:
This works but probably isn't the best solution as it pins the Lightning Workflow version for all end users, which is likely to lead to its own set of problems in the future.

PR #3925, opened 2019-11-19
DT-596: Fixes #3052: Add Git pre-push validation.

### LINKED ISSUE #3052: DT-596: Move slow validation from Git pre-commit hook to pre-push
Running linters and sniffers takes a lot of time, whereas committing should be quick. It really only matters that code isn't pushed with errors and violations in it. I suggest moving BLT's `pre-commit` hook implementation to `pre-push`.

danepowell:
Fixes #3052 
--------

Changes proposed
---------
- Add a new pre-push git hook that runs the full validation suite

Steps to replicate the issue
----------
1. Run `blt blt:init:git-hooks` to install the new hook.
2. Create a custom file that fails validation (maybe a custom module, or modify composer.json without updating composer.lock)
3. Create a new commit and try to push it. Ensure that validation runs both during commits and pushes and catches the failure.
4. Follow instructions in the linked docs to optionally disable either the commit or push validation.

Additional details
-----------
Note that the Git hooks can be modified or disabled entirely using the process described here: https://github.com/acquia/blt/blob/11.x/docs/extending-blt.md#git-hooks

So @TravisCarden would probably want to disable pre-commit and just use pre-push, while @wu-edward might want the opposite. Let me know if this satisfies your use cases.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -86,6 +86,7 @@ git:
   # You should execute `blt blt:init:git-hooks` after modifying these values.
   hooks:
     pre-commit: ${blt.root}/scripts/git-hooks
+    pre-push: ${blt.root}/scripts/git-hooks
     commit-msg: ${blt.root}/scripts/git-hooks
   commit-msg:
     # Commit messages must conform to this pattern.
--- a/scripts/git-hooks/pre-push
+++ b/scripts/git-hooks/pre-push
@@ -1,11 +1,9 @@
 #!/usr/bin/env bash
 
 ROOT_DIR="$(pwd)/"
-LIST=$( git diff --name-only --cached --diff-filter=ACM )
-PHPCS_BIN=vendor/bin/phpcs
 
 echo "Executing .git/hooks/pre-push..."
-"${ROOT_DIR}"/vendor/bin/blt internal:git-hook:execute:pre-push "${LIST}"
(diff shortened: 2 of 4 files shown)


danepowell:
@TravisCarden @wu-edward pinging you since you were on the original feature request... note that for pre-push, we have to run the full validation suite (not just validation of changed files), which takes a bit longer. Do you think most customers would want that behavior by default, or should we disable the pre-push hook by default?

TravisCarden:
Nice, @danepowell! I know _I_ would want the behavior by default. Thinking from the perspective of a technical architect, I would probably prefer the `pre-push` check to `pre-commit` as the default since `pre-commit` can be overridden (`git commit --no-verify`), whereas `pre-push` cannot to my knowledge. And it seems preferable to do a full check once last time for very rapid feedback compared to the delay involved in getting a failed build on CI.

hugovk:
> Note that the Git hooks can be modified or disabled entirely using the process described here: [/docs/extending-blt.md@`11.x`#git-hooks](https://github.com/acquia/blt/blob/11.x/docs/extending-blt.md#git-hooks) [](/acquia/blt/blob/HEAD@{2019-11-19T22:34:34Z}/docs/extending-blt.md#git-hooks)

New link: https://github.com/acquia/blt/blob/11.x/docs/extending-blt.rst#git-hooks

PR #3926, opened 2019-11-20
DT-589: Fixes #3559: Support Behat 3.4+.

### LINKED ISSUE #3559: DT-589: Support Behat 3.4+

danepowell:
Fixes # 
--------

Changes proposed
---------

-
-

Steps to replicate the issue
----------
1. 
2. 

----------
(remove if this is not a bug fix)

-----------

Additional details
-----------

### DIFF
--- a/subtree-splits/blt-require-dev/composer.json
+++ b/subtree-splits/blt-require-dev/composer.json
@@ -4,7 +4,7 @@
     "description": "A wrapper package for BLT's require dev dependencies.",
     "license": "GPL-2.0-only",
     "require": {
-        "behat/behat": ">=3.1 <3.4",
+        "behat/behat": "~3.1",
         "bex/behat-screenshot": "^1.2",
         "dmore/behat-chrome-extension": "^1.0.0",
         "drupal/drupal-extension": "~3.2",


danepowell:
Seems like the change to the dir / arg may not be necessary if we apply the patch from here, or wait for it to be fixed and released: https://github.com/Behat/Behat/issues/1076

danepowell:
Aha... indeed, applying the upstream patch by itself fixes everything. We may need to wait for a new release with that.

Although changing the dir/arg as in this PR fixes one problem, it just leads to another because we rely on passing args to constrain what tests get run (i.e. the currently failing tests are trying to run Lightning tests on a minimal install, which obviously won't work). Getting rid of the args breaks a fundamental feature and isn't an acceptable workaround.

PR #3927, opened 2019-11-21
DT-587: Fixes #3564: Simulate deploys on TravisCI.

### LINKED ISSUE #3564: DT-587: Changes to checkDirty behavior during deployments

danepowell:
Fixes #3564 
--------

Changes proposed
---------
- Stop ignoring dirty build directories during deploys on Travis
- Since ^ will likely generate subtle failures for some customers, simulate deploys during the test process to make any errors obvious.

Additional details
-----------
I've verified that the new .travis.yml file works great on a BLTed project, and only adds a few seconds to the build process.

### DIFF
--- a/scripts/blt/deploy/.gitignore
+++ b/scripts/blt/deploy/.gitignore
@@ -11,6 +11,7 @@ local.blt.yml
 # Ignore build artifacts.
 /tmp/
 /reports/
+/travis_wait*
 
 # Node.js package manager.
 npm-debug.log
--- a/scripts/travis/.travis.yml
+++ b/scripts/travis/.travis.yml
@@ -48,6 +48,7 @@ script:
   # Uncomment these lines to test database updates using live content.
   # - blt drupal:sync:default:site
   - source ${BLT_DIR}/scripts/travis/run_tests
+  - source ${BLT_DIR}/scripts/travis/simulate_deploy
 
 deploy:
    - provider: script
--- a/scripts/travis/deploy_branch
+++ b/scripts/travis/deploy_branch
@@ -2,6 +2,6 @@
 
 set -ev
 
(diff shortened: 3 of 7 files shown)


scripts/blt/deploy/.gitignore
```diff
@@ -11,6 +11,7 @@ local.blt.yml
 # Ignore build artifacts.
 /tmp/
 /reports/
+/travis_wait*
```
danepowell: Travis sometimes generates this log file during tests. It will prevent deploys unless specifically excluded from Git. Plus, no one actually wants to deploy this file.

Issue #3563, opened 2019-04-21
Use composer-scaffold to build new BLT projects


lcatlett:
**Is your feature request related to a problem? Please describe.**
Given the recent progress in adding a composer scaffolding plugin to core to resolve https://www.drupal.org/project/drupal/issues/2982684, BLT should update the current method of assembling files when a new project is created or updated to use @grasmash's new [composer-scaffold plugin](https://github.com/grasmash/composer-scaffold) since it is now the basis for the in-progress core-patch which does the same: https://www.drupal.org/project/drupal/issues/2982684#comment-13067069)

**Additional context**
The readme in `grasmash/composer-scaffold` has significant updates in the not-yet-merged https://github.com/grasmash/composer-scaffold/blob/824b0eb12600ffb9edeca46cd68e3a27efe417de/README.md which outline how vendors and projects like BLT are expected to manage vendor overrides to config and the location of other files provided by core.

danepowell:
This is the first I've heard of composer-scaffold. If I understand correctly, the goal here would essentially be to replace drupal-scaffold (which BLT currently uses to download non-composer Drupal files) with composer-scaffold? composer-scaffold's readme also mentions being an alternative to composer/installers, but I don't see how composer-scaffold supports placing modules and other packages into arbitrary directories, it only has examples for core.

Drupal Scaffold is pretty well established and broadly used, such as by drupal-project and lightning-project. I'm hesitant to hitch our cart to a new horse, as neat as it sounds, unless that's where the whole community is going. Have the maintainers of drupal-scaffold talk about officially deprecating that project? Has lightning-project talked about switching? These are the kinds of questions we'd need to answer before making such a big shift.

Maybe @grasmash can provide some context?

grasmash:
The idea is to actually merge Composer Scaffold into Drupal Core and then deprecate Drupal Scaffold. Once it's committed to core in a dev branch, it would be worthwhile for BLT to begin looking at deprecating Drupal Scaffold in preparation for Drupal 8.8.0, which is the target release for core scaffolding functionality.

danepowell:
Thanks, that context is a lot more helpful. I'll add this to the 11.0 milestone, since BLT 11.0 will be released coincidentally with Drupal 8.8.

danepowell:
This has been merged in core: https://www.drupal.org/docs/develop/using-composer/using-drupals-composer-scaffold

[...]

PR #3930, opened 2019-11-26
fix(drupal-8.8): Replace core, drupal-scaffold and drupal-core-requir…

### LINKED ISSUE #3929: DT-1035: drupal-composer/drupal-scaffold conflicts with drupal/core-composer-scaffold
From Drupal 8.8 it is recommended to use the drupal/recommended-project composer project template as per https://www.drupal.org/docs/develop/using-composer/starting-a-site-using-drupal-composer-project-templates.

Adding BLT to this currently fails due to the conflict between drupal/core-composer-scaffold (new scaffold plugin, maintained by core team) and drupal-composer/drupal-scaffold (old scaffold plugin, not maintained by core team).

BLT should require the new drupal/core-composer-scaffold plugin instead.

CraigGardener:
Replace drupal core, require-dev, and drupal-scaffold with recommended packages from Drupal 8.8.

Fixes  #3929 and allows BLT to be added to projects created with drupal/recommended-project as per https://www.drupal.org/docs/develop/using-composer/starting-a-site-using-drupal-composer-project-templates


danepowell:
Thanks for this first pass. We're going to implement this in #3563. An important distinction is that the changes will go into the BLT project template's composer.json, not into BLT's own composer.json.

PR #3937, opened 2019-12-02
Allow blt:init:shell-alias install confirmation to be configurable when run with --no-interaction

lpeabody:
Changes proposed
---------
When issuing `blt blt:init:shell-alias --no-interaction`, the shell alias should be installed rather than not installed (by default). This could be a configurable value for BC purposes. For automated project setups that run with `--no-interaction`, this just results in the alias not ever being setup properly.

Steps to replicate the issue
----------
1. Without the shell alias installed, run `blt blt:init:shell-alias --no-interaction`.
2. Run `blt --version` and note that `blt` cannot be found.

----------
Running `blt blt:init:shell-alias --no-interaction` does not setup the shell alias.

-----------
Running `blt blt:init:shell-alias --no-interaction` sets up the shell alias after setting in blt.yml:
```
blt:
  alias:
    auto-install: true
```

Additional details
-----------
Previous to BLT 10.x you could run `composer run-script blt-alias` and this would install the alias for you, which is what our automated setup scripts have used. However on 10.x it's no longer possible to get this behavior since the composer script was removed and the dedicated BLT command does not work in the same manner.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -24,6 +24,8 @@ blt:
     example-local: ${repo.root}/blt/example.local.blt.yml
     schema-version: ${repo.root}/blt/.schema_version
   command-cache-dir: ${blt.root}/cache/commands
+  alias:
+    auto-install: false
 
 composer:
   bin: ${repo.root}/${bin.path}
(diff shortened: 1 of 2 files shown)


lpeabody:
Rad! Figured I'd ask before merging, but do you think it makes more sense to default `blt.alias.auto-install` to `true` or `false`? Right now I have it set to `false` for BC purposes, but I think the reasonable default is `true`. So, do you think it's more important to stick with a BC approach or go with the reasonable default in this instance?

danepowell:
I can merge it into 10.x in the BC-compatible way, then change the default when I forward-port it to 11.x (which should get a stable release in the next week or two).

PR #3942, opened 2019-12-06
DT-1035: Fixes #3929: Move Drupal Scaffold to project composer.json.

### LINKED ISSUE #3929: DT-1035: drupal-composer/drupal-scaffold conflicts with drupal/core-composer-scaffold

danepowell:
Fixes #3929 
--------

Changes proposed
---------
- Remove Drupal Scaffold from BLT's dependencies
- Add Drupal Scaffold to the project template so new projects still get it
- Add an update hook so existing projects still get it

Steps to replicate the issue
----------
1. Apply this patch and run `blt update`, ensure that Drupal Scaffold is added to your composer.json file and otherwise nothing changes. 

Additional details
-----------
Should have no functional impact on projects, but it will free them and us up to migrate to Composer Scaffold at one's leisure. I don't want to backport this to 10.x since it requires a manual run of `composer update`, which users should be doing anyway during the 11.x upgrade.


danepowell:
ORCA is pulling the 10.x stable release of blt-project, which is doubly bad because it should be 11.x, and it should also be the dev version from this PR.

reubenmoes:
👍

danepowell:
@phenaproxima @TravisCarden BLT's ORCA builds here have suddenly started hitting the missing vfsStream dependency problem, as you can see in the last test run above. Did ya'll end up patching that in ORCA? Just wondering how to get this build passing.

@alexxed @adrianthedev please review this despite the failing test if you don't mind, so we can get it merged Monday.

TravisCarden:
@danepowell I'm actually rolling an ORCA hotfix for this issue right now. I'll report back shortly.

TravisCarden:
@danepowell I released the hotfix and restarted your failing build, but it now fails due to a patch that no longer applies. Obviously, I can't fix that for you on my side.

danepowell:
We still need the patch from https://github.com/acquia/orca/pull/52 (against ORCA `develop`), but it conflicts due to the ORCA hotfix that only exists on `master`. Travis is going to fix that and then we should be unblocked.

PR #3944, opened 2019-12-09
DT-588: New projects use Composer Scaffold.

### LINKED ISSUE #3563: Use composer-scaffold to build new BLT projects

danepowell:
Fixes #3563 
--------

Composer Scaffold replaces and deprecates the older Drupal Scaffold as of Drupal 8.8. However, the two packages conflict and since only Drupal Scaffold is available prior to 8.8, some accommodations have to be made. Namely, we have to disable the ORCA "previous Drupal release" test until Drupal 8.9 is released. This is fine since in real-world usage, customers are unlikely to be creating new Drupal 8.7 projects.

Note that this PR only affects _new_ projects. Existing Drupal 8.7 users can still update to BLT 11, or add BLT 11 to their projects.

Changes proposed
---------
- Update template composer.json to use `acquia/composer-scaffold` instead of `drupal-scaffold`, which pulls in Composer Scaffold.
- Update template composer.json extra config as necessary to work with Composer Scaffold.
- Remove template files that are now managed by Composer Scaffold
- Fix ACSF commands to work with Composer Scaffold.

Steps to replicate the issue
----------
1. Copy blt-project template to tmp directory, due to Composer limitation: `cp -r subtree-splits/blt-project /tmp/`
1. Tell Composer to use your local blt-project: `composer config --global repositories.blt-project path /tmp/blt-project`
1. Tell Composer to use your local blt: `composer config --global repositories.blt path /home/dane/packages/blt`
1. Test creating a new project: `composer create-project --no-interaction acquia/blt-project:dev-master`
1. Ensure the newly create project functions normally.
1. Restore your global Composer config: `composer config --global --unset repositories.blt && composer config --global --unset repositories.blt-project`


composer.json
```diff
@@ -28,7 +28,6 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
         "dflydev/dot-access-data": "^1.1.0",
         "doctrine/common": "^2.5",
-        "doctrine/inflector": "~1.1.0",
```
danepowell: This "dependency" only existed to support PHP 5.6: https://github.com/acquia/blt/pull/1829

danepowell:
@TravisCarden I have a conundrum. This PR is changing BLT's project template to use Composer Scaffold, which is only available in Drupal 8.8. Thus you cannot create Drupal 8.7 projects anymore. From a customer standpoint, I think that's fine (no one wants to create new projects on an old version of core.) But this causes ORCA's previous version test to choke.

Any ideas how to solve this? To solve this in the spirit of ORCA, for the previous version test, it seems like you'd want to test that you can add BLT to an _existing_ Drupal 8.7 project (which is possible), rather than testing that you can _create_ a new Drupal 8.7 project with BLT (which is not possible). But that's a pretty big architectural change for ORCA.

.travis.yml
```diff
@@ -26,7 +26,7 @@ env:
     - COMPOSER_BIN=$TRAVIS_BUILD_DIR/vendor/bin
     - BLT_DIR=$TRAVIS_BUILD_DIR
     - DRUPAL_CORE_HEAD=8.8.x-dev
-    - BLT_PRINT_COMMAND_OUTPUT=0
+    - BLT_PRINT_COMMAND_OUTPUT=1
```
danepowell: need to revert this after tests pass

subtree-splits/blt-project/composer.json
```diff
@@ -33,9 +33,14 @@
...
-                "sites/default/default.services.yml": "sites/default/services.yml",
-                "sites/default/default.settings.php": "sites/default/settings.php"
+            "locations": {
+                "web-root": "./docroot"
+            },
+            "allowed-packages": {
+                "drupal/core": "*"
+            },
+            "file-mapping": {
+                "[web-root]/sites/default/default.services.yml": false
```
danepowell: workaround for https://www.drupal.org/project/drupal/issues/3091285, make a note to remove this after it's fixed

TravisCarden:
Well, @danepowell, ORCA's policy is to test all _supported_ scenarios. (See, for example, how [Lightning Core only runs a couple of jobs](https://github.com/acquia/lightning_core/blob/8.x-3.11/.travis.yml#L24).) So if you _can't_ create a new BLT site on Drupal 8.7, then you don't need to. On the other hand, as you point out, BLT has functions independent of new site creation that should continue to work and are still supported. If (as you will know better than I) you have a clean version break for core version support (e.g., 10.x supports `<8.8.0` and 11.x support `>=8.8.0`) then you can just run the appropriate tests on each branch. Lightning has such a branching strategy, as can be seen for example in [Lightning Core's Branch Compabitility table](https://www.drupal.org/project/lightning_core). If you don't currently take such an approach, you might consider it. As far as I've thought it through, it's about the only sane solution within the current constraints of ORCA's design.

danepowell:
Thanks. It's not ideal but I agree that just killing that test is the best approach for now. Done and done.

subtree-splits/blt-project/composer.json
danepowell: Workaround will be removed in DT-1076.

composer.json
```diff
@@ -53,11 +52,11 @@
     "require-dev": {
         "acquia/coding-standards": "^0.4.0",
         "composer/composer": "^1.6.4",
+        "drupal/core-dev": "^8.8.0",
         "knplabs/github-api": "^2.6",
         "php-http/guzzle6-adapter": "^1.1",
         "phpunit/phpunit": "^4.8 || ^6.5",
-        "squizlabs/php_codesniffer": "^3.0.1",
-        "webflo/drupal-core-require-dev": "^8.6.0"
```
alexxed: :tada:

PR #3960, opened 2019-12-17
ORCA create BLT 10 projects.

danepowell:
This should contain the fix from https://github.com/acquia/orca/pull/52 that hopefully fixes tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
     # ORCA configuration.
     - ORCA_SUT_NAME=acquia/blt
     - ORCA_SUT_BRANCH=10.x
-    - ORCA_VERSION=develop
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,13 +33,6 @@ matrix:
   include:
     # Standard ORCA jobs.
     - { name: "Static code analysis", env: ORCA_JOB=STATIC_CODE_ANALYSIS }
-    - { name: "Deprecated code scan", env: ORCA_JOB=DEPRECATED_CODE_SCAN }
```
danepowell: We can't run any of these because ORCA tries to add Drupal 8.8 dependencies that conflict with our support of Drupal 8.7. I think webflo/drupal-core-require-dev might be the main problem.

PR #3963, opened 2019-12-18
Switch back to ORCA dev-master

danepowell:
If this works, wait for ORCA stable release and change ORCA_VERSION to `dev-master` and merge.


danepowell:
Per suggestion from Travis in Slack, I'm going to remove the previous version test after all. Although it happens to work right now, it still tests and undefined scenario, which is to create a BLT project using the 10.x template and then immediately install BLT 11.x _without running any update scripts_. Since it's likely to break at some point and not a supported scenario to begin with, we should just keep it out.

PR #3975, opened 2020-01-07
Adds a config value for deployment docroot.

b-sharpe:
Changes proposed
---------
Set the deploy docroot as config so it can be different than "docroot", this is helpful if your remote environment uses something other docroot such as pantheon which uses "web"

See #3740 for previous history

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -54,6 +54,7 @@ deploy:
   # If true, dependencies will be built during deploy. If false, you should commit dependencies directly.
   build-dependencies: true
   dir: ${repo.root}/deploy
+  docroot: ${deploy.dir}/docroot
   exclude_file: ${blt.root}/scripts/blt/deploy/deploy-exclude.txt
   exclude_additions_file: ${repo.root}/blt/deploy-exclude-additions.txt
   gitignore_file: ${blt.root}/scripts/blt/deploy/.gitignore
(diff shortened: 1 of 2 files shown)


danepowell:
Thanks for the contribution!

danepowell:
Ugh... this has the unfortunate side effect of breaking deploys for anyone that sets a custom `deploy.dir` without also setting `deploy.docroot`. It appears that when you pass `deploy.dir` via the CLI, it's not incorporated into the value of `deploy.docroot` as we've defined it in config/build.yml. Robo has a few edge cases when it comes to recursion / order of operations with configuration, I guess this is one of them.

danepowell:
I'm going to revert it in the 10.x branch for the above reason. 10.x should only have completely stable patch releases.

b-sharpe:
@danepowell I'm not sure what to do here. The first PR ( #3740 ) was to keep things relative to repo root, in which case it was determined that it is usually defined to include repo root, but here you are saying that for deploy it would need to be relative or else it won't change when passed via CLI.

I'm happy to re-patch, but there needs to be a decision here. 

If we keep as above then someone passing to CLI `deploy.dir` also needs to pass `deploy.docroot` and it's just understood as with the repo root/docroot noted above. 

Otherwise, we need `deploy.docroot` to be relative, but then it functions different from the main one.

Let me know your thoughts here as we need this functionality to deploy to hosts not using `docroot`

PR #3986, opened 2020-01-17
Added source:link command to support local package development.

danepowell:
Developing Drupal modules and packages in the era of Composer and virtual environments like DrupalVM and Lando can be hard.

I prefer to keep my modules and packages under development in a separate directory from my Drupal application, so they can be version controlled independently of Composer. But then I obviously need to test them in a real Drupal application somehow.

This can be done using Composer path repositories and NFS / volume mounts for DrupalVM / Lando. It's tricky set up, so I put together a command to automate it.

Unfortunately there are still some limitations, mainly around the directory structure. Because the Composer path repo symlink gets mounted in the container, the symlink must be relative _and_ it must be valid in both environments. I haven't tested all permutations of this, just sibling and cousin directories. This is noted in the comments.

This code is a bit ugly, but it's also an ugly problem. If we had time, there are a few slight improvements possible, such as using composer to guess the package name and version instead of requiring them as arguments.


danepowell:
Oh, we need inline help for the parameters...

PR #3988, opened 2020-01-20
Fix README.md 404 links.

leymannx:
Now that the docs were changed to RST instead of markdown, a lot of links produce 404s. I replaced these links with absolute links to the generated docs on https://docs.acquia.com. I couldn't find any replacement for the FAQs.

Actually the LICENSE on the very bottom of the README probably should be updated or removed completely as well.


leymannx:
Added reviewers from suggestions.

danepowell:
Thanks a lot for the contribution!

PR #4004, opened 2020-02-06
Remove platform.php Composer configuration in blt-project

### LINKED ISSUE #4003: DT-1183: config.platform.php in blt-project prevents installation of packages requiring PHP >7.2
https://github.com/acquia/blt/pull/3769 changed `blt-project`'s configuration to make Composer negotiate requirements as if the installed version of PHP were 7.2, regardless of the version actually installed. I understand the motivation, but it's probably a bad practice because it may mask errors off-prod by allowing developers to develop and test locally with a different version of PHP than prod has. But at least as important, it prevents the installation of Composer libraries requiring later versions of PHP... like Drupal 9. Thus this is a D9 readiness blocker and I request that we remove it.

Now it might be argued that this isn't urgent or even necessary since developers can just remove the config from their own implementations. But I would observe that besides the fact that the setting is not well-known and can thus cause difficult to diagnose errors, it will prevent ORCA from installing Drupal 9 and already prevents ORCA users from requiring PHP >7.2 in their own packages.

As to the motivation behind the original change, I have argued that it's a bad idea. But users can always add the configuration themselves. And as it's probably a minority use case, that seems like the more appropriate approach to take. For users who really like bad ideas and want to make their futures selves sorry, I've confirmed that you can set the config globally: `composer config --global platform.php 7.2`. (But seriously, don't do that. Just install the correct version of PHP locally. If you need to switch between versions for different projects and you're using Homebrew on Mac, use [Brew PHP Switcher](https://github.com/philcook/brew-php-switcher).)

TravisCarden:
Fixes #4003

### DIFF
--- a/subtree-splits/blt-project/composer.json
+++ b/subtree-splits/blt-project/composer.json
@@ -25,9 +25,6 @@
         "acquia/blt-require-dev": "11.x-dev"
     },
     "config": {
-        "platform": {
-            "php": "7.2"
-        },
         "sort-packages": true
     },
     "extra": {


wimleers:
Discovered in https://github.com/acquia/acquia_migrate/pull/22 — many thanks @TravisCarden for digging into this and determining the root cause! 🙏

danepowell:
Thanks for the PR, but I'd rather solve this without completely removing the config. See #4003

PR #4008, opened 2020-02-12
Add --prefer-dist to artifact

### LINKED ISSUE #4007: Artifact build executes composer with unideal parameters
**Bug description**
Currently the build executes `composer install --no-dev --no-interaction --optimize-autoloader`. This works fine in most cases but has a noteworthy shortcoming: In some cases the source is fetched even though a tagged version is available (we saw this with an RC) and thus versioning information is not available in the info.yml.

This can break Drupal's dependency checks.

Example:
```
 [error]  CRM Core Activity requires this module and version. Currently using Name Field version  (Currently using Unresolved dependency Name Field (Version &gt;=8.x-1.0-rc2 required)) 
```

**System information**
* BLT version: 11

The PR adds the `--prefer-dist` argument to avoid this issue.

grahl:
Fixes #4007
--------

Changes proposed
---------
- Build with `--prefer-dist`

### DIFF
--- a/src/Robo/Commands/Artifact/DeployCommand.php
+++ b/src/Robo/Commands/Artifact/DeployCommand.php
@@ -543,7 +543,7 @@ class DeployCommand extends BltTasks {
       ->copy($this->getConfigValue('repo.root') . '/composer.lock', $this->deployDir . '/composer.lock', TRUE)
       ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
       ->run();
-    $command = 'composer install --no-dev --no-interaction --optimize-autoloader';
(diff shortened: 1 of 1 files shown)


danepowell:
We specifically do not use `--prefer-dist` because it breaks compatibility with commit refs: https://github.com/acquia/blt/pull/606

It shouldn't be an issue because even without `--prefer-dist`, Composer should always pull a stable package when available. If it doesn't, that seems like a bug with Composer, not BLT.

If there's some other way that BLT can solve this, or if we can show that https://github.com/acquia/blt/pull/606 is no longer relevant, I'm happy to consider it, but as it stands this seems like it would introduce a regression.

PR #4020, opened 2020-02-20
Fixes #4005: DT-1193: Remove usage of $_ENV

### LINKED ISSUE #4005: DT-1193: Use getenv instead of $_ENV
The `$_ENV` variable values are dependent on how `variables_order` is [configured](https://www.php.net/manual/en/ini.core.php#ini.variables-order) in php. If the `E` option isn't present, normal environment variables don't appear. The `getenv` function always returns environment variables that are present, so switching to that throughout would be more stable. 

I encountered this issue with the code in `EnvironmentDetector::isCiEnv()` but a search of the codebase reveals many usages of `$_ENV`.

danepowell:
Fixes #4005 
--------

Changes proposed
---------
- Use getenv() / putenv() instead of $_ENV

Additional details
---------
We might want to mention this change in the release notes, in case anyone is setting $_ENV directly via a php script or setting (instead of via the shell), which this would break.

Additionally, this may break BLT in CI environments where variables_order has not been set to export `$_ENV`, and BLT was previously erroneously detecting said environments as `local`. That's what broke ORCA....


danepowell:
It looks like BLT was previously not properly detecting that it was running in a CI environment during ORCA tests, and now it is. While BLT's behavior is correct now, that breaks ORCA (which writes to local settings instead of CI settings). Why....

PR #4025, opened 2020-02-26
DT-1180: Upgrade dependencies for Drupal 9.

danepowell:


danepowell:
Tests are incorrectly exiting early.  wth is a "Invalid symmetric difference expression", is Travis CI providing an incorrect commit range, and why?

danepowell:
Per Travis in Slack, Drupal 8 is hardcoded into ORCA, so if we want this to pass ORCA tests we need to hack its scripts or wait for it to support D9.

danepowell:
Take two... we'll worry about ORCA later... https://github.com/acquia/blt/pull/4037

PR #4029, opened 2020-02-26
updating readme to better represent supported drush versions.

mikemadison13:
Fixes updates the readme to reflect that BLT 11 and 12 still work with Drush 9.5.x.


danepowell:
Thanks!

PR #4033, opened 2020-02-28
DT-1178: Kill blt-project and blt-require-dev, aka BYOD

danepowell:
Let's make adding BLT to an existing project the _preferred_ installation method so that you can add it to any distribution you want, not just Lightning, or easily add it to a project that's already under development. Call it "bring your own distribution" (BYOD) 😄 

To accomplish this, when BLT is installed via Composer, it will copy over any necessary configuration files at that time. For other things like Behat that aren't critical for all projects, only install them on-demand (this also lets us get rid of blt-require-dev.)

For backwards compatibility at least for now, we'll keep a skeleton blt-project that doesn't contain anything specific to BLT just for testing purposes until we can get ORCA converted to the new paradigm as well.


danepowell:
If this works, consider getting rid of the rsync functionality in the updates command file.

leymannx:
Wow! That probably would be a drastic change, right? But I'd definitely welcome this. If this is lowering the entry threshold and/or is making BLT more generic to also be used in non-Acquia environments that would be great.

danepowell:
Yes it's a big change but only a good way, as far as I can see. Anyone who wants to do things the old way (creating projects with blt-project) still can, but this will make it easier to use BLT as a normal composer package (`composer require acquia/blt`), as well as drastically reduce the number of dependencies so it's lighter weight and faster to install.

danepowell:
Hmm... I should actually split this up into separate PRs for getting PHPUnit/Behat components out of the project template, and then doing the rest of the cleanup. Also let's keep all of the other template components in the blt-project directory for now, just to preserve backwards compatibility.

danepowell:
This PR should no longer be necessary following the related PRs. There's still some related work to do around this but it's mostly around testing, documentation, and getting the word out.

PR #4037, opened 2020-03-04
DT-1181: Updated dependencies for Drupal 9.

danepowell:
I've verified locally that this can be added to a Drupal 9 project as expected, now we just need to get ORCA working. After that and this is merged...

- [x] Come up with a plan to fix disabled tests (packages like ACSF and Config Import that aren't ready for D9)


danepowell:
@TravisCarden @balsama how can we coordinate to get all of BLT / Lightning / ORCA upgraded to Drupal 9? I know you mentioned that BLT would need to fork ORCA's scripts to run Drupal 9, but if we do that and merge this PR, won't it just break all other ORCA packages that _aren't_ running Drupal 9?

danepowell:
Per Travis, I need to disable/remove the failing ORCA jobs since they will always try to run on the "current" version of Drupal core (8.8), which is incompatible with BLT 12, and replace them with a [new job that sets something](https://github.com/acquia/orca/blob/develop/docs/advanced-usage.md#travis-ci-scripts) like `ORCA_CUSTOM_FIXTURE_INIT_ARGS="--core=9.0.0-alpha1"`

I'll also need to disable all BLT's tests that rely on packages such as the ACSF module that aren't compatible with D9.

danepowell:
This is ready to go if we can just figure out why ORCA is trying to run PHPUnit from the fixture directory, which doesn't make sense because PHPUnit generally shouldn't be a production dependency of packages.

PR #4038, opened 2020-03-04
Allow pre-push hook to fail invalid commit

yyss8:
Fixes # 
--------
- Allows `pre-push` hook to fail invalid commit.

Changes proposed
---------
- The `pre-push` hook introduced in https://github.com/acquia/blt/pull/3925 is currently not breaking any invalid commits.  I think it makes no sense to run the hook every time before a `git push` if it's not preventing invalid commit from being pushed to remote.

Steps to replicate the issue
----------
1. Make some changes that are going to fail when doing `blt validate`.
2. Disable `pre-commit` hook or commit using `git commit --no-verify`
3. Push the commit.

----------
`pre-push` is not preventing invalid commit from being pushed to remote.
image

-----------
`pre-push` will prevent invalid commit from being pushed to remote.
image

Additional details
-----------

### DIFF
--- a/src/Robo/Commands/Internal/GitCommand.php
+++ b/src/Robo/Commands/Internal/GitCommand.php
@@ -103,7 +103,7 @@ class GitCommand extends BltTasks {
     }
     catch (BltException $e) {
       $this->yell('Your code has failed pre-push validation.');
-      return;
+      return 1;
     }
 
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks for the catch and PR. I was able to reproduce this and confirm the fix. However this will need to go into the 12.x branch first: #4044

PR #4045, opened 2020-03-09
Switch to Acquia Coding Standards.

danepowell:
Before merging, I need to think about how this should work for users adding BLT to an existing project. We can't add acquia/coding-standards to composer dynamically on install. Could be a hard dependency of BLT, but that's not ideal.

Probably put the coding standards in a separate composer package and suggest that people install it after installing BLT.


danepowell:
Replaced by #4077

PR #4049, opened 2020-03-10
Fixes #4046: Remove dependency on MySQL database

### LINKED ISSUE #4046: Support PostgreSQL
BLT assumes that Drupal is using MySQL.

Much of BLT's functionality can not be used with sites using PostgreSQL.

It should be possible for PostgreSQL (and other DBs) to be supported by changing the way BLT checks for Drupal being installed by tweaking MySQL specific code such as https://github.com/acquia/blt/blob/1047fc82e0aab4ef465c345d8d9beb28f8ab3db3/src/Robo/Inspector/Inspector.php#L225

CraigGardener:
Fixes # 
--------

https://github.com/acquia/blt/issues/4046

Changes proposed
---------
- Separate checks for database availability
- isDrupalInstalled to use "drush status bootstrap" so not dependent on MySQL

### DIFF
--- a/src/Robo/Inspector/Inspector.php
+++ b/src/Robo/Inspector/Inspector.php
@@ -56,6 +56,13 @@ class Inspector implements BuilderAwareInterface, ConfigAwareInterface, Containe
   protected $isMySqlAvailable = NULL;
 
   /**
+   * Is PostgreSQL available.
+   *
+   * @var null
+   */
+  protected $isPostgreSqlAvailable = NULL;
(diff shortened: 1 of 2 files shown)


danepowell:
Thanks for the PR!

danepowell:
@CraigGardener wait isn't there a bug here? It doesn't look like `isPostgreSqlAvailable()` is ever called. Don't you mean to call to call that from the switch condition?

PR #4052, opened 2020-03-12
Correctly detect `$_acsf_site_name` for local multisites setup.

### LINKED ISSUE #4051: Wrongly detect `$_acsf_site_name` for local multisites setup?
**Describe the bug**

From the below comment, it instructs the local domain name should be `local.[site-name].com

https://github.com/acquia/blt/blob/81ca1cade4eb4b5c5ab8a36a6c668c5d128bee9e/settings/blt.settings.php#L87

later on 

https://github.com/acquia/blt/blob/81ca1cade4eb4b5c5ab8a36a6c668c5d128bee9e/settings/blt.settings.php#L90 

slice the domain fragments array starting index 1, which will result in an array `[ 'site-name', 'com']`, 

then

https://github.com/acquia/blt/blob/81ca1cade4eb4b5c5ab8a36a6c668c5d128bee9e/settings/blt.settings.php#L92

 checks if the above `$name` is in the blt `multisites` settings array `['default', 'site-name']` which would result in a FALSE always.

OR, did I miss something here?

weynhamz:
Fix #4051


danepowell:
Thanks for the PR!

PR #4057, opened 2020-03-16
Remove BLT Project subtree split

danepowell:
BLT Project will continue to exist as a standalone repo but will not be managed concurrently with BLT. Because of idiosyncrasies of ORCA, Composer path repos, and Git subtrees, it was impossible to test changes to BLT Project concurrently with BLT anyway, so I don't feel that we lose much by just killing the subtree split. If anything, it better reflects the reality of the architecture and testing situation.

https://github.com/acquia/blt-project


danepowell:
Depends on #4048

PR #4061, opened 2020-03-18
Replace factory-hooks/post-install by post-site-install

rmuit:
Changes proposed
---------
The Site Factory 'post-install' hook was implemented as PHP scripts that are being run through 'drush scr'. (And its public exposure was never completely thoroughly coordinated among various department, it kinda just made its way into public docs at one point.)

As of today, Site Factory has a 'post-site-install' hook, which runs executable scripts like other hooks (e.g. db-update) do. Also, logging and timeout handling of this hook is unified with other hooks / slightly better than the old one.

Because the hook is implemented on the platform side, this is irrespective of BLT version; every (supported) version could have this PR applied without side effects or differing testing results. I've checked through older versions, and the file is the same in every BLT version except for PHPCS hints (which are not applicable for the new version).

There is no timeline for fully removing the 'post-install' hook, but we'd still love it if everyone was able to update their installs to the new hook.

-----------
* The new file has the 'executable' bit(s) set.
* After site install, the blt command should still be executed just like it is before. (This is the only thing that the existing post-install hook script does, effectively.)
* Behavior change: if the blt command exits non-zero, the script also does. (The effect on the Site Factory platform is that it will send an e-mail about something having gone wrong with the post-site-install process. The site itself will still be installed and online - because this hook is only run after a successful install. I believe that's an acceptable / non-dangerous behavior change, and in fact we want this.)

_Note I have not done tests myself yet, I'm open to how this should be tested but for now I'm just adding the changed code._

Additional details
-----------
While checking around, I saw that db-update/db-update.sh and post-install/post-install.php essentially do the same thing: run 'blt drupal:update'. I took db-update.sh for inspiration to create a shell script. (Better unification with existing code, though bigger diff).

A lot of code in the old post-install.php revolves around deriving the "preferred domain" and then the site name. All that is simplified because the (internal / the only one available at install time) domain is passed into the script. So that code is all removed.


scripts/factory-hooks/post-site-install/post-site-install.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Print to Site Factory SiteInstall task log.
+echo "Executing: DRUSH_PATHS_CACHE_DIRECTORY=$cache_dir $blt drupal:update --environment=$env --site=${name[0]} --define drush.uri=$internal_domain --verbose --no-interaction"
+
+# Execute the updates.
+DRUSH_PATHS_CACHE_DIRECTORY="$cache_dir" $blt drupal:update --environment=$env --site=${name[0]} --define drush.uri=$internal_domain --verbose --no-interaction
+result=$?
+
+# Clean up the drush cache directory.
+rm -rf "$cache_dir"
```
rmuit: This is not in db-update.sh but it is in the old post-install.php; let's keep it.

scripts/factory-hooks/post-site-install/post-site-install.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+#
+# Factory Hook: post-site-install
+#
+# This is used so that an ACSF site install is identical to the local BLT site
+# install, with the environment, site, and uri CLI runtime arguments overriding
+# all other configuration.
```
rmuit: I can't completely parse what this paragraph means so I kept it.

scripts/factory-hooks/post-site-install/post-site-install.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# Usage: SCRIPTNAME site env db-role domain
+
+# Exit immediately on error and enable verbose log output.
+set -ev
```
rmuit: 'set -e' is not doing anything in this script. Then again 1) it isn't doing anything in db-update.sh either 2) it feels like a good thing to do in general. So.

scripts/factory-hooks/post-site-install/post-site-install.sh
```diff
@@ -0,0 +1,54 @@
...
+# To get only the site name in ${name[0]}:
+IFS='.' read -a name <<< $internal_domain
+
+# BLT executable:
+blt="/mnt/www/html/$sitegroup.$env/vendor/acquia/blt/bin/blt"
+
+# Create and set Drush cache to unique local temporary storage per site.
+# This approach isolates drush processes to completely avoid race conditions
+# that persist after initial attempts at addressing in BLT: https://github.com/acquia/blt/pull/2922
+cache_dir=`/usr/bin/env php /mnt/www/html/$sitegroup.$env/vendor/acquia/blt/scripts/blt/drush/cache.php $sitegroup $env $internal_domain`
```
rmuit: Deleted the mkdir which was in post-install.php (and not in db-update.sh) because cache.php itself creates the directory.

scripts/factory-hooks/post-site-install/post-site-install.sh
```diff
@@ -0,0 +1,54 @@
...
+# Clean up the drush cache directory.
+rm -rf "$cache_dir"
+
+set +v
+
+# Exit with the status of the BLT commmand. If the exit status is non-zero,
+# Site Factory will send a notification of a partiolly failed install and will
+# stop executing any further post-site-install hook scripts that would be in
+# this directory (and get executed in alphabetical order).
+exit $result
```
rmuit: There are a few changes I couldn't resist making relative to db-update.sh, and/or didn't know whether I should make, it's fairly subjective:
* use snake_case for bash variables ($cache_dir not $cacheDir) because I've never seen camelCase in bash before.
* convention seems to be to uppercase bash variables, actually. But I don't know if I should - I just kept it equal to db-update.sh
* quote all variables? Officially yes. But none of the inputs ever contain spaces and things become less readable and db-update.sh does _mostly not_ do it (though it isn't consistent). I did quote $cache_dir because I feel icky _not_ doing it with a variable that is used in a 'rm -rf' command. (And then I'm doing it consistently throughout the script.)

rmuit:
> Also, logging and timeout handling of this hook is unified with other hooks / slightly better than the old one.

I should state this more strongly. There have been several customers in the past year who have suffered from timeout issues of the old post-install hooks, when the blt drupal:update command was running 'too long'. (Read: an arbitrary timeout of around 4 minutes that should never have been in place.) @lcatlett's customer was suffering from this presently. So this change is actually a definite fix/improvement/help for some customers.

rmuit:
Update: the PR has been rebased on 11.x now, following a PR from @lcatlett to my PR, to consolidate the earlier changes, which are now rolled into one.

Summary of changes to the PR (i.e. the diff from @lcatlett's first commit onwards):

* The post-site-install hook has had the 'header comment' changed. The rest is unchanged, so IMHO we can consider this file reviewed and tested by @lcatlett (who has had #4070 running in production, which is essentially the same, and fixed issues with it).

* An update hook for BLT 11 was added from #4070. (I did not test it myself because of lack of BLT experience).

* The db-update hook, which I hadn't touched in my original PR, has now had changes of comments / placing / variables names to unify with this new post-site-install hook, reviewed by me and I don't see any behavior change. I'd like to draw your attention to a TODO added to the very bottom of the file: it would _likely_ be a good idea to change the exit code of the script sometime in the future, I agree... but that's a behavior change for existing customers so the actual change is kept out of this PR.

danepowell:
Thanks a lot for the PR, I manually tested it and everything seems to be working great!

PR #4064, opened 2020-03-19
DT-1201: Run BLT tests with ORCA fixture

danepowell:
- It would be nice not to have to hack the symlink, but I don't know how to avoid that: https://github.com/composer/composer/issues/8700
- Update testing docs (against docs repo first). 

Testing steps:
- Remove the vendor directory from the BLT working directory to avoid an autoloading issue: https://github.com/acquia/orca/issues/73
- Create the test fixture: `/home/dane/packages/orca/bin/orca fixture:init -f --sut=acquia/blt --sut-only --core='~9' --dev`
- Run the tests: `ORCA_FIXTURE_DIR=/home/dane/packages/orca-build ./bin/orca qa:automated-tests --sut=acquia/blt --sut-only`


danepowell:
Ah... the problem now is that the Doctor (and possibly other components) expect MySQL and don't know what to do with SQLite.

danepowell:
Jeez... now blocked on an undisclosed massive breaking change in a Symfony Console patch release: https://github.com/symfony/symfony/pull/36027

PR #4070, opened 2020-03-20
Resolve ACSF site install breakages

lcatlett:
It seems there was a breaking change in a recent ACSF [platform update](https://docs.acquia.com/site-factory/release-notes/#acquia-cloud-site-factory-2-100-march-18-2020) when the post-install factory hook was replaced with a post-site-install hook. Since the post-install hooks that run `blt drupal:update` tasks are no longer executing (though they are being recognized), creating a site on my BLT / ACSF project is resulting in a very broken site. 

Changes proposed
---------
This PR: 
- adds the needed post-site-install hook 
- removes the deprecated post-install scripts
- adds a blt update hook to automatically re-generate factory hooks for projects using the previous templates
- adds minor docs updates and fixes

Steps to replicate the issue
----------
1. Create a site on ACSF using any version of BLT
2. Observe that site is only partially installed and functionality relying on BLT tasks like importing environment splits fail to run

-----------
Creating a site on ACSF successfully runs `blt drupal:update` tasks in the `post-site-install` hook


mikemadison13:
@lcatlett this looks like a dup of @rmuit's open PR here: https://github.com/acquia/blt/pull/4061. we should try to consolidate between the two

rmuit:
@lcatlett  there was *no breaking change* in the last ACSF platform update. post-install hooks are still being executed exactly the same way, you hint at this in your own message. So I highly doubt that your 'steps to reproduce' are generally applicable to many customers.

Regardless, 
* moving over is a good thing because post-install hooks have always been more vulnerable to e.g. timeouts of long running processes, than other hooks have been. (Maybe you happened to hit a timeout at the same time?)
* thank you for the addition of the update hook.

rmuit:
As far as consolidation goes:
* The addition to the db-update.sh is good
* The addition of the update hook is necessary  / shows my lack of BLT expertise. (Will that need changing over BLT versions?)
* This comment in post-site-install is undoubtedly better than mine (which I kept from the oldo one):
```
# This is necessary so that blt drupal:install tasks are invoked automatically
# when a site is created on ACSF. 
```
* Passing the domain through uri.php is not necessary (and therefore discouraged), since it's always going to be the internal domain at the post-site-install stage so we can derive the sitename directly. Also, exiting with the exit code of the blt command is a good thing (from the old script) to keep. So IMHO we can take my version of the script, with the replacement of  the above comment in the header.

lcatlett:
@mikemadison13 and @rmuit - apologies for the overlapping PRs; I totally missed #4061. @rmuit - I put in a PR to your fork consolidating the two and implementing your suggestions: https://github.com/rmuit/blt/pull/1. Once you merge it will update #4601 so closing this PR

rmuit:
@lcatlett thanks. But
* I'll need to rebase my PR (which is against 12.x) on 11.x because it is now version specific (which is a good thing!)
* This changes too much because the uri.php handling _is_ necessary in db-update.sh. Also I am hesitant to introduce behavior changes to db-update.sh into this PR. (Even though the exit code change _feels_ like a good thing, if you want to propose that change IMHO a separate PR should be opened.) I'm fine with the comment or capitalization changes.

Since I need to rebase the PR anyway, I'll take over your commits but back out some changes by monday morning if nothing else happens.

(And to close off the initial assertion in this PR's description: there was _no_ breaking change - because your customer's post-install hooks have been behaving the same for weeks, before and after the ACSF update. That is: they've been failing.
Apparently this was related to timeouts / a long running blt command, which is indeed an issue with the old post-install hooks and the reason why we added the post-site-install hooks. So your change from post-install to post-site-install could fix the weeks-old issue.)

lcatlett:
@rmuit - the modifications to uri.php / $domain logic were made in error so I've updated the PR to back out those modifications and address your other feedback

PR #4073, opened 2020-03-24
Public/Private File Sync Options Aren't Respected In "drupal:sync:all-sites"

yyss8:
Fixes # 
--------
- Allows options `sync.public-files` and `sync.private-files` to be respected in the command `drupal:sync:all-sites`

Changes proposed
---------
- Currently, running command like `blt drupal:sync:all-sites -D sync.public-files` won't trigger the command to sync down the files. I believe it's because of the configurations are being re-initialized by the `switchSiteContext` method.
image

Since the options aren't being passed to the method `sync` at line 33 of file `src/Robo/Commands/Drupal/SyncCommand.php` and the value of those two options won't exist in the current configurations, the command `blt drupal:sync:all-sites -D sync.public-files` will never try to sync down the files.

Steps to replicate the issue
----------
1. Run `blt drupal:sync:all-sites -D sync.public-files` and see if the public or private files are synced.

----------
- The file syncing commands are not running when the options `sync.public-files`/`sync.private-files` are presented.

-----------
Passing the options `sync.public-files`/`sync.private-files` in the command `blt drupal:sync:all-sites` will allow files to be downloaded (synced).

Additional details
-----------
I remember the command was working without making such changes, perhaps the value of those two options are lost somehow during configuration re-initializing in `switchSiteContext` due to some other changes?


danepowell:
My guess is that no CLI config params are getting set when switching site context, but we can follow up with a more general fix later. Thanks for the detailed report and PR!

PR #4077, opened 2020-03-26
Move phpcs to standalone plugin.

### LINKED ISSUE #4024: Unexclude the NamespacedDependency sniff.
Thought it might be appropriate to revisit #2295.

It's a bit odd to ignore this one single code sniff. Obviously BLT projects have the ability to add this if they desire, or any other code sniff, so maybe we should not exclude it by default.

Several of the reasons given 2 years ago have been addressed in Core.

This isn't really blocking anything for me, b/c I can just exclude the BLT default. Just thought I would raise it.

danepowell:
Fixes #4024 
--------

Changes proposed
---------
- Move phpcs validation to a standalone plugin: https://github.com/acquia/blt-phpcs (not yet published)

For folks creating blt projects using blt-project, nothing changes because we'll install that plugin automatically: https://github.com/acquia/blt-project/commit/ea00b47c9469f77ddbc8db41474c329c3b8dcd67

But there are several advantages to this new approach:
1. Folks adding BLT to an existing project get a template phpunit.xml.dist if they'd like
2. You can remove this if you don't want it.
3. phpcs validation can be added to require-dev so that it doesn't get shipped to production.


danepowell:
I suspect the problem is that our test cases don't actually shell out to run BLT commands... they build a new Blt/Robo container within the PHPUnit process, and don't pass in the autoloader. And then Robo can't discover the commands because there is no autoloader and hence no registered PSR4 namespaces (I confirmed the PSR4 namespaces are empty).

PR #4080, opened 2020-03-27
Remove custom sandbox manager.

danepowell:
Just a POC to confirm this works on Travis, as it does locally. If so, I'll clean it up.


tests/phpunit/src/BltProjectTestBase.php
```diff
@@ -92,9 +92,8 @@ public function setUp(): void {
     $this->printTestName();
     $this->bltDirectory = realpath(dirname(__FILE__) . '/../../../');
     $this->fs = new Filesystem();
-    $this->sandboxManager = new SandboxManager();
-    $this->sandboxManager->replaceSandboxInstance();
-    $this->sandboxInstance = $this->sandboxManager->getSandboxInstance();
+    $this->execute('./bin/orca fixture:reset -f', getenv('ORCA_ROOT'));
+    $this->sandboxInstance = getenv('ORCA_FIXTURE_DIR');
```
danepowell: @TravisCarden we talked about abstracting this functionality into a `OrcaFixtureTestCase` base class that downstream projects like BLT could inherit to have their test fixture automatically reset for every test case.

For this to work, `OrcaFixtureTestCase` needs a process runner that's aware of ORCA_ROOT and ORCA_FIXTURE_DIR. But since it's being instantiated by PHPUnit and not by ORCA itself, we can't inject a process runner like you do everywhere else in ORCA.

Any creative ideas on how to handle that?

danepowell:
This cuts our test run times in _half_! (From ~8 minutes to 4 minutes).

PR #4085, opened 2020-03-30
Cache environment detectors.

### LINKED ISSUE #4084: Cache overridden environment detectors in `EnvironmentDetector::getSubclassResults()`
I was profiling a site and noticed some 10000 calls to the closure in `EnvironmentDetector::getSubclassResults()`. It's re-running the detection on each call (about ~13 locally, which multiplied by some ~700 results in the class map, works out to the ~10k calls to the closure.

It would be good to at least cache the initial detectors found statically. On sites w/o an overridden detector, that would drop the closure calls to however many classes exist in the map.

jhedstrom:
- Closes #4084.


danepowell:
Great catch, and thanks for the PR!

PR #4088, opened 2020-03-31
DT-1399: Evaluate Composer 2.0 readiness.

danepowell:
Known blockers:

- [x] https://github.com/composer/installers/issues/447
- [ ] https://github.com/Dealerdirect/phpcodesniffer-composer-installer/issues/108
- [x] https://github.com/cweagans/composer-patches/pull/309
- [x] https://github.com/grasmash/drupal-security-warning/pull/8
- [x] https://github.com/zaporylie/composer-drupal-optimizations/issues/16 (just remove this plugin)
- [x] https://github.com/mglaman/drupal-check/issues/160
- [x] https://www.drupal.org/project/drupal/issues/3126566


composer.json
```diff
@@ -19,27 +19,24 @@
         "php": ">=7.3",
         "ext-json": "*",
         "acquia/drupal-environment-detector": "*",
-        "composer-plugin-api": "^1.1.0",
+        "composer-plugin-api": "^1.1.0 || ^2.0",
         "composer/semver": "^1.4",
         "consolidation/comments": "^1.0",
         "consolidation/config": "^1.0.0",
         "consolidation/robo": "^1.4.12 || ^2",
-        "cweagans/composer-patches": "^1.6.5",
```
wimleers, line 27: This is a huge BC break and caused more than a dozen of hours of time spent debugging https://github.com/acquia/acquia_migrate/pull/185 😭

danepowell: Sorry to cause you grief, but just to be clear this only happened in the 12.x branch which is explicitly _unstable_ and _unsupported_. However we just cut 12.0.0-alpha1 and I don't foresee any more BC breaks, although they could still happen.

danepowell: Also I suppose this is an example of the paradigm that if you rely on a dependency, you should declare it. You might want to see what else BLT is providing that you rely on, and make sure that you explicitly declare the dependency yourself.

wimleers: > but just to be clear this only happened in the 12.x branch which is explicitly unstable and unsupported

But we're explicitly building a _stable_ product on top of Drupal 9 over at https://github.com/acquia/migrate.

We _must_ use https://github.com/acquia/orca.

https://github.com/acquia/orca requires us to use https://github.com/acquia/blt, and specifically version 12 to test with Drupal 9.

So I don't see how we have any choice here?

I think we can be pragmatic here: explicitly ping me whenever there's a BC break being introduced in BLT 12, so that we can make whatever changes we need to not break the build, or only break it for a brief period of time :)

wimleers: > Also I suppose this is an example of the paradigm that if you rely on a dependency, you should declare it, even if you already get it "for free" from some upstream package. 

This is totally fair! :)

wimleers: FYI: https://backlog.acquia.com/browse/OCTO-3275

PR #4106, opened 2020-04-08
Sets initializer to use docroot from already processed files.

### LINKED ISSUE #4105: Docroot for multi-site detection is always default.
When using mult-site, the config to determine `docroot` is not using the already processed values from the project blt.yml file, meaning sites blt.yml will always look inside `/docroot` even if set otherwise in the project.

b-sharpe:
Fixes #4105
--------

Changes proposed
---------
Change the config initializer to use the already processed config to determine docroot rather than just the default. 

Steps to replicate the issue
----------
1. Change docroot in your project blt.yml
2. Attempt to run a site command with overridden config in <other docroot>/sites/<site>/blt.yml
3. Site yml will not be found/used

### DIFF
--- a/src/Robo/Config/ConfigInitializer.php
+++ b/src/Robo/Config/ConfigInitializer.php
@@ -162,6 +162,8 @@ class ConfigInitializer {
    */
   public function loadSiteConfig() {
     if ($this->site) {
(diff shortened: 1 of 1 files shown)


danepowell:
Makes sense, thanks for the PR!

PR #4130, opened 2020-04-22
DT-1515: Move Acquia Environment Detector to standalone package.

danepowell:
(not yet published) https://github.com/acquia/environment-detector

This would allow customers and other consumers to use BLT's environment detection capabilities without the rest of the overhead of BLT.


wimleers:
Very nice! 👏

PR #4138, opened 2020-04-23
Fixes #4136: Dynamically run all tests/validate commands in namespace.

### LINKED ISSUE #4136: Test and Validate Commands Fail in BLT 12
**Describe the bug**
Given that Behat and PHPCS have been moved to plugins (given https://github.com/acquia/blt/issues/3928 and https://github.com/acquia/blt/pull/4077) we need to update https://github.com/acquia/blt/blob/12.x/src/Robo/Commands/Tests/TestsAllCommand.php and https://github.com/acquia/blt/blob/12.x/src/Robo/Commands/Tests/ValidateAllCommand.php to not run behat by default (currently all execution of the `blt tests` command in BLT 12 is broken due to missing Behat). We need to do the same for PHPCS.

**To Reproduce**
1. Create a blt 12 project
2. Run composer install
3. Run blt tests and/or blt validate without adding the phpcs / behat plugins
4. See errors

**Expected behavior**
I would expect that one of a few things happens:

- Either we keep the commands in the tests/validate commands (but somehow figure out if the plugin is present) OR
- Move the execution of these commands into their respective plugins and somehow subclass the tests command like we have with Environment Detector. 

**Detailed error output**
```
blt tests
[error]  Command "tests:behat:run" is not defined.

Did you mean one of these?
    setup:behat
    tests:behat:init:config
    tests:frontend:run 
```

```
blt validate
> tests:composer:validate
Validating composer.json and composer.lock...
[ExecStack] composer validate --no-check-all --ansi
[ExecStack] Running composer validate --no-check-all --ansi in /Users/mike.madison/git/canary9
./composer.json is valid
[ExecStack] Done in 0.371s
> tests:php:lint
Linting PHP files...
[error]  There are no commands defined in the "tests:phpcs:sniff" namespace. 
```

**Additional context**
For some of the testing / validation we do it based on config (see https://github.com/acquia/blt/blob/12.x/src/Robo/Commands/Tests/ValidateAllCommand.php#L33) but I'm not sure that is the best option here. It seems that doing the subclass results like we do in the Environment Detector (https://github.com/acquia/blt/blob/12.x/src/Robo/Common/EnvironmentDetector.php#L122) would be the best option?

danepowell:
Fixes #4136 
--------

Changes proposed
---------
- Move validation commands out of the top-level `tests` namespace and into a new `validate` namespace (not sure why we lumped them all under "tests" in the first place)
- Add an `invokeNamespace()` method, which will search for and run all second-level commands in a namespace (`validate:composer`, `validate:phpcs`, etc...)
- Use this method for `validate` and `tests` commands so they automatically invoke plugin-provided commands as well.

Must be tested and merged in conjunction with https://github.com/acquia/blt-phpcs/pull/1


danepowell:
Actually I am going to refactor this to do it properly...

danepowell:
This works now in my testing of `blt validate` and `blt tests`, both with and without plugins such as `blt-phpcs` installed. Note that you might not be able to even install `blt-behat` at the moment since Drupal Extension conflicts with Drupal 9.

mikemadison13:
i'll patch both blt and blt-phpcs to give this a whirl. I assume that the acsf validate option is gone because it moved into the acsf plugin?

mikemadison13:
Tests is passing, validate is failing:

```
[error]  There are no commands defined in the "tests:composer" namespace.

Did you mean one of these?
    setup:composer
    tests
    tests:frontend
    tests:server 
```

I wonder if this is because of: https://github.com/acquia/blt/blob/d29510891fa319f3e5caaec8d51ac85c65c70981/src/Robo/Commands/Validate/ComposerCommand.php#L9

danepowell:
I fixed the comment but I don't think that's functional. It's not part of the command annotation. What's the full command and output?

I suspect you aren't completely using this feature branch or something. Maybe nuke and reinstall composer dependencies?

mikemadison13:
Believe it or not (and digging into this) the issue is actually that the `blt validate` command is somehow picking up the contents of the Validate Test?! When I update https://github.com/acquia/blt/blob/12.x/src/Robo/Commands/Tests/ValidateAllCommand.php#L21 to be `validate:composer` instead of `tests:composer:validate` it seems to run fine. I'm wondering if something in the way you're getting the sub-results here is accidentally picking up the tests as if they were real commands?

PR #4168, opened 2020-06-08
Switch to acquia/drupal-recommended-project

danepowell:
- [x] Fixes https://github.com/acquia/blt-vm/issues/5, close that with this as well


danepowell:
I think the remaining problem here is that ORCA runs git clean -df, which kills the docroot/sites/all directory because docroot/sites/all/files is not gitignored in acquia/drupal-recommended-project, as it was in blt-project.

Another more serious question is why the BLT tests are sensitive to that. It seems that they have traditionally keyed off "all" as a valid multisite, which probably isn't good. And when they use "default" (which is probably actually _right_) everything falls apart, seemingly because Drush interprets "default" as a valid URI and doesn't read the actual URI from local.drush.yml.

danepowell:
depends on https://github.com/acquia/drupal-recommended-project/pull/4

danepowell:
This is pulling in the new project template version, but throwing the same error...

PR #4193, opened 2020-06-19
BLT-4192: Fix validate twig syntax with Drupal 9.

### LINKED ISSUE #4192: validate:twig not working anymore with core twig filter
**Describe the bug**
The validate:twig command is not working anymore with the default Twig filters from Drupal core. I have errors for all default twig filters or functions like t, clean_class, without, render and so on...

**To Reproduce**
You need to have some custom templates using default twig filters from Drupal and run the command validate:twig.

**Expected behavior**
The command validate:twig should not throw errors for default twig filters. It should be considered as a valid syntax.

**System information**
* Drupal VM
* Drupal core 9.0.1
* BLT version: 12.0.1

**Proposed resolution**
The syntax in /core/lib/Drupal/Core/Template/TwigExtension.php has changed and the new filters are now declared with "new TwigFilter" instead of "new \Twig_SimpleFilter". Same thing for TwigFunction.
We need to update the file /src/Robo/Commands/Validate/TwigCommand.php to reflect those changes.

agiraud85:
Fixes #4192  
--------

Changes proposed
---------
The syntax in /core/lib/Drupal/Core/Template/TwigExtension.php has changed and the new filters are now declared with "new TwigFilter" instead of "new \Twig_SimpleFilter". Same thing for TwigFunction.
We need to update the file /src/Robo/Commands/Validate/TwigCommand.php to reflect those changes.

Steps to replicate the issue
----------
1. Install new BLT 12 project with Drupal 9
2. Add custom twig templates to use core twig filters
3. Run blt validate:twig function.

----------
The TwigCommand.php is searching for "new \Twig_SimpleFilter" in Drupal core to add exception but these filters don't exist anymore.

-----------
Run the command blt validate:twig and it should contain valid syntax.

### DIFF
--- a/src/Robo/Commands/Validate/TwigCommand.php
+++ b/src/Robo/Commands/Validate/TwigCommand.php
@@ -9,6 +9,8 @@ use Symfony\Bridge\Twig\Command\LintCommand as TwigLintCommand;
 use Symfony\Component\Console\Input\ArrayInput;
 use Twig\Environment;
 use Twig\Loader\FilesystemLoader;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
 
 /**
  * Defines commands in the "validate:twig:lint:all*" namespace.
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks a lot for the PR!

PR #4214, opened 2020-07-27
4213: Use the provided URI in the drush site alias

### LINKED ISSUE #4213: Multisite recipe should use actual site URL in drush alias

When I set up a multisite with `blt recipes:multisite:init`, it prompts me for a "Local domain name". This is used in the local `blt.yml` but not in the drush alias file.

Am I the only one who is tired of

1. `drush @mysite uli`
1. Copy everything after "http://default".
1. Paste into the location bar of my browser.

Update `createSiteDrushAlias()` so that it accepts a second, optional parameter.

See above.

**Additional context**

benjifisher:
Fixes #4213 
--------

Changes proposed
---------

- Add an optional parameter to `MultisiteCommand::createSiteDrushAlias()`
- Pass the URI to `MultisiteCommand::createSiteDrushAlias()` from ``MultisiteCommand::generate()`

Steps to replicate the issue
----------
1. `blt recipes:multisite:init`
1. At the prompt `Site machine name (e.g. 'example')`, answer `mysite`.
1. At the prompt `Local domain name [http://local.multi1.com]`, answer `https://mysite.lndo.site`.
1. Once the site is running, try `drush @mysite status` or `drush @mysite uli`.

----------

Drush uses `http://mysite`.

-----------

Drush uses `https://mysite.lndo.site`.

Additional details
-----------

I believe that the `uri` specified in `drush/sites/mysite.yml` overrides the setting in `sites/mysite/local.drush.yml`.

### DIFF
--- a/src/Robo/Commands/Recipes/MultisiteCommand.php
+++ b/src/Robo/Commands/Recipes/MultisiteCommand.php
@@ -57,7 +57,7 @@ class MultisiteCommand extends BltTasks {
     $local_alias = $this->getNewSiteAlias($site_name, $options, 'local');
     $this->createNewBltSiteYml($new_site_dir, $site_name, $url, $local_alias, $remote_alias, $newDBSettings);
     $this->createNewSiteConfigDir($site_name);
-    $this->createSiteDrushAlias($site_name);
+    $this->createSiteDrushAlias($site_name, $domain);
     $this->resetMultisiteConfig();
 
     $this->invokeCommand('blt:init:settings');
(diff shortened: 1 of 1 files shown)


danepowell:
Makes sense, thanks! I'm not sure why it was created the way it was. I'm a little worried about how this might interact with sites.php, I think the convention for multisites was to use e.g. `--uri=foo` instead of `--uri=local.foo.com`, but as long as this only modifies the local alias it seems alright. I'm a little rusty on all of this.

PR #4219, opened 2020-07-29
4218: Add an option to bypass installing the shell alias

### LINKED ISSUE #4218: Make it possible to skip the BLT alias

I am running BLT from a Docker container, using Lando. I have told Lando to use `/app/vendor/bin/blt` as the executable, so I do not need an alias. Yet I still get nuisance warnings:

```bash
$ lando blt install
...
> blt:init:shell-alias
[warning] Could not find your CLI configuration file.
[warning] Looked in ~/.zsh, ~/.bash_profile, ~/.bashrc, ~/.profile, and ~/.functions.
[warning] Please create one of the aforementioned files, or create the BLT alias manually.
```

Add an option to `blt/blt.yml` that lets me skip the `blt:init:shell-alias` task at the end of `blt setup`.

I could add `~/.functions` or similar to the Docker container, but that seems like fixing the problem in the wrong place. Besides, I still get one line of success noise (green) instead of a few lines of warning noise (yellow).

**Additional context**

I plan to submit a PR soon.

benjifisher:
Fixes #4218 
--------

Changes proposed
---------

- Add a new option, `blt.alias.bypass`, to `config/build.yml`, with default value `false`.
- Check for this option at the start of `installBltAlias()` and exit early if it is set.
- The change will have no effect on users unless they choose to override the new option in `blt/blt.yml`.

Steps to replicate the issue
----------
1. Without installing the alias, run the `setup` task. For example, `vendor/bin/blt/setup`.
2. Note the warning message at the end of the task.

----------

Noise:

```bash
> blt:init:shell-alias
[warning] Could not find your CLI configuration file.
[warning] Looked in ~/.zsh, ~/.bash_profile, ~/.bashrc, ~/.profile, and ~/.functions.
[warning] Please create one of the aforementioned files, or create the BLT alias manually.
```

-----------

Much less noise:

```bash
> blt:init:shell-alias
```

Additional details
-----------

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -26,6 +26,7 @@ blt:
   command-cache-dir: ${blt.root}/cache/commands
   alias:
     auto-install: true
+    bypass: false
 
 composer:
   bin: ${repo.root}/${bin.path}
--- a/src/Robo/Commands/Blt/ShellAliasCommand.php
+++ b/src/Robo/Commands/Blt/ShellAliasCommand.php
@@ -20,6 +20,9 @@ class ShellAliasCommand extends BltTasks {
    * @aliases blt:init:shell-alias alias install-alias
    */
   public function installBltAlias() {
+    if ($this->getConfigValue('blt.alias.bypass')) {
+      return;
+    }
     if (isset($_SERVER['ComSpec'])) {
       $bltRoot = $this->getConfigValue('blt.root') . '\\vendor\\bin';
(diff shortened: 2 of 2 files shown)


danepowell:
Thanks for the PR, but BLT already has a standard way of disabling arbitrary commands so we don't need to hardcode support for them individually. Let's discuss in #4218.

PR #4230, opened 2020-08-25
Fixed #4229: Use the correct check before munging database name

### LINKED ISSUE #4229: BLT 12 checks isAcsfInited() instead of isAcsfEnv() in default.local.settings.php
**Describe the bug**

With Drupal 9 and BLT 12, I have recently run `blt recipes:acsf:init:all` and committed the results. Now, when I run `blt setup -n`, I see the following:

```bash
blt setup -n
...
> internal:drupal:install
[Acquia\Blt\Robo\Tasks\DrushTask] Running /app/vendor/bin/drush @self site-install myprofile install_configure_form.enable_update_status_module=NULL --sites-subdir=default --site-name='My ACSF Implementation' --site-mail=no-reply@example.com --account-name=3nwOL3QIoj --account-mail=no-reply@example.com --locale=en --no-interaction -v --ansi in /app/docroot
 [info] Executing: command -v mysql
 [info] Executing: mysql --defaults-file=/tmp/drush_INe98z --database=drupal9_ --host=database --port=3306 --silent -A < /tmp/drush_mglprV
...
```
The database name should be `drupal9`, not `drupal9_`. This leads to errors a few lines later, when BLT tries to invoke `drush site:install`.

**Expected behavior**

The `blt setup` command should execute without errors, setting up a local site with the default profle.

**System information**
* Operating system type: Linux
* Operating system version: Ubunto 20.04
* BLT version: 12.x-dev

**Additional context**

The unwanted `_` character comes from these lines in `default.local.settings.php`, which  were added in 5a4aa878a as part of the fix for Issue #3983: 

```php
$db_name = '${drupal.db.database}';
if (EnvironmentDetector::isAcsfInited()) {
  $db_name .= '_' . EnvironmentDetector::getAcsfDbName();
}
```

After `recipes:acsf:init:all`, the test `isAcsfInited()` returns TRUE, and the underscore is added to the database name.

A minimal fix is to replace `isAcsfInited()` with `isAcsfEnv()`. I will submit a PR that does this.

If I understand correctly, the BLT strategy for settings files is that the `local` version should only be used in local environments, not CI and not on Acquia Cloud. If so, then a better fix would be to remove that entire `if` clause.

benjifisher:
Fixes #4229 
--------

Motivation
----------

After `blt recipes:acsf:init:all`, the command `blt setup` fails.

Proposed changes
---------
What does this PR change? How does this impact end users? Are manual or automatic updates required?

This PR checks `isAcsfEnv()` instead of `isAcsfInited()` so that ACSF-specific code is not run on local development environments.

Alternatives considered
---------

Testing steps
---------

```bash
blt recipes:acsf:init:all
blt setup -n
```

### DIFF
--- a/settings/default.local.settings.php
+++ b/settings/default.local.settings.php
@@ -9,7 +9,7 @@ use Acquia\Blt\Robo\Common\EnvironmentDetector;
 use Drupal\Component\Assertion\Handle;
 
 $db_name = '${drupal.db.database}';
-if (EnvironmentDetector::isAcsfInited()) {
+if (EnvironmentDetector::isAcsfEnv()) {
   $db_name .= '_' . EnvironmentDetector::getAcsfDbName();
 }
 


danepowell:
This is a local settings file, so it should never be loaded in a hosted environment, therefore the conditional would always return false.

The intent of the conditional (to check if the codebase is initialized for ACSF) I believe is correct, the error must lie somewhere else.

PR #4239, opened 2020-09-02
DX-2267: Restore local.blt.yml generation

### LINKED ISSUE #4231: local.blt.yml does not get autogenerated in BLT v12
**Describe the bug**
when running blt setup (first time setup on a newly forked project) local.blt.yml does not get generated using BLT v12 with Drupal 9.  You have to manually create the local.blt.yml file from the example.local.blt.yml file in order for blt setup to work. This bug appears to have been introduced with https://github.com/acquia/blt/pull/4140/files.

**To Reproduce**
1. Clone https://github.com/Drupal4Gov/drupal4gov.us (not 100% sure if `composer create-project acquia/drupal-recommended-project --no-interaction` is setup for blt 12)
2. run blt setup
3. cd to blt and ll
4. no local.blt.yml file is present

**Expected behavior**
BLT setup on first time setup should create local.blt.yml for me.  If the local.blt.yml file is removed then local.blt.yml should be created on blt sync

**System information**
* Operating system type: Mac OS
* Operating system version: 10.15.6 (19G73)
* BLT version: 12.1.0

danepowell:
Fixes #4231 
--------

This should create `local.blt.yml` if `example.local.blt.yml` exists.

Note that the new project templates no longer provide `example.local.blt.yml` out of the box, you'd need to manually create it if it's missing. I could see an argument for BLT to automatically create `example.local.blt.yml` if it's missing, or on composer install, but that should be a follow-up ticket.

@mikemadison13 can you help test this? I'd like to cut a new BLT release soon, and include this fix.

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -19,7 +19,7 @@ For major releases, coordinate with ORCA prior to starting this process to ensur
 
 ## Review issue labels
 
-Review the issues and pull requests that will make up the release and ensure they are labelled correctly so that the changelog and release notes will be generated with the correct categories (new features, bug fixes, etc...).
+Review the issues and pull requests that will make up the release and ensure they are labelled correctly so that the changelog and release notes will be generated with the correct categories (new features, bug fixes, etc...). [Github releases](https://github.com/acquia/blt/releases) have a link to display all commits since a given release.
 
 ## Test via Canary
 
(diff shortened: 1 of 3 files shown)


danepowell:
I tested manually and confirmed this generates the file as expected.

PR #4248, opened 2020-09-17
Fixes #4247: defaults the local.drush hostname to self.

### LINKED ISSUE #4247: BLT 12 Issues w/ Default Local Alias
This is a rewrite of #4242 in which I articulated a problem but wrongly identified the root cause. So I'm starting over clean here so we have something sensible (and not as backwards). 

**Describe the bug**
After spinning up a new BLT 12 project and spinning up Drupal VM, I can run blt setup:drupal:install w/o config, but as soon as I add config and the re-run, the config import fails because of BLT's default config for local drush alias.

In the past, we have defaulted to self as the default but now we have local.${project.machine_name}.com.

This appears to be "properly" configured in BLT's config, but the stock drush file is wrong. see https://github.com/acquia/blt/blob/12.x/settings/default.local.drush.yml

**To Reproduce**
1. Run composer create-project acquia/drupal-recommended-project --no-interaction
2. Add drupal vm and setup a box
3. SSH into the container
4. Run blt setup:drupal:install -n
5. export configuration w/ drush cex -y
6. re-run blt setup:drupal:install -n

**Expected behavior**

**Detailed error output**
```
blt drupal:update
[Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/scenario/vendor/bin/drush @self updb --no-interaction --ansi in /var/www/scenario/docroot

In Request.php line 1183:
                                                     
  Invalid Host "local.${project.machine_name}.com".  
                                                     

updatedb [--cache-clear [CACHE-CLEAR]] [--entity-updates] [--post-updates [POST-UPDATES]] [--no-cache-clear] [--no-post-updates] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--no] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--notify [NOTIFY]] [--druplicon] [--xh-link XH-LINK] [--] <command>

[Acquia\Blt\Robo\Tasks\DrushTask]  Exit code 1  Time 0.256s
[error]  Failed to execute database updates!
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 
```

**System information**
* Operating system type: Drupal VM / Ubuntu
* BLT version: 12.3

mikemadison13:
Fixes #4247 
--------
If this fixes an existing issue, link to that issue. Otherwise remove this section.

Motivation
----------
The stock BLT 12 default.local.drush.yml file does not align with config/build.yml in that the stock drush alias is not `self`.

Proposed changes
---------
Set it to `self`

Alternatives considered
---------
N/A

Testing steps
---------
1. Spin up a new blt 12 site with a Drupal VM or Lando w/ this patch
2. Run a command (e.g. `blt drupal:update`) that requires a drush alias
3. confirm that it works (not fails)

### DIFF
--- a/settings/default.local.drush.yml
+++ b/settings/default.local.drush.yml
@@ -1,2 +1,2 @@
 options:
-  uri: '${project.local.uri}'
+  uri: 'self'


danepowell:
The ORCA test fixture expects an actual URI here, not `self`. Using an actual URI seems like the correct behavior to me.

mikemadison13:
@danepowell  what would the URI be, though? because "inside" a VM there "is no URI"

danepowell:
I'm not sure how to do this in a way that's compatible with ORCA, if you can get tests passing let me know.

PR #4272, opened 2020-11-02
Loosen composer/semver version constraint for drupal/core 9.1.x

TravisCarden:
BLT is not currently installable on Drupal core 9.1.x because BLT requires `composer/semver` v1, but Drupal core 9.1.x requires v3:

```
Problem 1
  - drupal/core-recommended 9.1.x-dev requires composer/semver 3.2.2 -> satisfiable by composer/semver[3.2.2].
  - acquia/blt 12.x-dev requires composer/semver ^1.4 -> satisfiable by composer/semver[1.4.0, 1.4.1, 1.4.2, 1.5.0, 1.5.1, 1.5.2, 1.6.0, 1.7.0, 1.7.1, 1.x-dev].
  - Can only install one of: composer/semver[3.2.2, 1.7.1].
  - Can only install one of: composer/semver[3.2.2, 1.x-dev].
```

Loosening BLT's version constraint resolves the conflict. I assume the CI build will tell us whether or not it breaks any functionality.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "ext-json": "*",
         "acquia/drupal-environment-detector": "*",
         "composer-plugin-api": "^1.1.0 || ^2.0",
-        "composer/semver": "^1.4",
+        "composer/semver": "^1.4 || ^3.2",
         "consolidation/comments": "^1.0",
         "consolidation/config": "^1.0.0",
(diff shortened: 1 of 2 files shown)


danepowell:
Thanks for the PR, in fact I think this dependency can be removed completely: https://github.com/acquia/blt/pull/4273

PR #4273, opened 2020-11-03
Remove unused semver dependency

danepowell:
We introduced this dependency way back in https://github.com/acquia/blt/commit/44c532546 to support the security updates command.

Shortly after we removed it when refactoring the command: https://github.com/acquia/blt/commit/dda7343cce0b5ab5fcce2d63455ea242bb90d4a7

But we never actually removed the library from composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,6 @@
         "ext-json": "*",
         "acquia/drupal-environment-detector": "*",
         "composer-plugin-api": "^1.1.0 || ^2.0",
-        "composer/semver": "^1.4",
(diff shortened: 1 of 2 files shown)


danepowell:
@TravisCarden do you need a new stable release to unblock ORCA?

TravisCarden:
It's not blocking me today, @danepowell, but it will become an issue for everybody as soon as Drupal core 9.1.x gets an alpha. So don't treat it like a fire, but do plan for sooner than later. 👍

danepowell:
Okay I'll cut one today

PR #4285, opened 2020-11-16
Upgrade to ORCA 3

danepowell:
Motivation
----------
ORCA 3 has expanded coverage for Drupal 9 and we need to upgrade: https://github.com/acquia/orca/pull/115

Proposed changes
---------
Upgrade to ORCA 3 and use new job definitions that it provides

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: php
 
 os: linux
+dist: bionic
 
 version: ~> 1.0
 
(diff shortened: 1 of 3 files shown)


danepowell:
Well this is a conundrum. First Drupal 9.1 / PHPUnit 9 was causing a test failure because we were using a deprecated method: https://travis-ci.com/github/acquia/blt/jobs/443156480

Fine, I thought, we'll just rename the method, but that causes Drupal 9.0 / PHPUnit 8 tests to fail because the method does not exist yet: https://travis-ci.com/github/acquia/blt/jobs/443180411

danepowell:
Testing a possible fix from https://github.com/acquia/orca/pull/116

PR #4295, opened 2020-12-11
DX-3042: toggle:modules should use normal environment detection

danepowell:
Motivation
----------
The toggle modules command uses custom environment detection. This can produce unexpected results compared to other commands that use BLT core environment detection.

Proposed changes
---------
Rely on core environment detection by just checking the environment config.

### DIFF
--- a/src/Robo/Commands/Drupal/ToggleModulesCommand.php
+++ b/src/Robo/Commands/Drupal/ToggleModulesCommand.php
@@ -28,14 +28,8 @@ class ToggleModulesCommand extends BltTasks {
    * @validateDrushConfig
    */
   public function toggleModules() {
-    if ($this->input()->hasArgument('environment')) {
-      $environment = $this->input()->getArgument('environment');
-    }
-    elseif ($this->getConfig()->has('environment')) {
+    if ($this->getConfig()->has('environment')) {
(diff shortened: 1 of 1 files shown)


shelane:
So now when this is called in the context of another command that has the environment passed in as an argument, it reads all environments as "local" instead of the argument.
```
         try {
            $commands = ['internal:drupal:install'];
            $strategy = $this->getConfigValue('cm.strategy');
            if (in_array($strategy, ['core-only', 'config-split', 'features'])) {
              $commands[] = 'drupal:config:import';
            }
            $commands[] = 'llnl:site:content';
            $commands[] = 'drupal:toggle:modules';
            $this->invokeCommands($commands);
          }
          catch (BltException $e) {
            $this->say('<comment>Note:</comment> file permission error on Acquia Cloud can be safely ignored.');
          }
```
How exactly should I set the "config" per environment? (I only had the problem when moving from BLT 11.6.0 to 12.5.1)

shelane:
BTW, the comments for this command were not changed with this update:
```
   * You may define the environment for which modules should be toggled by
   * passing the --environment=[value] option to this command, setting the
   * 'environnment' environment variable, or defining environment in one of your
   * BLT configuration files.
```

PR #4301, opened 2020-12-16
Fixes #4300: adding config split as a BLT dependency.

### LINKED ISSUE #4300: BLT 12 assumes Config Split but does not ship Config Split
**Describe the bug**
BLT 12's default CM strategy is using the config split module. However, the new drupal-recommended-project repo doesn't have a dependency in composer for config split. The result is this bug.

**To Reproduce**
1. Run blt setup
2. Run drush cex 
3. Re-run blt setup (config will run this time)

**Expected behavior**
Config split should work out of the box.

**Detailed error output**
```
In PmCommands.php line 253:
                                                                               
  Unable to install modules config_split due to missing modules config_split.  
                                                                               
```

**System information**
* Operating system type: Mac OS
* Operating system version: 10.15.7
* BLT version: 12.4.0

**Additional context**
This used to be in blt-project so it wasn't an issue in BLT 11.

mikemadison13:
Motivation
----------
Fixes #4300

Adds config split.

Proposed changes
---------
Adds config split as a dependency of BLT so that the config commands "work" out of the box.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "consolidation/robo": "^1.4.12 || ^2",
         "dflydev/dot-access-data": "^1.1.0",
         "doctrine/annotations": "^1.10.0",
+        "drupal/config_split": "^1.7",
         "drupal/core": "^9.0.0-alpha1",
         "drupol/phposinfo": "^1.6",
         "drush/drush": "^10.2.2",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -25,6 +25,7 @@
         "consolidation/robo": "^1.4.12 || ^2",
         "dflydev/dot-access-data": "^1.1.0",
         "doctrine/annotations": "^1.10.0",
+        "drupal/config_split": "^1.7",
```
lpeabody, line 28: Maybe also allow for Config Split v2 to be more forward compatible?

mikemadison13: oh that's a great call @lpeabody. i just grabbed a rule and didn't think much about it!

mikemadison13:
closing in favor of ongoing architectural discussions on #4300

mikemadison13:
thanks for the suggestions @lpeabody !

lpeabody:
@mikemadison13 always happy to give unsolicited opinions ;)

PR #4305, opened 2020-12-17
Move Behat components to blt-behat plugin

danepowell:
Motivation
----------
To be released in conjunction with https://github.com/acquia/blt-behat/pull/7

Proposed changes
---------
We have a lot of Behat functionality still in BLT core that needs to be moved to the plugin, since Behat support is deprecated.


composer.json
```diff
@@ -38,7 +38,8 @@
         "zumba/amplitude-php": "^1.0"
     },
     "conflict": {
-        "drupal/drupal": "*"
+        "drupal/drupal": "*",
+        "acquia/blt-behat": "<=1.0.0"
```
danepowell, line 42: @mikemadison13 I think this will prevent users from upgrading BLT before the plugin and breaking Behat integration... does it look right to you? Kind of hard to test.

PR #4322, opened 2021-01-25
Update cache.settings.php for coding standard and require_once

vbouchet31:
File was using mixed variable naming convention.
File was included using require but the file may be included multiple times.


danepowell (approved):
Thanks for the PR!

PR #4326, opened 2021-01-28
Replace abandoned Sensiolabs security checker

paras-malhotra:
Motivation
----------
This PR replaces the abandoned [Sensiolabs security checker](https://github.com/sensiolabs/security-checker) with the [Enlightn security checker](https://github.com/enlightn/security-checker).

Alternatives considered
---------
I considered the local php security checker but it is licensed under AGPLv3 and can create issues. Also, it is based on binaries that depend on system architecture so this solution seemed better.

Testing steps
---------
You can run the security updates test.


danepowell:
Thanks, hopefully this is really a drop-in replacement. It's unfortunate that this deprecation was neither documented nor forewarned anywhere that I can find.

paras-malhotra:
@danepowell yes, this package works very similar to the older one. The only difference is that it only supports the JSON format for now. Everything else is the same (format of JSON, exit status code, advisories database used, etc.).

danepowell:
I couldn't resolve the merge conflicts on this because you used `main` instead of a feature branch, so I'm reopening as a new PR: https://github.com/acquia/blt/pull/4329

PR #4328, opened 2021-01-29
Behat tests giving timeout error on lando.

nileshlohar:
Motivation
----------
https://support.acquia.com/hc/en-us/articles/360047166273-Configuring-Acquia-BLT-with-Lando

This resolves the "Chrome timeouts" issue on behat test while running on lando.

Changes
---------
Add `--no-sandbox` option to launchChrome function.

Alternatives considered
---------
We might not able merge this as it can affect everyone but, just adding ready patch here for all lando users facing same issue.


danepowell:
Thanks. I think there is precedent for using the `no-sandbox` parameter (in ADS and Pipelines) so I'm fine including it here as well for all users. Since it's a dedicated test process I don't think the security implications are significant.

PR #4329, opened 2021-01-29
Replace abandoned Sensiolabs security checker

danepowell:
Replaces #4326


danepowell:
Test failure was due to https://github.com/ergebnis/composer-normalize/issues/673

Will work around it by manually resorting composer.json for now

santhosh-fernando:
I believe we should have a release for 11.x as well since this is breaking the current validation so we can consider this as a `Bug` than `Enhancement`

Issue #4352, opened 2021-03-01
Enable hooks support all stacks on site install


shashwatpurav-qed42:
Environments with multiple stacks e.g. 01dev, 02dev......08dev should have a common configurations for enabling and uninstalling the modules while installation.

Rewrite toggleModules method from ToggleModulesCommand class to have a common environments for multiple stacks to enable and uninstall modules while installation. E.g. for multiple stacks such as 01dev, 02dev......08dev should have common environment such as dev under modules key in blt.yml
1. This will save lines of code
2. Easier to maintain
3. Fallback mechanism wherein if any specific stack isn't mentioned in the blt.yml file, default environment can be used

danepowell:
Thanks for the PR. Unfortunately the Toggle Modules command was removed in BLT 13, so I don't think this PR or issue is relevant any longer. The community has contributed a toggle-modules plugin you might want to try: https://github.com/shelane/toggle-modules

PR #4353, opened 2021-03-01
4352: Enable hooks support all stacks on site install

shashwatpurav-qed42:
Fixes [4352](https://github.com/acquia/blt/issues/4352)

1. This will save lines of code
2. Easier to maintain
3. Fallback mechanism wherein if any specific stack isn't mentioned in the blt.yml file, default environment can be used

Rewrite toggleModules method from ToggleModulesCommand class to have a common environments for multiple stacks to enable and uninstall modules while installation. E.g. for multiple stacks such as 01dev, 02dev......08dev should have common environment such as dev under modules key in blt.yml

### DIFF
--- a/src/Robo/Commands/Drupal/ToggleModulesCommand.php
+++ b/src/Robo/Commands/Drupal/ToggleModulesCommand.php
@@ -16,30 +16,47 @@ class ToggleModulesCommand extends BltTasks {
   /**
    * Enables and uninstalls specified modules.
    *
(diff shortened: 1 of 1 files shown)


danepowell:
Thanks for the PR. Unfortunately the Toggle Modules command was removed in BLT 13, so I don't think this PR or issue is relevant any longer. The community has contributed a toggle-modules plugin you might want to try: https://github.com/shelane/toggle-modules

PR #4359, opened 2021-03-05
Update deprecated method root() call

TravisCarden:
[According to my tests](https://travis-ci.com/github/acquia/orca/jobs/488534302#L3088), there's _one single_ use of deprecated code in BLT--in fact, it's the last one in all Acquia modules _and_ contrib dependencies! 😮

I believe I've updated the code correctly according to the deprecation notice... We'll see what the tests say.


danepowell:
Thanks, I think I fixed tests for this in https://github.com/acquia/blt/pull/4363

TravisCarden:
Nice, @danepowell; thanks!

PR #4365, opened 2021-03-10
Remove toggle:modules command

danepowell:
**Motivation**

The toggle:modules command was added early in the Drupal 8 lifecycle when the config management ecosystem was still pretty immature. Nowadays there are much better alternatives for managing modules on a per-environment basis (namely, [Config Split](https://www.drupal.org/project/config_split)), and in fact modifying modules outside of that CM system is likely to cause conflicts.

**Proposed changes**

Remove toggle:modules.

**Merge requirements**
- [ ] _Major change_, _Minor change_, _Bug_, _Enhancement_, and/or _Chore_ label applied


dpagini:
As a `Major change` I assume this wouldn't be published in a release until a 13 branch, is that right?

I have projects that don't use config management from BLT (you can choose "none" as your cm option), so in that case "config:split" is not an option to use. I realize this makes sense for a majority of sites that can easily just config import/export for workflow, but for running larger multi-sites that's not an ideal option. Personally I would like to see this option stay in BLT. Are there stats you have just saying that most projects don't use this? J/w.

danepowell:
Correct, it'll be a 13 release.

Yes, this is one of the lesser used features of BLT, and just looking at it from a prescriptive point of view, a lot of discussion indicated that config split was a superior alternative.

If you're interested in continuing to use the functionality, I'd imagine it could be rolled into a plugin pretty easily. The command itself can be ported straight to a plugin, and then you just need to invoke it as part of the build steps, which can probably be done with hooks. If not, we can consider adding a hook.

shelane:
Yes, I would want it as a hook and plug-in. I use config split for different multi sites. You add environments on top of that and it’s hard to manage.

marvil07-adapt:
@danepowell , @shelane : I am wondering if this was moved into a plugin.
I have not been able to find it on the `acquia/` namespace in github.
Also, if not there, which blt hook/interaction point is recommended to use for this?

shelane:
I haven't seen it. I haven't updated to BLT 13 because of it. I have a multisite that already has config split for a few of the sites. I don't want to add a split for environments too. Usually it's to turn off modules either in local (SAML, acquia connector) or in prod (views UI, field UI, etc). If Acquia doesn't put it in their space as a plugin, I'll have to add it to my custom commands.

danepowell:
Sorry, Acquia does not plan to make `toggle:modules` available as a plugin.

marvil07-adapt:
@danepowell , @shelane : Thanks for the feedback!

PR #4380, opened 2021-06-29
Upgrade to bionic

danepowell:


yyss8:
Just for anyone who encounters this issue on behat tests after upgrading to `bionic`.
It looks like the `localhost` is now resolved as `127.0.1.1` but `BLT` is still running chrome on `127.0.0.1`.
We updated the `api_url` in the `example.behat.yml` file to use `127.0.0.1:9222` instead of `localhost:9222` and it's working now.

image

PR #4401, opened 2021-09-22
Update PHPUnit version to match Drupal 9

danepowell:


danepowell:
Actually this is pointless because we should move this recipe to https://github.com/acquia/blt-drupal-test, and it already requires core-dev

PR #4406, opened 2021-10-13
Adding PHPCS validation to preCommitHook.

logicbus:
When we get PHPCS errors on Pipeline builds, it's usually a surprise.  We want to avoid that surprise.

This change makes it so that PHPCS validate runs as part of precommit hooks.

I don't have alternate solutions, but I acknowledge that this PR may get refused because no everyone wants PHPCS validate on commit.  But since it's part of Pipelines, I think it should be.  

To test, make a code change that you know will fail PHPCS validation and commit it.  Before this PR, you can proceed and then you don't know there's a problem until your pipeline build fails.  Maybe that's 30 minutes later?  The next day?  With my change, you see the validation failure before you complete the commit.

The twig and yaml linters accept a file_list param, but it looks like PHPCS doesn't.  

I'd say this is an enhancement.

### DIFF
--- a/src/Robo/Commands/Internal/GitCommand.php
+++ b/src/Robo/Commands/Internal/GitCommand.php
@@ -79,6 +79,12 @@ class GitCommand extends BltTasks {
       );
     }
 
+    $collection->addCode(
+      function() {
+        return $this->invokeCommand('validate:phpcs');
+      }
+    );
+
     $result = $collection
       ->setVerbosityThreshold(VerbosityThresholdInterface::VERBOSITY_VERBOSE)
       ->run();


lpeabody:
Would be cool to see this made a more generic feature that accepts an arbitrary list of additional BLT targets to execute.

danepowell:
As Mike mentioned, this can't be merged as-is because it would throw an exception when the PHPCS plugin isn't installed. You probably also want to run `validate:phpcs:staged` as that will only scan changed files that are about to be committed, greatly speeding it up.

The reason this wasn't implemented initially was to keep commit linting as fast and lightweight as possible, but if `validate:phpcs:staged` works I'm open to the idea.

danepowell:
Actually, it looks like this could be a bug that will be fixed by https://github.com/acquia/blt/issues/4409

PR #4429, opened 2021-11-23
Allow admin password to be set by config if present.

pavlosdan:
**Motivation**
Currently BLT always generates a random password for user 1. On local environments we want to be able to define a username and password so we don't have to reset it after every single setup or use drush uli to get into the site.

**Proposed changes**
Since site:install allows for the --account-pass option, if a password is given in the blt config files, use it.

**Testing steps**
1. apply this patch
2. without changing anything else, run blt setup:drupal:instal -y --no-interaction
3. confirm random user name and password
4. add drupal.account.pass to your blt.yml file (optionally define a drupal.account.name as well for an easier test).
5. re-run blt setup:drupal:instal -y --no-interaction
6. confirm admin name and password are set to your custom ones defined in step 4.

### DIFF
--- a/config/build.yml
+++ b/config/build.yml
@@ -54,13 +54,16 @@ deploy:
   exclude_additions_file: ${repo.root}/blt/deploy-exclude-additions.txt
   gitignore_file: ${blt.root}/scripts/blt/deploy/.gitignore
 
-# File and Directory locations.
+# Drupal docroot.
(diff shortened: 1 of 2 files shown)


danepowell:
Thanks for the PR, looks good. I added the new `drupal.account.pass` config to build.yml.

PR #4455, opened 2022-01-11
Fixes UserConfig warnings for read-only systems

### LINKED ISSUE #4454: UserConfig assumes a writable directory
**Describe the bug**
When running any blt command, the `src/Robo/Common/UserConfig.php` class is instantiated which attempts to create a directory at `$HOME/.config/blt` if it does not exist. In a read-only system this throws a warning for every command which in some cases such as a deployment can make the logs extremely difficult to read (i.e. `blt artifact:ac-hooks:post-code-deploy`), it'd be nice if this could fail silently since there is no impact on the commands.

Note: `--no-interaction` has no bearing on this.

**To Reproduce**
- Set `$HOME/.config` to read only
- Run any blt command

You will see:
`W: PHP Warning:  mkdir(): Read-only file system in /app/vendor/acquia/blt/src/Robo/Common/UserConfig.php on line 42`

Depending on what command is run, you may also see:
`W: PHP Notice:  Trying to access array offset on value of type null in /app/vendor/acquia/blt/src/Robo/Common/UserConfig.php on line 90`

b-sharpe:
**Motivation**
Fixes #4454 

Prevents PHP warnings for `mkdir()` when a home directory is read-only. Also fixes a warning produced from UserConfig as a result of the read-only system.

**Testing steps**
- Set $HOME to readonly
- Run any blt command

Patch will prevent warnings and carry on.

### DIFF
--- a/src/Robo/Common/UserConfig.php
+++ b/src/Robo/Common/UserConfig.php
@@ -39,7 +39,7 @@ class UserConfig {
     }
     else {
       // Make sure directory tree for user config file exists.
-      if (!file_exists($configDir) && !mkdir($configDir, 0777, TRUE)) {
+      if (!file_exists($configDir) && (!is_writable(getenv('HOME')) || !mkdir($configDir, 0777, TRUE))) {
         return;
       }
       // Make sure directory for user config file is writable.
@@ -87,7 +87,7 @@ class UserConfig {
    *   Telemetry user data.
    */
   public function getTelemetryUserData() {
-    $data = $this->config['telemetryUserData'];
+    $data = $this->config['telemetryUserData'] ?? [];
     $data['app_version'] = Blt::getVersion();
 
     return $data;


danepowell:
I agree that a plethora of warnings on read-only filesystems is not ideal, but neither is silently failing to write config. This could cause serious confusion if people don't understand e.g. why they have to keep answering the telemetry prompt because it's not writing their preference.

b-sharpe:
@danepowell I'd agree with that. I see a few options here, let me know the most acceptable path forward:

1. Tie this into `--no-interaction` in which UserConfig is just not instantiated at all
2. Adjust the current PR, in which a empty/non-writable file is a "No" response so the question isn't asked over and over
3. Add another options such as `--no-telemetry` or similar to disable this

danepowell:
All of those are fine ideas, but not really solutions to the problem as I see it. We need to think bigger, since user config is about more than just telemetry.

If BLT cannot write user configuration, regardless of context, it should warn the user _at least once_ during each execution. The only exception would be if there's some sort of "suppress read-only warnings" flag, where a user explicitly acknowledges that no config will be written.

b-sharpe:
Seems to me it's very much about telemetry (at this time at least) given it's methods:
```
isTelemetrySet
isTelemetryEnabled
setTelemetryEnabled
getTelemetryUserData
setTelemetryUserData
save
```
Also from my search in code, `src/Robo/Blt::configDir()` also appears to only ever be used to construct UserConfig... are there other things I'm missing here?

The alternative to the above solutions I suppose would be to have the first call to configDir() check if writeable and set a property of the BLT class (since we can't write anywher) that then uses that property for subsequent checks.

PR #4466, opened 2022-01-28
Fix #4437: Allow child environment detectors to override false

### LINKED ISSUE #4437: The recommended way to use a custom Environment Detector in a non-acquia environment.
**I want to...**
I want to use BLT with a custom environment detector. After BLT 11, the documented method seems to be invalid. 
https://docs.acquia.com/blt/extending-blt/#overriding-the-environment-detector

**It's not working because...**
Environment Detector will incorrectly return true. What we will see, is in production environments config splits for local and prod are active. 

```
$ cat blt/src/Blt/Plugin/EnvironmentDetector/CustomEnvironmentDetector.php 
<?php

namespace SmusSetup\Blt\Plugin\EnvironmentDetector;

class CustomEnvironmentDetector extends EnvironmentDetector {


  public static function isLocalEnv() {
    return !self::isStageEnv() && !self::isProdEnv() && !self::isDevEnv();
  }

  public static function isDevEnv() {
    return getenv('ENVIRONMENT_ID') == 'dev';
  }

  public static function isStageEnv() {
    return getenv('ENVIRONMENT_ID') == 'stage';
  }

  public static function isProdEnv() {
    return getenv('ENVIRONMENT_ID') == 'prod';
  }

}
```

After digging into it, it looks like it is related to: vendor/acquia/blt/src/Robo/Common/EnvironmentDetector.php 
 which calls `parent::isLocalEnv()`, which will return true if it does not have the AH_ENV. 

```
#File: vendor/acquia/blt/src/Robo/Common/EnvironmentDetector.php 
....
  /**
   * Is local.
   */
  public static function isLocalEnv() {
    $results = self::getSubclassResults(__FUNCTION__);
    if ($results) {
      return TRUE;
    }

    return parent::isLocalEnv() && !self::isPantheonEnv() && !self::isCiEnv();
  }

```

```
#file vendor/acquia/drupal-environment-detector/src/AcquiaDrupalEnvironmentDetector.php


  /**
   * If this isn't a Cloud environment, assume it's local.
   */
  public static function isLocalEnv() {
    return !self::isAhEnv() || self::isAcquiaLandoEnv();
  }
...
...
  /**
   * Get Acquia hosting environment.
   *
   * @return string
   *   Environment name (e.g. dev, stage, prod).
   */
  public static function getAhEnv() {
    return getenv('AH_SITE_ENVIRONMENT');
  }

...
...
  /**
   * Is AH env.
   */
  public static function isAhEnv() {
    return (bool) self::getAhEnv();
  }



```
* BLT version: 12.x or 13.x

reubenmoes:
…e false is returned.

**Motivation**

Fixes #4437
There is currently no way to allow a custom environment indicator to say that an environment is **not active**. This was due to the parent filtering out any results if they are 'false'. This results in config split applying the 'local' config splits in any remote environment that is not on Acquia. 

**Proposed changes**

reserve the results from the custom environment indicator whether they return true or false. Then evaluate what the subclasses have returned. The assumption is that if we have a custom environment detector, that should override the capability and returning true/false within the custom environment indicator should be the end result, and it should not need to continue to the parent to check.  

**Alternatives considered**

I'm a bit unsure about the logic used. open to suggestions. 

**Testing steps**

Create a custom environment indicator and put it in a state where isLocalEnv returns false. Test in a non-acquia environment. Easiest to test with config split. User should see that local is not active (or local config split is not active) 

Example custom environment detector to test with.
```
<?php

namespace ConcertSetup\Blt\Plugin\EnvironmentDetector;

class CustomEnvironmentDetector extends EnvironmentDetector {

  public static function isLocalEnv() {  
  
     // This will return false, because we have set isProduction() to return true.
    return !self::isStageEnv() && !self::isProdEnv() && !self::isDevEnv();
  }

  /**
   * Is stage (on google cloud custom hosting).
   */
  public static function isStageEnv() {
    return getenv('ENVIRONMENT_ID') == 'stage';
  }

  /**
   * Is stage (on google cloud custom hosting).
   */
  public static function isDevEnv() {
    return getenv('ENVIRONMENT_ID') == 'dev';
  }

  /**
   * Is Prod (on google cloud custom hosting).
   */
  public static function isProdEnv() {
    return true;
    return $_ENV['ENVIRONMENT_ID'] == 'prod';
  }

}

```

**Merge requirements**
- [ ] _Major change_, _Minor change_, _Bug_, _Enhancement_, and/or _Chore_ label applied

### DIFF
--- a/src/Robo/Common/EnvironmentDetector.php
+++ b/src/Robo/Common/EnvironmentDetector.php
@@ -104,8 +104,8 @@ class EnvironmentDetector extends AcquiaDrupalEnvironmentDetector {
    */
   public static function isLocalEnv() {
     $results = self::getSubclassResults(__FUNCTION__);
-    if ($results) {
-      return TRUE;
+    if ( !empty($results) ) {
+      return !in_array(FALSE, $results);
     }
 
     return parent::isLocalEnv() && !self::isPantheonEnv() && !self::isCiEnv();
@@ -116,8 +116,8 @@ class EnvironmentDetector extends AcquiaDrupalEnvironmentDetector {
    */
   public static function isDevEnv() {
     $results = self::getSubclassResults(__FUNCTION__);
-    if ($results) {
-      return TRUE;
+    if ( !empty($results) ) {
+      return !in_array(FALSE, $results);
     }
 
     return self::isAhDevEnv() || self::isPantheonDevEnv();
@@ -128,8 +128,8 @@ class EnvironmentDetector extends AcquiaDrupalEnvironmentDetector {
    */
   public static function isStageEnv() {
     $results = self::getSubclassResults(__FUNCTION__);
-    if ($results) {
-      return TRUE;
+    if ( !empty($results) ) {
+      return !in_array(FALSE, $results);
     }
 
     return self::isAhStageEnv() || self::isPantheonStageEnv();
@@ -140,8 +140,8 @@ class EnvironmentDetector extends AcquiaDrupalEnvironmentDetector {
    */
   public static function isProdEnv() {
     $results = self::getSubclassResults(__FUNCTION__);
-    if ($results) {
-      return TRUE;
+    if ( !empty($results) ) {
+      return in_array(TRUE, $results);
     }
 
     return self::isAhProdEnv() || self::isPantheonProdEnv();
@@ -339,7 +339,7 @@ class EnvironmentDetector extends AcquiaDrupalEnvironmentDetector {
         $results[] = call_user_func([$detector, $functionName]);
       }
     }
-    return array_filter($results);
+    return $results;
   }
 
 }


src/Robo/Common/EnvironmentDetector.php
```diff
@@ -339,7 +339,7 @@ private static function getSubclassResults($functionName) {
         $results[] = call_user_func([$detector, $functionName]);
       }
     }
-    return array_filter($results);
+    return $results;
```
danepowell, line 342: The doc block for this function needs to be updated to account for false results being returned.

This significantly changes the behavior of multiple detectors being called, doesn't it? Previously, if any detector returned `true` that detector would 'win', i.e. the parent method would return true. Now, the result is going to vary by method. For instance, what should happen if one detector returns `false` and another returns `true` for `isStageEnv` or `isProdEnv`? I think this behavior needs to be consistent, and it needs to be documented.

reubenmoes: oh i have some inconsistency here. I had a change of approach last minute.  Which one would be preferred? 

`should happen if one detector returns false and another returns true for isStageEnv or isProdEnv`
This would result in it being overall false. 
If there is only one environment detector returning true - would be true
If there is only one environment detector returning false - would be false (and would not continue to call the parent method)
```
    if ( !empty($results) ) {
      return !in_array(FALSE, $results);
    }
```

`should happen if one detector returns false and another returns true for isStageEnv or isProdEnv`
This would result in it being overall true. 
If there is only one environment detector returning true - would be true
If there is only one environment detector returning false - would be false (and would not continue to call the parent method)
```
    if ( !empty($results) ) {
      return in_array(TRUE, $results);
    }
```

danepowell:
I don't see any clean solution to this that preserves backwards compatibility, sorry. If we can do this in a way that preserves BC I'd be willing to reconsider.

PR #4489, opened 2022-03-16
Fixes #4397 to place ci.blt.yml by default.

### LINKED ISSUE #4397: ci.blt.yml file missing
**Describe the bug**
As of late, I've noticed on new projects that the ci.blt.yml file is no longer being properly placed inside the blt folder, and it seems that projects attempting to run behat tests no longer function unless you manually place it. 

I did find that there's a ci.yml file floating around, maybe that is supposed to be working (but it doesn't appear to be). 

**To Reproduce**
1. Run `composer create-project acquia/drupal-recommended-project --no-interaction`
2. Add the blt-behat plugin
3. Generate the files with the recipe
4. Try to run behat tests in CI
5. they will fail
6. manually add the old ci.blt.yml file (see https://github.com/Drupal4Gov/Drupal-GovCon-2017/blob/develop/blt/ci.blt.yml) and it works. 

**Expected behavior**
If the new ci.yml file is expected to copy or work, it should (and it isn't). See https://github.com/acquia/blt/blob/main/scripts/blt/ci/internal/ci.yml.

mikemadison13:
**Motivation**
BLT ships with a ci.yml file (see https://github.com/acquia/blt/blob/main/scripts/blt/ci/internal/ci.yml) but it no longer places it as ci.blt.yml in the blt folder.

**Proposed changes**
Place the file during the scaffolding process.

**Testing steps**
1. Setup a new project using acquia/drupal-recommended-project
2. Add BLT
3. confirm that blt/ci.blt.yml is in place

### DIFF
--- a/scripts/blt/ci/internal/ci.yml
+++ b/scripts/blt/ci/internal/ci.yml
@@ -1,8 +1,5 @@
-modules.ci.enable: [ views_ui ]
-# We cannot use 127.0.0.1:8888 for vm hostname due to vagrant restrictions.
-vm.vagrant.hostname: local.${project.machine_name}.com
-drush.debug: false
-drush.verbose: true
 tests.run-server: true
 tests.drupal.sudo-run-tests: false
-project.local.hostname: 127.0.0.1:8080
+# The local.hostname must be set to 127.0.0.1:8888 because we are using drush runserver to test the site.
+project.local.hostname: 127.0.0.1:8888
+drush.debug: false
--- a/src/Robo/Commands/Blt/UpdateCommand.php
+++ b/src/Robo/Commands/Blt/UpdateCommand.php
@@ -51,8 +51,13 @@ class UpdateCommand extends BltTasks {
    * @throws \Acquia\Blt\Robo\Exceptions\BltException
    */
   public function addToProject() {
+    // Initializes the project template / scaffold.
     $this->initializeBlt();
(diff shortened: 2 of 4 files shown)


mikemadison13:
I'm getting a test failure here: https://github.com/acquia/blt/blob/main/tests/phpunit/src/DrupalSettingsTest.php#L41

```
$this->assertEquals($output_array['uri'], $drush_local_site_yml['options']['uri']);
```

This is failing with:

```
There was 1 failure:

1) Acquia\Blt\Tests\DrupalSettingsTest::testSetupDefaultLocalSettings
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://127.0.0.1:8080/'
+'http://127.0.0.1:8888/'
```

I think what's happening is that the ci.blt.yml file is resetting the value in the local.yml file (because we are in CI). I've updated the test to compare the local drush version output and the ci.blt.yml output to ensure they are the same. the drush status output is likely never going to show the 8888 port!

mikemadison13:
FYI @danepowell i am experiencing test failures here and the values are flipped in PHP 7.4 and 8.0. i'm not sure what's going on, but i may need some help digging into it.

pavlosdan:
This was what was missing to get our tests running in gitlab. +1 for creating the ci.blt.yml file during scaffolding. Or at the very list when the ci recipes or behat ini recipes runs.

PR #4497, opened 2022-03-28
Fixes #4496 to make a zsh version of git hooks.

### LINKED ISSUE #4496: Git Hooks Incompatible with zsh
**Describe the bug**
The default BLT git hooks (in https://github.com/acquia/blt/tree/main/scripts/git-hooks) explicitly define bash as the scripting engine, which causes them to fail on zsh terminals.

**To Reproduce**
1. run blt setup to install the hooks
2. attempt a git commit with the stock git hooks in place (and note failure)

I think that ideally BLT would provide a shell variable that lets us differentiate between bash or zsh hooks and make the path to the hooks dynamic.

mikemadison13:
**Motivation**
After updating to an M1 Mac the git hooks are failing. 

**Proposed changes**
Adds a configurable shell variable to BLT that will allow the toggling of git hooks based on shell. This can be customized per developer using local.blt.yml or per project using blt.yml. The default remains bash.


danepowell:
Didn't we discuss this? A custom zsh version shouldn't be necessary since the shebang forces the scripts to run in Bash regardless of your default shell.

The only reason you'd need a zsh version is if bash isn't installed at all, in which case you have much bigger problems than BLT.

PR #4510, opened 2022-05-02
Fixes #4504 to provide PHP 8.1 and Drupal 10 support.

### LINKED ISSUE #4504: DX-5105: Drupal 10 Support
Given the Drupal 10 development roadmap coming this year, I think we need to discuss how we want BLT to support Drupal 10. Some key considerations include:

- PHP 8.1.x requirement
- Other breaking dependency changes

Does this mean that we need to keep 13.x for Drupal 9.x and roll a BLT 14.x branch for Drupal 10.x? 

I also wonder if it makes sense to try and do some sort of reset on the BLT version numbers to better align with Drupal version. I realize that we have had a BLT 8 and 9 and 10 in the past, but "knowing that" BLT 11 is for 8, 13 (also 12) is for 9, and whatever else is for 10 seems confusing.

mikemadison13:
**Motivation**
With Drupal 10.x coming and PHP 8.1 support required, we need a version of BLT that is PHP 8.1 compatible. 

**Proposed changes**
Significant updates to packages and major / breaking changes to get us into PHP 8.1 compatibility. 

**Merge requirements**
- [ ] We must roll a BLT version for Drupal 10.x and PHP 8.1.x. We should discuss the naming schema (is it 14.x or something clearly marked as Drupal 10.x.?)

### DIFF
--- a/.github/workflows/orca.yml
+++ b/.github/workflows/orca.yml
@@ -37,14 +37,11 @@ jobs:
           - ISOLATED_TEST_ON_NEXT_MINOR_DEV
           - INTEGRATED_UPGRADE_TEST_TO_NEXT_MINOR_DEV
           - LOOSE_DEPRECATED_CODE_SCAN
-        php-version: [ "7.4" ]
+        php-version: [ "8.1" ]
         coveralls-enable: [ "FALSE" ]
         include:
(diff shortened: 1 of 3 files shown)


mikemadison13:
closing in favor of https://github.com/acquia/blt/pull/4511

PR #4511, opened 2022-05-02
Fixes #4504 to test PHP 8.1 and Drupal 10.x.

### LINKED ISSUE #4504: DX-5105: Drupal 10 Support

mikemadison13:
**Motivation**
With Drupal 10.x coming and PHP 8.1 support required, we need a version of BLT that is PHP 8.1 compatible. Fixes #4504 

**Proposed changes**
Adds PHP 8.1 supported versions of symfony packages (generally ^6) and specifically updates grasmash/yaml-cli and zumba/amplitude-php to php 8.1.x compatible versions. 

**Alternatives considered**
We could split out into a D10 specific branch, not sure if that is worth the effort.

### DIFF
--- a/.github/workflows/orca.yml
+++ b/.github/workflows/orca.yml
@@ -41,6 +41,9 @@ jobs:
         coveralls-enable: [ "FALSE" ]
         include:
           - orca-job: ISOLATED_TEST_ON_CURRENT
+            php-version: "8.1"
+            coveralls-enable: "FALSE"
+          - orca-job: ISOLATED_TEST_ON_CURRENT
             php-version: "8.0"
             coveralls-enable: "FALSE"
           - orca-job: ISOLATED_TEST_ON_CURRENT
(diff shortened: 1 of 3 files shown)


danepowell:
Looks great, thanks for the help. Have you actually tested BLT with D10? Are you aware of any other integration issues?

mikemadison13:
i haven't actually tested yet, no @danepowell but i'll be doing so and will open other issues as they crop up!

PR #4542, opened 2022-06-14
Fixes #4541: updating remaining exec logic.

### LINKED ISSUE #4541: DX-5126: Process Tasks in Executor Throwing Deprecated Warnings
**Describe the bug**
After updating our calls to use the array syntax instead of strings, it looks like we missed a few related to test execution as there are a number of exec calls that pass strings and the results are throwing:

```
Deprecation Warning: this command is passing a command string and should pass a command arary.
 [error]  You must have Google Chrome version 59+ to execute headless tests.
```

...I also can't apparently spell the word array.

See:
* https://github.com/acquia/blt/blob/main/src/Robo/Common/Executor.php#L137
* https://github.com/acquia/blt/blob/main/src/Robo/Common/Executor.php#L153

**To Reproduce**
1. run `blt tests:behat` and get the error

**Expected behavior**
Test execution works!

mikemadison13:
**Motivation**
Refactoring exec code to new command array structure.

**Proposed changes**
Remove the exec logic and pass command arrays.

**Alternatives considered**

**Testing steps**

**Merge requirements**
- [ ] Apply _one_ of the following labels: _Major change_, _Minor change_, _Bug_, _Enhancement_, or _Chore_

### DIFF
--- a/src/Robo/Commands/Tests/TestsCommandBase.php
+++ b/src/Robo/Commands/Tests/TestsCommandBase.php
@@ -157,7 +157,7 @@ class TestsCommandBase extends BltTasks {
    */
   protected function checkChromeVersion($bin) {
     $version = (int) $this->getContainer()->get('executor')
-      ->execute("'$bin' --version | cut -f1 -d'.' | rev | cut -f1 -d' ' | rev")
+      ->execute("'$bin'", "--version | cut", "-f1", "-d'.' | rev | cut -f1 -d'\" ' | rev",)
       ->run()
       ->getMessage();
 
--- a/src/Robo/Common/Executor.php
+++ b/src/Robo/Common/Executor.php
@@ -80,6 +80,7 @@ class Executor implements ConfigAwareInterface, IOAwareInterface, LoggerAwareInt
 
     // Backwards compatibility check for legacy commands.
     if (!is_array($command)) {
+      $this->say($command);
       $this->say(StringManipulator::stringToArrayMsg());
       $command = StringManipulator::commandConvert($command);
     }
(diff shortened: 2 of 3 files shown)


bmartinez287:
Thanks for fixing the Chromedriver issue(or at-least i imagine that fixes it). Took me a while to figure out the issue was in the 13.4 release of BLT. I rolled it back to 13.3 and I had no issues. The Chromedriver driver on 13.4 kept in Travis CI and the local environment at this point.
```
 /**
   * Verifies that Google Chrome meets minimum version requirement.
   *
   * @param string $bin
   *   Absolute file path to the google chrome bin.
   *
   * @throws \Acquia\Blt\Robo\Exceptions\BltException
   *   Throws exception if minimum version is not met.
   */
  protected function checkChromeVersion($bin) {
    $version = (int) $this->getContainer()->get('executor')
      ->execute("'$bin' --version | cut -f1 -d'.' | rev | cut -f1 -d' ' | rev")
      ->run()
      ->getMessage();

    if ($version < 59) {
      throw new BltException("You must have Google Chrome version 59+ to execute headless tests.");
    }
  }
````

File: src/Robo/Commands/Tests/TestsCommandBase.php [hunk 1]
```diff
@@ -157,7 +157,7 @@ protected function findChrome() {
    */
   protected function checkChromeVersion($bin) {
     $version = (int) $this->getContainer()->get('executor')
-      ->execute("'$bin' --version | cut -f1 -d'.' | rev | cut -f1 -d' ' | rev")
+      ->execute("'$bin'", "--version | cut", "-f1", "-d'.' | rev | cut -f1 -d'\" ' | rev",)
```
bmartinez287, line 160: Why does this make a difference? I saw the error but i was unsure how to test for it or what could have change.

bmartinez287: never-mind, its due to the command arrays change ah. That's interesting.

mikemadison13:
@bmartinez287 i am still working on this, i had dane revert but we need to dig in deeper. i'm having trouble reproducing the issue locally. but i think this "type" of fix is what we need to do. any help would be appreciated.

PR #4557, opened 2022-07-13
Fixes #4541 to resolve Drush Runserver Issues and Failed Behat Tests.

### LINKED ISSUE #4541: DX-5126: Process Tasks in Executor Throwing Deprecated Warnings

mikemadison13:
**Motivation**
The refactoring of command structures that occurred in https://github.com/acquia/blt/pull/4514 ended up causing the drush runserver commands in https://github.com/acquia/blt/blob/main/src/Robo/Commands/Tests/ServerCommand.php#L55. This PR refactors that and gets it working again. 
Fixes #4541 

**Proposed changes**
- Refactor the drush execution to diverge between execute and executeShell
- Update the wasSuccessful logic for drush runserver
- Adds a test for drush runserver

**Alternatives considered**
Removing the runserver / chrome logic entirely. This was opted against since runserver is useful for things other than behat.

**Testing steps**
Confirm that blt tests will launch a server properly, specifically `lando blt tests:server:start`

### DIFF
--- a/src/Robo/Commands/Tests/ServerCommand.php
+++ b/src/Robo/Commands/Tests/ServerCommand.php
@@ -54,17 +54,27 @@ class ServerCommand extends TestsCommandBase {
     /** @var \Acquia\Blt\Robo\Common\Executor $executor */
     $executor = $this->getContainer()->get('executor');
     $result = $executor
-      ->drush(["runserver", "--quiet", "$this->serverUrl > $log_file 2>&1"])
+      ->drush("runserver --quiet --uri=$this->serverUrl > $log_file 2>&1")
       ->background(TRUE)
       ->run();
+
+    if (!$result->wasSuccessful()) {
+      $output = NULL;
+      $unsuccessful = "Failed to execute Drush runserver on $this->serverUrl";
+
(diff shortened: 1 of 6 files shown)


mikemadison13:
@danepowell thanks to this passing PR and equivalent test on DGC (https://github.com/Drupal4Gov/Drupal-GovCon-2017/pull/1010) I'm pretty confident that runserver and tests are working again. I believe we can cut a 13.5 after merging this.

PR #4562, opened 2022-07-16
Update drush executor uri check

### LINKED ISSUE #4561: Inspector not calling correct site for config status on drupal:update
**Describe the bug**
When `drupal:update` is syncing updating configuration and then checking that the configuration matches, it is not checking the same site on a non-default site.

**To Reproduce**
1. Have config ignore with some different values in the database but with otherwise ignored config differences (site.settings is good)
2. Run `blt drupal:update --site=mysite` (site must be multisite and not the default) 
3. Read ` [notice] No differences between DB and sync directory.` on config-status output where the `--uri` is clearly called. Then see this error message that only comes from one place in blt: `checkConfigOverrides` and then the logic check of `$this->getInspector()->isActiveConfigIdentical()`.

```
[error]  Configuration in the database does not match configuration on disk. This indicates that your configuration on disk needs attention. Please read https://support.acquia.com/hc/en-us/articles/360034687394--Configuration-in-the-database-does-not-match-configuration-on-disk-when-using-BLT
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 
The command failed. This often indicates a problem with your configuration. Review the command output above for more detailed errors, and consider re-running with verbose output for more information.
```

**Expected behavior**
If the config-status is `No differences between DB and sync directory`, then why would the inspector return false? This is where @mikemadison13 and I spent some time debugging. Although there was a recent change to the `isActiveConfigIdentical` to remove the `--uri` check there, it was because it was expected to be already set in the executor drush function. PR is coming with tested fix.

**Detailed error output**
```
 [Acquia\Blt\Robo\Tasks\DrushTask] Running /var/www/vendor/bin/drush @self config-status --uri=trip --no-interaction in /var/www/docroot
 [notice] No differences between DB and sync directory.
 [Acquia\Blt\Robo\Tasks\DrushTask] Done in 21.663s
 [error]  Configuration in the database does not match configuration on disk. This indicates that your configuration on disk needs attention. Please read https://support.acquia.com/hc/en-us/articles/360034687394--Configuration-in-the-database-does-not-match-configuration-on-disk-when-using-BLT
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 
The command failed. This often indicates a problem with your configuration. Review the command output above for more detailed errors, and consider re-running with verbose output for more information.

 [error]  Command `drupal:config:import --site=trip` exited with code 1.
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 

real	4m22.574s
user	0m4.357s
sys	0m7.712s

[ ~/.../docksal/d9]
 (feature/DRUPAL8-22-22-updates-reorder)-> drush @self config-status --uri=trip
 [notice] No differences between DB and sync directory.
 ```

**System information**
* BLT version: 13.5.1

shelane:
**Motivation**
Solves issue where different site is compared than the site original check on config-status. 
Fixes #4561

**Proposed changes**
Wider solution also to correctly set the `--uri` in the Executor drush function so that ALL executions will have `--uri` preloaded and it will not need to be added to individual functions.
This PR changes the check of the drush alias to check that the site is not default. Sets the drush_array to include the `--uri` from the config value.
Also changed is removing the redundant check (no longer necessary by virtue of change above) in the inspector `isDrupalInstalled` function.
Added is the debug logging for the actual status difference so that when users are running in verbose mode and the inspector  `isActiveConfigIdentical` contains a difference, the user will know exactly what that difference being returned is without having to make additional calls to find that difference. (In the case of this bug, it was not the same as calling the drush command directly, which made finding the issue to be more difficult.)

This should not impact users or require any additional changes to user code or blt code.

**Alternatives considered**

The `--uri` check could have been added back to inspector function, but that is less efficient than providing it for all executed drush functions and removing redundancies.

**Testing steps**
Issue #4561 contains testing steps.

### DIFF
--- a/src/Robo/Common/Executor.php
+++ b/src/Robo/Common/Executor.php
@@ -82,8 +82,8 @@ class Executor implements ConfigAwareInterface, IOAwareInterface, LoggerAwareInt
 
     // URIs do not work on remote drush aliases in Drush 9. Instead, it is
     // expected that the alias define the uri in its configuration.
-    if ($this->getConfigValue('drush.alias') != 'self') {
-      $drush_array[] = ' --uri=' . $this->getConfigValue('site');
+    if ($this->getConfigValue('site') !== 'default') {
+      $drush_array[] = '--uri=' . $this->getConfigValue('drush.uri');
     }
 
     if (is_array($command)) {
--- a/src/Robo/Inspector/Inspector.php
+++ b/src/Robo/Inspector/Inspector.php
@@ -189,8 +189,7 @@ class Inspector implements BuilderAwareInterface, ConfigAwareInterface, Containe
    */
   public function isDrupalInstalled() {
     $this->logger->debug("Verifying that Drupal is installed...");
-    $uri = $this->getConfigValue('drush.uri');
-    $result = $this->executor->drush(["--uri=$uri", "status", "bootstrap"])->run();
+    $result = $this->executor->drush(["status", "bootstrap"])->run();
     $output = trim($result->getMessage());
     $installed = $result->wasSuccessful() && strpos($output, 'Drupal bootstrap : Successful') !== FALSE;
     $this->logger->debug("Drupal bootstrap results: $output");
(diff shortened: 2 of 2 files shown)


mikemadison13:
we shouldn't merge this just yet, see my commentary on the issue #4561. this is a work around until we establish if we're going to make it worse, though.

danepowell:
@mikemadison13 advised in Slack that this is good to merge

PR #4566, opened 2022-07-21
testing acms install command.

mikemadison13:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**


danepowell:
@mikemadison13 are you still working on this? I think people have also proposed running install commands from ACLI or even from the project templates via a plugin. Seems like we should agree on an approach.

mikemadison13:
i think this is a non-starter given it would only work with the 2.x branch of acms.

PR #4582, opened 2022-08-23
update environment detector

danepowell:


vishnu11moorthy:
Hello @danepowell  

Thanks for the fixes which actually through error in the environment detection below the error.
Fatal error: Declaration of Acquia\Blt\Robo\Common\EnvironmentDetector::getSiteName($site_path) must be compatible with Acquia\DrupalEnvironmentDetector\AcquiaDrupalEnvironmentDetector::getSiteName(string $site_path): ?string in /var/www/html/vendor/acquia/blt/src/Robo/Common/EnvironmentDetector.php on line 233


Your commit changes fix the issue but unfortunately the version 12.8.4 not even have this changes its still in main branch. can you help to move this to the latest tag.

Thanks

mikemadison13:
@vishnu11moorthy 12.x is no longer supported and we will not be back porting this fix to 12.x. You can pin the environment detector to version 1.4.1 to avoid the issue, or update to blt 13.x (preferably update to blt 13.x).

Issue #4522, opened 2022-05-16
Execute `drush deploy:hook` at the end of `blt drupal:update`


pfrenssen:
When running `blt drupal:update` we are currently missing a way to execute post update hooks after config import. Support for `hook_deploy_NAME()` has been added 2 years ago to Drush (ref. [drush deploy command](https://www.drush.org/latest/deploycommand/)).

There was a feature request to completely replace `blt drupal:update` with `drush deploy` in #4252 but this idea was abandoned since we would lose the more advanced use cases that are supported in `blt drupal:update` (such as allowing separate core-only and config-split config management strategies).

**Describe the solution you'd like**
We are already doing the majority of the steps done by `drush deploy`, except running the deploy hooks. Would it be perhaps a good idea to run `drush deploy:hook` at the end of our current `blt drupal:update` process? This way we can keep our advanced use cases, and still offer a "standard" way for site owners to run update hooks after importing config.

jcnventura:
I support this.

jcnventura:
This should be closed now that #4612  was merged.

PR #4612, opened 2022-10-14
Add support for drush's deploy hook.

pavlosdan:
**Motivation**
In certain cases we want to run update hooks that need to happen after updatedb and config:import have ran. E.g to update content after new fields have been added to a content type. Or to update module configuration that do not use drupal's configuration api and as such their configuration is not captured by the config management system.

**Proposed changes**
Drush provides a hook we can implement to facilitate such changes which runs when invoking `drush deploy` or `drush deploy:hook`. See: 
- https://www.drush.org/latest/commands/deploy_hook/ 
- https://github.com/drush-ops/drush/blob/master/tests/functional/resources/modules/d8/woot/woot.deploy.php
- https://github.com/drush-ops/drush/pull/4359

**Alternatives considered**
BLT's native "[post-config-import](https://docs.acquia.com/blt/extending-blt/#pre-and-post-command-hooks)" command.
- Not every project is running BLT and customers adopting BLT may already be used to running their drush deploy hooks since that's a more prevalent Drupal tool in the community.

**Testing steps**
- Implement a deploy hook as shown in the 2nd bullet point above. 
- Apply the patch from this PR.
- Run `blt drupal:update`

Related issue: #4522

### DIFF
--- a/src/Robo/Commands/Drupal/ConfigCommand.php
+++ b/src/Robo/Commands/Drupal/ConfigCommand.php
@@ -186,7 +186,8 @@ class ConfigCommand extends BltTasks {
 
   /**
    * Runs drush's deploy hook.
-   * See: https://www.drush.org/latest/commands/deploy_hook/.
+   *
+   * @see https://www.drush.org/latest/commands/deploy_hook/
    *
    * @command drupal:deploy:hook
    *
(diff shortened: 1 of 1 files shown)


mikemadison13:
this request originally made in https://github.com/acquia/blt/issues/4522

danepowell:
Thanks for the PR. Please fix the failing tests.

Issue #4692, opened 2023-07-18
Additional logging during drupal:update command


dan612:
**Is your feature request related to a problem? Please describe.**
When performing code deployments using BLT the `drupal:update` command is invoked. This command should update the database and import configuration. However, if the configuration management strategy is set to `none` in BLT.yml then this will simply return with a cache clear (https://github.com/acquia/blt/blob/main/src/Robo/Commands/Drupal/ConfigCommand.php#L60) prior to any additional information being output.

**Describe the solution you'd like**
Add some more logging within the [conditional](https://github.com/acquia/blt/blob/main/src/Robo/Commands/Drupal/ConfigCommand.php#L58C5-L58C32):
```
if ($strategy === 'none') 
```
something to let the user running the command know that in their BLT.yml file they have specified `none` - and this means no configuration import is going to take place. 

**Describe alternatives you've considered**
n/a

PR #4693, opened 2023-07-18
BLT-4692: Add logging message if cm strategy is none.

dan612:
**Motivation**
Addresses #4692 

**Proposed changes**
Additional logging in drupal:update command

**Alternatives considered**
n/a

**Testing steps**
Run a deployment with a BLT based site with the following in `blt.yml`
```
cm:
  strategy: none
```
then see the additional log output message during deployment logs. Or simply run `blt drupal:update` locally with that same setting in place


danepowell:
Thanks for the contribution!

PR #4694, opened 2023-07-19
Add BLT to ORCA fixtures

secretsayan:
**Motivation**
ORCA removed BLT from its integrated fixtures, so we need to add BLT back using its own packages_alter so that it continues to be tested with ORCA.


danepowell:
Thanks!

Issue #4701, opened 2023-09-22
BLT-5224: Add back PHP code sniffing in pre-commit hook


brunodbo:
PHP code sniffing was removed from the pre-commit hook in https://github.com/acquia/blt/commit/552ff157a6b6de3c11207ed505a4ea5ca7b7053e. Would it be possible to add it back? Unless there is a reason not to - it's unclear why it was removed, there is no issue referenced on the commit.

danepowell:
Code sniffing was moved to a plugin, which is why it can't be part of the core commit hook: https://github.com/acquia/blt-phpcs

brunodbo:
@danepowell That makes sense, closing this issue and associated PR. Thanks for the insight!

PR #4702, opened 2023-09-22
#4701: Add back PHP code sniffing in pre-commit hook

brunodbo:
PHP code sniffing was removed from the pre-commit hook in https://github.com/acquia/blt/commit/552ff157a6b6de3c11207ed505a4ea5ca7b7053e. This commit adds it back to the pre-commit hook. Unless there is a reason not to - it's unclear why it was removed, there is no issue referenced on the commit where it was removed.


brunodbo:
Closing this pull request, following @danepowell's comment in #4701.

PR #4747, opened 2024-04-02
BLT-5237: Fix failing PHPUnit tests.

vishalkhode1:


danepowell:
It's a little hard to tell, but I think it's actually a problem with recommended-settings tests killing the integrated test:
```
PHP Fatal error:  Uncaught Error: Class "Acquia\Drupal\RecommendedSettings\Tests\FunctionalTestBase" not found in /home/runner/work/blt/orca-build/vendor/acquia/drupal-recommended-settings/tests/src/Functional/Config/ConfigInitializerTest.php:17
```

I don't understand what would cause that since the class seems to exist in DRS 1.0.0 as expected.

vishalkhode1:
> It's a little hard to tell, but I think it's actually a problem with recommended-settings tests killing the integrated test:
> 
> ```
> PHP Fatal error:  Uncaught Error: Class "Acquia\Drupal\RecommendedSettings\Tests\FunctionalTestBase" not found in /home/runner/work/blt/orca-build/vendor/acquia/drupal-recommended-settings/tests/src/Functional/Config/ConfigInitializerTest.php:17
> ```
> 
> I don't understand what would cause that since the class seems to exist in DRS 1.0.0 as expected.

Should be good now. The above error is coming on [acquia/drupal-recommended-settings](https://github.com/acquia/drupal-recommended-settings) plugin as well and it's due to a bug in ORCA and we've resolved it by creating a patch for it. See [this](https://github.com/acquia/drupal-recommended-settings/blob/develop/.github/workflows/drs_ci.yml#L80).
The same error was coming on BLT becaus  we've updated packages_alter.yml file in last PR and forgot to remove later.

